Issue #2, opened 2016-06-20
[Enhancement] Add wayback machine archival support.


belldandu:
Since the wayback machine tries to preserve the integrity of each site there should be no real change to any parser stuff except for maybe small parts. 
```parserFactory.register("web.archive.org/web/*someregexthatallowsnumbersonly*/http://www.baka-tsuki.org", function() { return new BakaTsukiParser() });```

This is just a suggestion and it would be neat to have implemented. I emailed you already about this so yeah.

belldandu:
Looked into this some more and decided to change it up a bit.

```parserFactory.register({url: "www.baka-tsuki.org", allowedArchives: ["web.archive.org"]}, function() { return new BakaTsukiParser() });```

Also after doing some testing i noticed that sometimes the archive site gives a 
![gen](https://cloud.githubusercontent.com/assets/5639384/16183782/ba6c07e4-3683-11e6-9a11-79238671bc4c.png)
I'm Guessing there is a rouge image that doesn't follow your image scheme somewhere on the archival page. Which means you would have to make sure that its actually following the image scheme before putting it in the list of images you want to parse.

dteviot:
URL that has problem please?
If you can identify the rouge image, even better.

belldandu:
https://web.archive.org/web/20140803022146/http://www.baka-tsuki.org/project/index.php?title=Toaru_Majutsu_no_Index:Volume1
https://web.archive.org/web/20140805114418/https://www.baka-tsuki.org/project/index.php?title=CubexCursedxCurious:Volume1
https://web.archive.org/web/20140810093554/http://baka-tsuki.org/project/index.php?title=High_School_DxD:Volume_1
https://web.archive.org/web/20140803143924/https://www.baka-tsuki.org/project/index.php?title=No_Game_No_Life:Volume_1
Etc etc

belldandu:
Well that's quite odd. It looks like its having trouble replacing the cover image file on the archive because that's the url returned. It doesn't matter which file i select as the cover either.

belldandu:
It looks like this issue is occuring on the thumbnail images. Specifically the ones that appear twice on the page (for whatever reason). I would suggest making the list ignore duplicates as thats the underlieing issue here. The parent element is already replaced and then it tries to replace it a second time when none exists. @dteviot

belldandu:
At this point the easiest thing to do is to check if parentElement is not null.
[...]

PR #3, opened 2016-06-20
Add web.archive.org support

belldandu:
This pull request is to add the enhancement at https://github.com/dteviot/WebToEpub/issues/2

### DIFF
--- a/plugin/js/ParserFactory.js
+++ b/plugin/js/ParserFactory.js
@@ -8,6 +8,15 @@ var parserFactory = (function () {
 
     let parsers = new Map();
 
+    var isWebArchive = function(hostName) {
+        return extractHostName(hostName).startsWith("web.archive.org") ? 1 : 0;
+    }
+
+    var stripWebArchive = function(url) {
+        var hostName = url.split('://');
+        return hostName[2] ? "https://" + hostName[2] : url; 
+    }
+
     var stripLeadingWww = function(hostName) {
         return hostName.startsWith("www.") ? hostName.substring(4) : hostName;            
     }
@@ -21,6 +30,9 @@ var parserFactory = (function () {
     };
 
     var fetch = function(url) {
+        if (isWebArchive(url)) {
+            url = stripWebArchive(url);
+        }
         let hostName = stripLeadingWww(extractHostName(url));
         let constructor = parsers.get(hostName);
         return (constructor === undefined) ? undefined : constructor();
--- a/plugin/js/ParserFactory.js
+++ b/plugin/js/ParserFactory.js
@@ -9,7 +9,7 @@ var parserFactory = (function () {
     let parsers = new Map();
 
     var isWebArchive = function(hostName) {
-        return extractHostName(hostName).startsWith("web.archive.org") ? 1 : 0;
+        return extractHostName(hostName).startsWith("web.archive.org");
     }
 
     var stripWebArchive = function(url) {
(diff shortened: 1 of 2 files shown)


belldandu:
Yeah i forgot about that e.e its fixed now.

belldandu:
As far as i know by design your supposed to get the height and width While the image loads. I also tested this and all the images dimensions are set properly. However i can also parse the data listed on the html page to get the size if you still wan't to avoid taking any chances.

dteviot:
I think there's a mis-communication here.
I have trouble seeing how you can get the dimensions while loading is in
progress,  Until you have the file you don't know how big it is.

Here's my thoughts.
1. I'm happy with the WebArchive and the image removal bug fixes.
2. Fetching the large scale image.  Not so much.  I personally prefer the
scaled down images.  (Although I'm now suspecting most people want the full
scale images.)

So, I'd accept 1, and 2 I want' to think about a bit.

belldandu:
@dteviot 
I could do this but even this could have potential issues.
```
    let span = util.getElement(div, "span");
    var img = span.innerHTML.split("(")[1];
    if(img){
        img = img.split(" pixels,")[0].split(" x ");
        imageInfo.height = img[0].replace(",", "")
        imageInfo.width = img[1].replace(",", "");
    }
```

Also as for your one question https://davidwalsh.name/get-image-dimensions 
This has been around for a long time. Everything in onload is called imediatly after the image is loaded.

dteviot:
> Everything in onload is called imediatly after the image is loaded.

That's what I understood.
However, onLoad is only called when the image is loaded.  Which occurs some time in the future.  Until onLoad is called the height and width values of imageInfo are not valid.  So if the main thread tries to read imageInfo before onLoad is called, the values will be wrong.  I.e. need to wait until onLoad is called.  See http://www.html5rocks.com/en/tutorials/es6/promises/, then look at function BakaTsukiImageCollector.prototype.fetchImages.

dteviot:
i.e. I think updateImageInfoFromImagePage needs to return a promise that is resolved when onLoad finishes.

Please fogive me, I need to go away and do some thinking.

belldandu:
Thats why i put the return inside of onload.

In your promise, then cannot be called until after ```that.updateImageInfoFromImagePage(rawDom, imageInfo);``` has returned something.

By placing the return inside of onload you keep the integrity of the values as nothing is returned until after the image has been fully loaded.

At least thats how it should do and has done so far. I can even try limiting the connection in the browser and test to make sure its still working.

belldandu:
And ok

belldandu:
Ok yeah just tested definetly needs a promise.

belldandu:
There we go it now is promised and i tested it to. Works perfectly even when throttled to 50kbps with 500ms ping. @dteviot

dteviot:
Thanks.  
On quick glance, looks OK.
I'm afraid it's getting late for me, and I need to go to sleep.
Will probably merge tomorrow.

belldandu:
Awesome. and night

belldandu:
@dteviot o.o

dteviot:
I've merged it into the current dev branch "Sonako" and will send updated build 0.0.0.4 to Chrome Webstore.  If I hear no complaints in a few days, will merge to trunk.

belldandu:
Ok

belldandu:
Ignore those last 2

dteviot:
> Confused as to how de40496 managed to be slightly different from 09dbe2c Did you edit it afterwards?

Yes.  Got merge conflicts and had to fix.
Took liberty to make some minor tweaks.

belldandu:
Awesome.

belldandu:
I'm moving onto the next pull request to fix https://github.com/dteviot/WebToEpub/issues/4
Feel free to make tweaks.

Issue #8, opened 2016-06-22
Remove duplicate images from image gallery at start of Web Page


dteviot:
Many Baka-Tsuki web pages have an image gallery at the start of the web page.
Some of these images also appear in the story text.
Provide an option to have the ePUB generator remove any images in the gallery that also appear in the text.

**Suggested implementation notes**.
1. - BakaTsukiImageCollector.prototype.findImagesUsedInDocument() is used to do an initial scan of the document, collecting information on each image.  This scan should be updated to check if image element is nested inside a list element of class gallerybox.  If image element is not nested inside gallerybox, add flag to imageInfo that it appears outside gallery. 
2. - Later on BakaTsukiParser.prototype.processImages() is used to replace the image elements.  This should be updated to check that if element is nested inside a gallery box, and the image is also not in a gallery box, then remove the element.

(Obviously, the above processing needs to be done BEFORE the images in the gallerybox are “flattened” to be outside the box.)

belldandu:
Finished in #10

dteviot:
Implemented in release 0.0.0.4

Issue #4, opened 2016-06-20
[Enhancement] BakaTsuki epub formatting


typhoon71:
First thing thanks for this extension, for which I'd like to suggest some improvement:

1. Download the hi-res images when available from BakaTsuki ("original image"), not the resized ones, let the epub reader scale it [for better quality].
2. Put one image /page [for cleaner layout and easier printing].
3. Improve formatting; if you compare the now dead BakaTsuki epub Generator output and the output of this plugin you will notice the output is messy. A good example is the "content" table appearing, or the text spacing being messy [for cleaner layout].

I suppose that if the old code could do it should be possible to do it now too.
I hope I don't sound annoying, it's not my intention, I'm asking since the web thing was really handy and had god results quality wise.

belldandu:
Please refer to #3 and #2 for your 1. and do +1 the pull request if you approve. As for 2. and 3.
I will be handleing these in another pull request.

typhoon71:
Good news then, I like how I just have to wait and stuff does itself...
About the "+1": I did it on the first comment, is it OK?

belldandu:
Yep

belldandu:
The new Format will be like this
```
/
	/META-INF
		/container.xml
	/mimetype
	/OEBPS
		/content.opf
		/Images
			/imagename.jpg
		/Styles
			stylesheet.css
		/Text
			ChapterName.xhtml
		/toc.ncx
```

dteviot:
@belldandu
I'm not entirely sure this is what Typhoon is asking for.
I think he's talking about the layout of the "Images" section that appears at the start of the ePUB.

I'll add that I've seen at least three people asking for the images that appear in both the images table and the main body to be removed from the images table.

belldandu:
@dteviot 

Yes i know but the entire file structure needs organizing as well. So i figured i would do that as part of it.

Also because some chapters can have the same names (in the case of some baka-tsuki books)
it will be ```[split_number] ChapterName.xhtml```

dteviot:
> Also because some chapters can have the same names (in the case of some baka-tsuki books)
it will be [split_number]ChapterName.xhtml

I don't understand why you need to do this.
The current code allocates a new unique name to each file it stores in the ePUB, regardless of where it came from.  Perhaps if you gave me an example?

belldandu:
[...]

PR #10, opened 2016-06-22
Improve Epub Format and remove images from gallery that are outside t…

belldandu:
…he gallery.

Adds the enhancements #4 and #8

### DIFF
--- a/plugin/js/EpubItemSupplier.js
+++ b/plugin/js/EpubItemSupplier.js
@@ -19,7 +19,7 @@ EpubItemSupplier.prototype.setChapters = function (chapters) {
 
 EpubItemSupplier.prototype.createXhtmlFileName = function (fileIndex) {
     let that = this;
-    return "OEBPS/Text/[" + util.zeroPad(fileIndex) + "]" + util.getChapterTitle(this.items[fileIndex].chapterTitle) + ".xhtml";
+    return "OEBPS/Text/[" + util.zeroPad(fileIndex) + "]" + util.getTitle(this.items[fileIndex].chapterTitle) + ".xhtml";
 }
 
 // return id attribute to go into <item> element in <manifest>
--- a/plugin/js/EpubItemSupplier.js
+++ b/plugin/js/EpubItemSupplier.js
@@ -18,8 +18,8 @@ EpubItemSupplier.prototype.setChapters = function (chapters) {
 }
 
 EpubItemSupplier.prototype.createXhtmlFileName = function (fileIndex) {
-    // Belldandu: For now only new format gets implemented because not sure where this files functions are even called and i dont want to accidentally break something.
-    return "OEBPS/Text/[" + util.zeroPad(fileIndex) + "]" + ".xhtml";
+    let that = this;
+    return "OEBPS/Text/[" + util.zeroPad(fileIndex) + "]" + util.getTruncatedChapterTitle(this.items[fileIndex].chapterTitle) + ".xhtml";
 }
 
 // return id attribute to go into <item> element in <manifest>
--- a/plugin/js/EpubItemSupplier.js
+++ b/plugin/js/EpubItemSupplier.js
@@ -19,7 +19,7 @@ EpubItemSupplier.prototype.setChapters = function (chapters) {
 
 EpubItemSupplier.prototype.createXhtmlFileName = function (fileIndex) {
     let that = this;
-    return "OEBPS/Text/[" + util.zeroPad(fileIndex) + "]" + util.getTruncatedChapterTitle(this.items[fileIndex].chapterTitle) + ".xhtml";
+    return "OEBPS/Text/[" + util.zeroPad(fileIndex) + "]" + util.getChapterTitle(this.items[fileIndex].chapterTitle) + ".xhtml";
 }
 
 // return id attribute to go into <item> element in <manifest>
--- a/plugin/js/EpubItemSupplier.js
+++ b/plugin/js/EpubItemSupplier.js
@@ -18,12 +18,13 @@ EpubItemSupplier.prototype.setChapters = function (chapters) {
 }
 
 EpubItemSupplier.prototype.createXhtmlFileName = function (fileIndex) {
-    return "index_split_" + util.zeroPad(fileIndex) + ".html";
+    // Belldandu: For now only new format gets implemented because not sure where this files functions are even called and i dont want to accidentally break something.
+    return "OEBPS/Text/[" + util.zeroPad(fileIndex) + "]" + ".xhtml";
 }
 
 // return id attribute to go into <item> element in <manifest>
 EpubItemSupplier.prototype.createId = function(fileIndex) {
-    return "html" + util.zeroPad(fileIndex);
+    return "xhtml" + util.zeroPad(fileIndex);
 }
 
 // used to populate manifest
--- a/plugin/js/EpubPacker.js
+++ b/plugin/js/EpubPacker.js
@@ -157,7 +157,7 @@ EpubPacker.prototype = {
         if (epubItemSupplier.hasCoverImageFile()) {
             let guide = that.createAndAppendChild(opf.documentElement, "guide");
             let reference = that.createAndAppendChild(guide, "reference");
-            reference.setAttribute("href", EpubPacker.coverImageXhtmlHref());
+            reference.setAttribute("href", EpubPacker.coverImageXhtmlHref().substr(6));
             reference.setAttribute("title", "Cover");
             reference.setAttribute("type", "cover");
         };
@@ -218,7 +218,7 @@ EpubPacker.prototype = {
         navPoint.setAttribute("playOrder", playOrder);
         let navLabel = that.createAndAppendChild(navPoint, "navLabel");
         that.createAndAppendChild(navLabel, "text", chapterInfo.title);
-        that.createAndAppendChild(navPoint, "content").setAttribute("src", chapterInfo.src);
+        that.createAndAppendChild(navPoint, "content").setAttribute("src", chapterInfo.src.substr(6));
         return navPoint;
     },
 
--- a/plugin/js/EpubPacker.js
+++ b/plugin/js/EpubPacker.js
@@ -20,11 +20,11 @@ function EpubPacker(metaInfo) {
 }
 
 EpubPacker.coverImageXhtmlHref = function() {
-    return "cover.xhtml";
+    return "OEBPS/Text/Cover.xhtml";
 }
 
 EpubPacker.coverImageXhtmlId = function() {
-    return "cover.xhtml";
+    return "cover";
 }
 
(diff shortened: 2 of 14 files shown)


plugin/js/EpubItemSupplier.js
```diff
@@ -18,12 +18,13 @@ EpubItemSupplier.prototype.setChapters = function (chapters) {
 }
 
 EpubItemSupplier.prototype.createXhtmlFileName = function (fileIndex) {
-    return "index_split_" + util.zeroPad(fileIndex) + ".html";
+    // Belldandu: For now only new format gets implemented because not sure where this files functions are even called and i dont want to accidentally break something.
+    return "OEBPS/Text/[" + util.zeroPad(fileIndex) + "]" + ".xhtml";
```
belldandu: Chapter Name was not done here because i'm not even sure this function gets called so there is no title to send to it.
Instead it follows part of the new format which is ```[0000].xhtml``` and so on.

plugin/js/parsers/BakaTsukiImageCollector.js
```diff
@@ -111,7 +112,11 @@ ImageElementConverter.prototype.replaceWithImagePageUrl = function (images) {
     let imageInfo = images.get(that.imagePageUrl);
     if (imageInfo != null && that.element.parentElement != null) {
         let newImage = imageInfo.createImageElement();
-        that.element.parentElement.replaceChild(newImage, that.element);
+        if(that.element.className === "thumb" && (imageInfo.isOutsideGallery || imageInfo.isCover)){
```
belldandu: imageInfo.isCover is never set to true or grabbed anywhere. This behavior should be changed in the future so that isCover is set to true for whatever image the checkmark is on for "Set Cover".

belldandu:
@dteviot whenever your ready.

belldandu:
@dteviot got any comments you want to add?

belldandu:
The overall changes are the following:
- Epub Format now uses a more common structure making editing easier.
- Chapter splits when possible use number + 20 character restricted chapter name otherwise they just use number
- Images that appear elsewhere within the epub that is to be made are removed from the novel illustration.
- Listing for images as well as their caption text in Novel Illustrations has been removed as this prevented images from getting their own pages.

All changes have been tested and seen working before this push was made.

dteviot:
Sorry, today has been chaos, (and tomorrow will probably be just as bad) so I haven't had time to examine your submission in detail.
On first look, it's pretty good, although there's some minor points I'd raise.

- There's not really any need to put your name in comments.  Git will show who's to blame.
- BakaTsukiParser.prototype.stripGalleryBoxWidthStyle() should realy be renamed as it's no longer doing that.
- After moving the images out of the list, you could discard the entire list.  which means you no longer need to dump the gallerytext elements.
- In EpubItem.prototype.getZipHref() I'd move the logic for creating/truncating a filename into separate function.  Share with EpubItemSupplier.prototype.createXhtmlFileName(), also BakaTsukiImageInfo.prototype.makeZipHref()  maybe put in Util?
- that.element.className === "thumb" && (imageInfo.isOutsideGallery || imageInfo.isCover), in ImageElementConverter.prototype.replaceWithImagePageUrl() I'd pull into own function.
- Unit tests are broken.
- Stripping images is an optional feature.  That said, I have some ideas on how to handle that and will take care of it this weekend.

Should have time to examine fully day after tomorrow.

belldandu:
Unit tests are broken for me regardless saying everything is undefined. They where broken the first day i cloned this repo.

as for 

> After moving the images out of the list, you could discard the entire list. which means you no longer need to dump the gallerytext elements.

I tried that. It led to images just not existing at all because i cant go higher or lower than 1 level for whatever reason. listItem.parentNode.parentNode === null and listItem.firstChild.firstChild === null

Other than that i cleaned up the other stuff based on your recommendations and attempted to fix the Utests for BakaTsuki.

@dteviot

plugin/js/Util.js
```diff
@@ -45,9 +45,10 @@ var util = (function () {
         svg.setAttribute("version", "1.1");
         svg.setAttribute("preserveAspectRatio", "xMidYMid meet");
         svg.setAttribute("viewBox", "0 0 " + width + " " + height);
+        svg.setAttribute("data-origin", origin);
         let newImage = doc.createElementNS("http://www.w3.org/2000/svg","image");
         svg.appendChild(newImage);
-        newImage.setAttribute("xlink:href", href);
+        newImage.setAttribute("xlink:href", ".." + href.substr(5));
```
belldandu: replace "OEBPS" with "..". Only gets done when retrieving href for the html. Do not do this elsewhere or it will break a lot of things.

plugin/js/EpubPacker.js
```diff
@@ -131,7 +131,7 @@ EpubPacker.prototype = {
     addManifestItem: function(manifest, href, id, mediaType) {
         let that = this;
         var item = that.createAndAppendChild(manifest, "item");
-        item.setAttribute("href", href);
+        item.setAttribute("href", href.substr(6));
```
belldandu: In certain places (specifically the manifest only) having "OEBPS/" before each item causes the list index to be out of range because of where content.opf and toc.ncx is.

We remove "OEBPS/" Only when inserting chapter href's into the manifest and nowhere else. Doing this anywhere else breaks a lot of things.

belldandu:
Well i'm off to bed. I don't mind fixing the unittests however i need them to actually work first and not say this.
UnitTests?

belldandu:
There we go fixed a bunch of unit test stuff for baka tsuki.
Gonna start working on the other ones now. @dteviot

belldandu:
Actually all the other errors are just "Access to restricted URI denied" @dteviot and i have no idea whats doing it so Have fun with that.

belldandu:
Seems like you have missing functions in the sonako branch and its trying to call them but can't. @dteviot

belldandu:
You seem to be mostly using ECMA6 so why not use Class and Class Extends instead of function and Object.create @dteviot ?

belldandu:
Because your current usage falls under "Bad practice: Extension of native prototypes"

Refer to https://developer.mozilla.org/en-US/docs/Web/JavaScript/Inheritance_and_the_prototype_chain#Bad_practice_Extension_of_native_prototypes

unitTest/UtestEpubPacker.js
```diff
@@ -200,7 +200,7 @@ test("makeCoverImageXhtmlFile", function (assert) {
             "<body>" +
                "<div class=\"svg_outer svg_inner\">" +
                     "<svg xmlns=\"http://www.w3.org/2000/svg\" xmlns:xlink=\"http://www.w3.org/1999/xlink\" height=\"100%\" width=\"100%\" version=\"1.1\" preserveAspectRatio=\"xMidYMid meet\" viewBox=\"0 0 400 200\">" +
-                        "<image xlink:href=\"cover.png\" height=\"200\" width=\"400\"/>" +
+                        "<image xlink:href=\"...png\" height=\"200\" width=\"400\" data-origin=\"http://dummy/cover.png\"/>" +
```
belldandu: This only happens because your function throws an error saying its not implemented yet.

belldandu:
Fixed some more UnitTest Stuff that i noticed

belldandu:
The rest is up to you on whether you want to implement class and class extends

plugin/js/parsers/BakaTsukiImageCollector.js
```diff
@@ -111,7 +112,18 @@ ImageElementConverter.prototype.replaceWithImagePageUrl = function (images) {
     let imageInfo = images.get(that.imagePageUrl);
     if (imageInfo != null && that.element.parentElement != null) {
         let newImage = imageInfo.createImageElement();
-        that.element.parentElement.replaceChild(newImage, that.element);
+        if(0 === 1){ // placeholder for some check if we should remove the duplicate images (currently resolves to false)
```
belldandu: Some variable that tells the epub gen to remove duplicate images. 
Resolves to false atm for obvious reasons.

belldandu:
@dteviot There i fixed as much of the unit tests as i could and added the ability to choose whether or not to remove duplicate images. I also made it so that the cover image is also properly removed from the Novel Illustration if said option is enabled.

belldandu:
removeduplicateimagesoption

belldandu:
Upon closer inspection of the "Restricted Access uri" errors i noticed that this is in its own essense cross origin policy issues. You cannot access the sites you are trying to access because they are of a different origin then your current url.

I suggest changing this behavior so you don't have these issues.

dteviot:
@belldandu 
> Apparently there is no way to even meet the spec of CORS for files loaded from the file:// protocol as the origin will always be null

see: http://www.chrome-allow-file-access-from-file.com/

> You seem to be mostly using ECMA6 so why not use Class and Class Extends instead of function and Object.create @dteviot ?

Because I'm still learning Javascript, and am not yet familiar with Classes.

> Because your current usage falls under "Bad practice: Extension of native prototypes"
Um.. no.  the relevant text is "One mis-feature that is often used is to extend **Object.prototype or one of the other built-in** prototypes."
I'm using this pattern: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Introduction_to_Object-Oriented_JavaScript

belldandu, I'm going to ask you to please stop submitting stuff for a couple of days until I have time to go through, assess and merge what you've done.  I know you're eager to get the gallery formatting stuff in.  (I am too.)  But I'd like to avoid you having to rework anything else you're doing based on what I'm willing to accept.
I'd also like to co-ordinate our development tasks so we don't wind up trying to do the same things.

belldandu:
@dteviot Ok

belldandu:
```
    var getChapterTitle = function (title) {
        if(title) {
            title = title.replace(/([^a-z0-9_\- ]+)/gi, '');
            return (title.length > 20 ? title.substr(0, 20) + "..." : title);
        }
        return "";
    }

    var getImageName = function (page) {
        if(page){
            let name = page.split(/file:/gi)[1];
            if(name){
                let image = name.split(/\./gi)[0].replace(/([^a-z0-9_\- ]+)/gi, '');
                return (image.length > 20 ? image.substr(0, 20) + "..." : image);
            }
        }
        return "";
    }
```

toshiya44:
This might be slightly off-topic, but can you write a piece javascript that saves images from a baka-tsuki page, and that can be run on Firefox [scratchpad](http://i.imgur.com/IeQ8xHm.png)? A bit of explanation on how it works will be good too.

dteviot:
@toshiya44

> can you write a piece javascript that saves images from a baka-tsuki page, and that can be run on Firefox scratchpad? 

I don't have time to do this at moment.
I suggest you have a look at these:
Using javascript to make HTTP requests http://www.html5rocks.com/en/tutorials/es6/promises/ 
Parsing HTML: http://www.codeproject.com/Articles/1060680/Web-to-EPUB-Extension-for-Chrome

Issue #24, opened 2016-06-28
Remove "empty" <div> sections.


dteviot:
"Guest" wrote:

Using the "remove duplicate images" option leaves a bunch of empty
```html
<div>
	
		</div>
<div>
			
		</div>
```
scattered around in the illustrations html.
I mean, I guess it doesn't really affect the user-facing result so it's hardly high priority, but it does seem a wee bit untidy.

belldandu:
Causes #28

belldandu:
@dteviot can this be closed.

dteviot:
I want to write a couple of test cases for it.

belldandu:
ok

belldandu:
@dteviot Have you added those test cases?

dteviot:
Yes. https://github.com/dteviot/WebToEpub/commit/338e07bfd5b9768109786e3127f84ae11c383d2d

PR #30, opened 2016-06-28
Remove blank elements

### LINKED ISSUE #27: BakaTsuki - Content list (table of contents)
Small fix needed.
If the table of content is hidden while viewing a BakaTsuki novel, the plugin will show a page like this.

ee

Can an option be added to remove it completely if it's hidden?
And maybe another option to automaticlly "show" it instead?

### LINKED ISSUE #28: BakaTsuki - Empty page after Image gallery
In calibre, after the images at the start of a epub, there's a white page, just before the novel start.
Can you check if it's caused by the reader or not? It's not supposed to be there.
I can't try with sumatra because there's the link issue still around (so the page isn't empty and I don't know if it's on calibre side).

belldandu:
Fixes #27 & #24 
Was supposed to fix #28 by fixing #24 however because the epub readers also treat the novel illustrations as all one page even though pressing down gives you a new image (dafuq) this is unfixable without causing a crap ton of discrepancies between epub readers.

The best way to get rid of the blank page would be to simply add content to it.

### DIFF
--- a/plugin/js/parsers/BakaTsukiParser.js
+++ b/plugin/js/parsers/BakaTsukiParser.js
@@ -182,9 +182,8 @@ BakaTsukiParser.prototype.stripGalleryBox = function (element) {
     }
 }
 
-// discard blank divs and page created when moving elements
+// discard blank divs created when moving elements
 BakaTsukiParser.prototype.stripBlankElements = function(element){
-	util.removeElements(util.getElements(element, "p", e => (e.innerHTML.replace(/\s/g, "") == "")));
 	util.removeElements(util.getElements(element, "div", e => (e.innerHTML.replace(/\s/g, "") == "")));
 }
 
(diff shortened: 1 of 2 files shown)


dteviot:
I'm going to display my ignorance of HTML by asking a couple of dumb questions.
- Why did you change search by className === "toc" to id === "toc"?  I'm guessing because id should be more specific?
- Should we remove empty <p> sections?  My understanding is that even an empty <p> will add a new line.  And it's possible this has been added for layout effects.  e.g. extra space between lines/paragraphs for emphasis.

belldandu:
Ok. I'll explain. On baka tsuki the div with the attribute "class" of  "toc" also has the attribute "id" of toc.

When a user (on the baka-tsuki pages that have the show and hide capabilities) chooses to hide the toc on the html page the attribute "class" of "toc" is changed to "toc tochidden" where as the attribute "id" remains "toc". 

className is javascript refers to an elements attribute "class" where as id in javascript refers to an elements attribute "id"

As for the new line stuff. We should in fact discard any empty ```<div>``` 

As for the ```<p>``` tags i see your point.

dteviot:
@belldandu
> When a user (on the baka-tsuki pages that have the show and hide
capabilities) chooses to hide the toc on the html page the attribute
"class" of "toc" is changed to "toc tochidden" where as the attribute "id"
remains "toc".

Sorry, I noticed #27 after reading this email.  Comes from reading email
newest first.

typhoon71:
Is it just me or the div are still there?

PR #154, opened 2018-01-27
new parser for German Gutenberg Projekt

GallusMax:
my poor contribution to this great tool.. adding epub functionality to license free german books

### DIFF
--- a/plugin/js/parsers/GutenbergDEParser.js
+++ b/plugin/js/parsers/GutenbergDEParser.js
@@ -1,0 +1,154 @@
+/*
+  Template to use to create a new parser
+*/
+"use strict";
+
+// Use one or more of these to specify when the parser is to be used
+
+// Use this function if site's host name is sufficient.  
+// i.e. All pages are on same site, and use same format.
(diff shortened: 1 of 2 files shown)


dteviot:
Thanks very much for the contribution.
However, I'd like a couple of changes made before I accept it.
1. Change name of class from TemplateParser.  I'd suggest GutenbergDEParser.  You'll need to update lines 10 and 22.
2. Remove the comments (and commented out code from the template.

GallusMax:
sorry, I obvioulsy missed that.
There are still some gutenbergDE books, that will not produce an epub. I will fix this in another PR.

GallusMax:
that was fun. The fix extracts Author and Title even for those books that do not come with a frontpage with clean metadata. please include as well! Thanks again for the shoulder to stand on! :-)

dteviot:
@GallusMax
> Thanks again for the shoulder to stand on! :-)

You are very welcome.  The whole point of providing the template was to allow others to contribute.
Minor note, I forgot to run eslint over the code before accepting it.  There were issues with formatting and some unreachable code which I've fixed.  But you might like to double check.  I've also uploaded new version to Chrome store, and submitted to Mozilla.

PR #177, opened 2018-05-27
Reader View

yetisc:
Firefox has useful "Reader View" feature that strips useless parts of page to improve reading experience. This is available as stand alone library [mozilla/readability](https://github.com/mozilla/readability) . This PR adds new setting to optionally enable running downloaded pages through Readability.js. I tested it on few pages, at least it works on anythingnovel.com using latest firefox and default parser. 

There may be errors in code, i'm not too familiar with JS. Wikipedia doesn't work with this, probably many other sites too.

Example with setting disabled/enabled:

disable
enable

### DIFF
--- a/plugin/_locales/en/messages.json
+++ b/plugin/_locales/en/messages.json
@@ -187,6 +187,10 @@
 		"message": "Use <svg> for images",
 		"description": "Label in front of checkbox to choose between <img> and <svg> elements for images."
 	},
+	"__MSG_label_Use_Reader_View__": {
+		"message": "Use reader view",
+		"description": "Label in front of checkbox to choose whether to run chapter through readability.js"
+	},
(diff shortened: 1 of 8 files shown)


dteviot:
@yetisc
Thank you very much for this submission.
The Mozilla Readability library seems like a very good idea.
However, there are several issues with your submission.
1. Git submodule should be used to include Readability library.
2. Readability should only be used by the default parser, not all parsers.  So, should not be hooked into the HttpClient class.
3. JSDom parser is not needed.
4. Register new setting by adding value in UserPreferences constructor.
5. Don't put an "untranslated" entry in the Vietnamese translations.

My apologies for pointing out all these issues.  
I'm not trying to insult your work, I'm actually really impressed with your efforts. 
But I think it might be easier if I just steal your idea and do the work.

PR #181, opened 2018-06-08
updated lnmtl parser to show or hide original(raw) text

Hogesyx:


dteviot:
Thank you for this.
There's a couple of minor issues I'd like you to correct, and then I'd be happy to accept this.
1. You need to add your name (Hogesyx is fine) to the Credits section of readme.md
2. You've changed default behavior.  "Remove Original" must default to true in popup.html and UserPreferences.js, so that existing behavior will continue until a user explicitly changes the setting. (Note, setting should be persisted, so once it is turned on, it will be remembered.)
3. Fix broken unit test "customRawDomToContentStep" in UtestLnmltParser.js  Basically, add the following after line 16 where the LnmtlParser is created.
```javascript
    parser.userPreferences = {
        removeOriginal: {
            value: true
        }
    };
```

Hogesyx:
Thanks for looking into this, I happen to found this code this morning and so did a quick hack so that I can download some chapters before my flight so didn't do much checking.

dteviot:
@Hogesyx
You didn't update UserPreferences.js to set default to true, but I'll take care of that.
Incidentally, or future reference, please don't commit each file change individually to Git with a comment like "Updated 'filename'", it's not useful to someone looking through the commits.  See: https://chris.beams.io/posts/git-commit/

PR #196, opened 2018-09-07
Added cgtranslations.me into WordpressBaseParser

kairusds:


dteviot:
Why do you think this change is needed?

This code in the WordpressBaseParser examines the HTML and decide if it's WordPress
https://github.com/dteviot/WebToEpub/blob/3c337de2fda5d82e8ad33eaf96b90578ad8ef88b/plugin/js/parsers/WordpressBaseParser.js#L16-L23

As far as I can tell, this code works for that site.
(Note, the parserFactory.register() lines predate the parserFactory.registerRule().

kairusds:
Okay, I'll close the issue then. I thought it was actually important to add sites into WordpressBaseProvider, my bad.

PR #206, opened 2018-10-12
Parse author name from a series page at wuxiaworld.co

asifm91:
Author's name is not being parsed in the current version and the default value `<unknown>` is being used. I've implemented the extractAuthor function in the WuxiaworldCoParser.js


dteviot:
Thank you for this.

PR #207, opened 2018-10-17
Qidian parser extractAuthor function

asifm91:
The current parser fails to extract author's name and falls back to default parser, returning `<unknown>`. This might be due to changes of the layout.

I've modified the extractAuthor function to properly parse the author's name.


dteviot:
Please commit to ExperimentalTabMode branch (that's the work in progress branch.) not Master. 
Master is for tracking releases.
See: https://github.com/dteviot/WebToEpub/blob/ExperimentalTabMode/CONTRIBUTING.md
Also, please squash this submission into a single commit.

PR #285, opened 2019-08-02
Russian localization

iStyx:
Adds Russian localization.


dteviot:
@iStyx  - Thank you

iStyx:
@dteviot Ukrainian localization is on its way as well :)

dteviot:
@iStyx When is it likely to be done?  If it's a few days, I'll hold off from updating the version in the store until then.  If it's a couple of weeks, I'll update the version in the store immediately.

iStyx:
@dteviot I'll make a PR tomorrow, so if you'd hold off updating store version until then that would be great.

dteviot:
@iStyx  Will do.

PR #378, opened 2020-09-08
Fix for comrademao

AlenToma:


dteviot:
@AlenToma 
Thank you, I appreciate this, but I'm not going to merge.
1. I've already added your changes to findContent() and fileCoverImageUrl()
2 Files ProjectSettings.json and applicationhost.config are not needed.  I think they were created by MS Visual Studio
3. Removing the "incognito" line from manifest.json means it won't work on Chrome in Incognito mode.  (Yes, I know that Firefox will refuse to load the extension if it sees that value, that's why the build scripts remove it from the Firefox version)
4. You commented out the "disabled()" function. Until the Site owners give permission, the version of extension in Chrome web store must not support Comrademao..

AlenToma:
I see im sorry,

I did not notice the extra files created. and thx for informing me about "incognito"

dteviot:
@AlenToma
No problem.  I really do appreciate your effort.  It's just I'd already looked at it and stolen the bits that were useful.

PR #456, opened 2021-01-13
fanfiction.net more info (+WLPublishingParser minor fix)

JimmXinu:
The Information page for fanfiction.net had three problems:
- The 'category' or 'fandom', such as 'Books > Harry Potter' or 'Cartoons > Young Justice' wasn't being included.  Now included.
- Published/Updated times of '19m ago' or '5h ago' aren't useful long term.  Parsed datetimes to string, plus keep hidden seconds-since-epoch attribute.
- URLs in info page where coming out relative to `chrome-extension://<big ID>/...`. Make URLs in info page absolute from document source.

Adding category seemed straight-forward.  I'm not sure that the other changes are done the most 'preferred' way; I considered using `cleanInformationNode()` instead, but it's not really 'cleaning' as such.  If you'd prefer a different approach, please advise me and I'll give it a try.

Keeping the hidden seconds-since-epoch attribute is purely selfish--I hope to parse WebToEpub output in FanFicFare and FFF users _really_ care about metadata like that; but parsing locale time is a pain and including ISO time is ugly.

Also, an unrelated minor fix for WLPublishingParser -- the 'manually select parser' list had two 'Default' values.  Gave WLPublishingParser a name.  Didn't think it was worth a separate PR.

### DIFF
--- a/package.json
+++ b/package.json
@@ -39,6 +39,7 @@
     { "name": "snnsnn"},
     { "name": "Tom Goetz"},
     { "name": "Sergii Pravdzivyi"}
+    { "name": "JimmXinu"}
   ],
   "license": "GPL-3.0-only",
   "bugs": {
--- a/plugin/js/parsers/FanFictionParser.js
+++ b/plugin/js/parsers/FanFictionParser.js
@@ -56,8 +56,33 @@ class FanFictionParser extends Parser {
         return this.extractTextFromProfile(dom, "a");
     }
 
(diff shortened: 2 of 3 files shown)


JimmXinu:
Of course, I noticed that 'Books > Harry Potter' was coming out as 'BooksHarry Potter' right _after_ submitting the PR.  Throw a `>` in place of the CSS :after.

dteviot:
@JimmXinu
Thank you very much.

Issue #468, opened 2021-02-02
Website Request https://www.novicetranslations.com


1jamie7:
please add this site https://www.novicetranslations.com 🙏🙏🙏 thank you ❤❤❤

gamebeaker:
i tried to make a pull request for #468 and #467 pls. write if it worked.

dteviot:
@gamebeaker 
>i tried to make a pull request for #468 and #467 pls. write if it worked.

It did not.
FWIW, here's a tutorial on how to do it.
Looking at your fork of WebToEpub, (https://github.com/gamebeaker/WebToEpub) you haven't pushed the change(s) to your repository.  (Step 6) below.

1. ~~Find a project you want to contribute to~~  (done)
2. ~~Fork it~~ (done)
3. Clone it to your local system
4. Make a new branch
5. Make your changes
6. Push it back to your repo
7. Click the Compare & pull request button
8. Click Create pull request to open a new pull request

gamebeaker:
@dteviot ok i think i did it

dteviot:
@gamebeaker 
> ok i think i did it

Yes, I got it.  Will take me a while to review, Hopefully get back to you within 24 hours.

dteviot:
@1jamie7, 
Test versions for Firefox and Chrome have been uploaded to https://drive.google.com/drive/folders/1B_X2WcsaI_eg9yA-5bHJb8VeTZGKExl8?usp=sharing. Follow the "How to install from Source" instructions at https://github.com/dteviot/WebToEpub#how-to-install-from-source and let me know how it goes.

Thanks to @gamebeaker for doing the work.

dteviot:
@1jamie7
Updated version (0.0.0.127) has been submitted to Firefox and Chrome stores.
Firefox version is available now.
Chrome might be avaiable in 1 to 3 weeks.
If you can't wait, Test versions for Firefox and Chrome have been uploaded to https://drive.google.com/drive/folders/1B_X2WcsaI_eg9yA-5bHJb8VeTZGKExl8?usp=sharing. Follow the "How to install from Source" instructions at https://github.com/dteviot/WebToEpub#how-to-install-from-source and let me know how it goes.

1jamie7:
Thank you so much! ❤❤

Issue #467, opened 2021-02-02
Support for alphapolis.co.jp


JasCU100:
Can https://www.alphapolis.co.jp/ be supported? Thank you!

dteviot:
@JasCU100,
Test versions for Firefox and Chrome have been uploaded to https://drive.google.com/drive/folders/1B_X2WcsaI_eg9yA-5bHJb8VeTZGKExl8?usp=sharing. Follow the "How to install from Source" instructions at https://github.com/dteviot/WebToEpub#how-to-install-from-source and let me know how it goes.

Thanks to @gamebeaker for doing the work.  Note, I can't read Japanese, so I can't verify the work.

dteviot:
@JasCU100
Updated version (0.0.0.127) has been submitted to Firefox and Chrome stores.
Firefox version is available now.
Chrome might be avaiable in 1 to 3 weeks.
If you can't wait, Test versions for Firefox and Chrome have been uploaded to https://drive.google.com/drive/folders/1B_X2WcsaI_eg9yA-5bHJb8VeTZGKExl8?usp=sharing. Follow the "How to install from Source" instructions at https://github.com/dteviot/WebToEpub#how-to-install-from-source and let me know how it goes.

PR #469, opened 2021-02-02
Alphapolis, Novicetranslations parser

gamebeaker:
issues #467
Alphapolisparser problem:
chapter titel are pretty long because they contain date of creation and word count.
issues #468
Novicetranslations problem:
gets too many chapterlinks (first few) because bad class selector.


dteviot:
Thanks very much
A couple of minor issues.
1. Some inconsistencies with the indentation (extractAuthor in Alphapolisparser) and popup.html
2. Package.json needs a comma.

But I'll accept it as is and fix it myself.

PR #481, opened 2021-03-01
Add ability to manually set fetching delay per chapter

kondeeza:
### Pull Request Check List

- Do all existing unit tests pass?  `yes. 566 assertions of 566 passed, 0 failed. (Firefox)`
- Have all warnings and errors reported by eslint been fixed? `yes. that was a tough one because the whole npm thing is new to me.  Managed to follow all instruction until this step "Command line for eslint is "eslint packed.js > error.txt" then got eslint command not recognised error. Ended up just installing eslint plugin for vscode which then throw indentation error on my line and I later corrected`
- If you've added new behavior, Can a user turn it off? Is it off by default? `It's off by default, Accessible in Advanced Options , no reason to throttle if it works normally after all`
- Have you updated the "contributors" section of packages.json with your name? (and ideally email, so I can contact you easily.) `yes`
- Have you updated the "Credits" section of readme.md with your name? `yes`
- Are you committing to the ExperimentalTab branch? `Eh, I'm new to working with others in Github and its branch feature. Should I be working on ExperimentalTab branch instead of the main one? Don't really know whether experimental branch is for "almost certainly working commits waiting for more extensive test" or "experimental experimental commits that's halfway working". Let me know if I should pull merge request to experimental branch instead. Sorry for wordy question`
- Have you rebased your commit?  `no`

### What does it do?

Options to Throttle/delay fetching webpage by x seconds/chapter. Accessible via Advanced option (Off by default).

I saw some people in the issue tab had similar issue of wanting to throttle and had to throttle via 'object inspector' as hotfix. I want to make it easier for basic users to throttle speed because even I don't know how to throttle via 'object inspector'.

### Why? isn't error 429, 509  auto exponential throttle enough?

There were some site that just outright block you for a whole day after fetching 50 chapters within an hour (even with no simultaneous fetching), causing the whole packing to fail.  This is useful for users with 24/7 PC who are willing to throttle chapter fetching  as long as epub gets compile eventually.

It's also compatible with simultaneous fetching, so you can get 1 Chapter /3 Secs , 2 Chapter /5 Secs,  8 Chapter /1 Minutes, etc. (Or even adjust speed during the fetching process)

### Screenshots

sleeping_icon

newconfig

### DIFF
--- a/package.json
+++ b/package.json
@@ -39,7 +39,8 @@
     { "name": "snnsnn"},
     { "name": "Tom Goetz"},
     { "name": "Sergii Pravdzivyi"},
-    { "name": "JimmXinu"}
+    { "name": "JimmXinu"},
+    { "name": "Kondeeza"}
   ],
   "license": "GPL-3.0-only",
   "bugs": {
--- a/plugin/_locales/en/messages.json
+++ b/plugin/_locales/en/messages.json
@@ -203,6 +203,10 @@
 		"message": "less tags",
 		"description": "Only extract genre"
 	},
+	"__MSG_label_Manual_Delay_Per_Chapter__": {
+		"message": "Manual Throttle|Delay per chapter",
+		"description": "Manual Throttle|Delay per chapter"
+	},
 	"__MSG_label_Manually_Select_Parser__": {
 		"message": "Manually Select Parser:",
 		"description": "Label for menu to change parser (in advanced options section)"
--- a/plugin/images/ChapterStateSleeping.svg
+++ b/plugin/images/ChapterStateSleeping.svg
@@ -1,0 +1,14 @@
+<svg version="1.1"
+     baseProfile="full"
+     width="32" height="32"
+     xmlns="http://www.w3.org/2000/svg">
+  <line stroke-width="2" y2="4.3" x2="15.1" y1="4.3" x1="6.2" stroke="#965607" />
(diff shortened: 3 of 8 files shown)


dteviot:
@kondeeza
Thank you very much for this, it really is appreciated.
However, there's one significant problem that needs fixing.  I think there will be a merge conflict with the Experimental branch.  (Specifically, the credits in package.json.)
To explain, I try to follow a master + development branch system, where master tracks the releases, and development is where development occurs until it's stable.  Something like https://www.atlassian.com/git/tutorials/comparing-workflows/gitflow-workflow.  Note, for historical reasons, the development branch has been called "Experimental Tab".  Anyway, if you look you will see that Experimental is some way ahead of Master, and conflicts with your changes.

There's a couple of minor things you might like to fix as well.  (If you don't want to fix these, I can do so after the merge.)
1. There's already a sleep() function in Util.js
2. You need to add an entry for ChapterStateSleeping.svg to pack.js (otherwise it won't be included when the extension is built.)
https://github.com/dteviot/WebToEpub/blob/2478bac99781063bf2cbc8175769425317dd369e/eslint/pack.js#L171-L180

kondeeza:
Haha no problem , Thanks for clearing it up. Turns out there was indeed some convention for git workflow, TIL. 

I'll just checkout from experimental branch and do all those including the minor fix. Let me know if there is a problem though because I'm not quite confident with npm tools.  Will just do another PR

PR #482, opened 2021-03-01
Added Feature: Manual Delay/Chapter (+minor refactor)

kondeeza:
### Pull Request Check List

- Do all existing unit tests pass?  `yes. 570 assertions of 570  passed, 0 failed. (Firefox)`
- Have all warnings and errors reported by eslint been fixed? `yes.`
- If you've added new behavior, Can a user turn it off? Is it off by default? `It's off by default, Accessible in Advanced Options , no reason to throttle if it works normally after all`
- Have you updated the "contributors" section of packages.json with your name? (and ideally email, so I can contact you easily.) `yes`
- Have you updated the "Credits" section of readme.md with your name? `yes`
- Are you committing to the ExperimentalTab branch? `yes`
- Have you rebased your commit?  `no`

### What does it do?

Options to Throttle/delay fetching webpage by x seconds/chapter. Accessible via Advanced option (Off by default).

I saw some people in the issue tab had similar issue of wanting to throttle and had to throttle via 'object inspector' as hotfix. I want to make it easier for basic users to throttle speed because even I don't know how to throttle via 'object inspector'.

### Minor correction
used existing util.sleep() instead
added an entry for ChapterStateSleeping.svg to pack.js 
also moved  randomInteger() to Util because it probably belongs there more than in the Parser

### Possible outstanding improvement (or new bug I introduced...)
If manual throttle is set then it also ended up delaying the first chapter fetch as well. It shouldn't need to do that, but the fix I could think of would make the code too messy so eh, this is fine enough.

### Why? isn't error 429, 509  auto exponential throttle enough?

There were some site that just outright block you for a whole day after fetching 50 chapters within an hour (even with no simultaneous fetching), causing the whole packing to fail.  This is useful for users with 24/7 PC who are willing to throttle chapter fetching  as long as epub gets compile eventually.

It's also compatible with simultaneous fetching, so you can get 1 Chapter /3 Secs , 2 Chapter /5 Secs,  8 Chapter /1 Minutes, etc. (Or even adjust speed during the fetching process)

### Screenshots

sleeping_icon

newconfig

**P.S. what would happen to other _locales/ that isn't English ?**

### DIFF
--- a/eslint/pack.js
+++ b/eslint/pack.js
@@ -177,6 +177,8 @@ var packNonManifestExtensionFiles = function(zip, packedFileName) {
     }).then(function () {
         return addImageFileToZip(zip, "ChapterStateNone.svg");
     }).then(function () {
+        return addImageFileToZip(zip, "ChapterStateSleeping.svg");
+    }).then(function () {
         return addCssFileToZip(zip, "default.css");
     }).then(function () {
         return addCssFileToZip(zip, "alwaysDark.css");
--- a/package.json
+++ b/package.json
@@ -40,7 +40,8 @@
     { "name": "Tom Goetz"},
     { "name": "Sergii Pravdzivyi"},
     { "name": "JimmXinu"},
-    { "name": "gamebeaker"}
+    { "name": "gamebeaker"},
+    { "name": "Kondeeza"}
   ],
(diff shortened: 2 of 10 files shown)


dteviot:
Thank you very much.

Issue #485, opened 2021-03-04
lightnovelworld.com changed to lightnovelpub.com


Jenniyah:
the url had changed to a new one lightnovelpub.com and it no longer pulls. Gives error instead

No parser found for this URL. Default Parser will be used. Please specify how to obtain content from each web page.

Thank you

gamebeaker:
@Jenniyah  Hi you can still use it to create epubs from lightnovelpub.com if you copy the link of the novel you want and open a new tab in it you click on webtoepub. Then you paste the link in the starting URL section and changes the link from lightnovelpub.com... to lightnovelworld.com... click load and analyse. The Cover Image is wrong but you can simple change this link.

dteviot:
@Jenniyah
New version 0.0.0.128 has been posted to the Firefox store and is available at https://addons.mozilla.org/en-US/firefox/addon/webtoepub-for-baka-tsuki/
Unfortunately, as I submitted a version of WebToEpub to the Chrome store yesterday, it will be 2 to 3 weeks before I can submit an updated version to the Chrome store.  (And then another 2 to 3 weeks for them to approve.)
If you REALLY want a Chrome version,
Test versions for Firefox and Chrome have been uploaded to https://drive.google.com/drive/folders/1B_X2WcsaI_eg9yA-5bHJb8VeTZGKExl8?usp=sharing. Follow the "How to install from Source" instructions at https://github.com/dteviot/WebToEpub#how-to-install-from-source and let me know how it goes.

PR #486, opened 2021-03-04
lightnovelworld.com changed to lightnovelpub.com

gamebeaker:
Added lightnovelpub.com to lightnovelworld parser. #485

### DIFF
--- a/plugin/js/parsers/LightNovelWorldParser.js
+++ b/plugin/js/parsers/LightNovelWorldParser.js
@@ -1,6 +1,7 @@
 "use strict";
 
 parserFactory.register("lightnovelworld.com", () => new LightNovelWorldParser());
+parserFactory.register("lightnovelpub.com", () => new LightNovelWorldParser());
 
 class LightNovelWorldParser extends Parser{
     constructor() {


dteviot:
@gamebeaker 
Thanks. I do appreciate the effort.  But I'm going to decline it.
Basically, I accept a pull there's an hour of work on my side to sync everything up.
For a one line change like this, It's much less work for me to just add the one line change myself.

Incidentally, while I have you
1. You might like to have a look at https://github.com/dteviot/WebToEpub/commit/f5d5799b42388255a19a4ae445dde054837e4b7e for how to solve that eslint warning
2. I'm wondering why you wrote this
```javascript
let tags = [...dom.querySelector("div .genres-content").querySelectorAll("[rel='tag']")];
```

instead of
```javascript
let tags = [...dom.querySelectorAll("div .genres-content [rel='tag']")];
```

gamebeaker:
@dteviot I wrote let tags = [...dom.querySelector("div .genres-content").querySelectorAll('[rel="tag"]')]; first and after i got eslint errors which was like use " for Strings not ' i tried ("[rel="tag"]") and it didn't worked i didn't knew you have to swap them to ("[rel='tag']") to work that is why i decided to disable eslint. Could you please change this in the template to. I made there the same mistake but it isn't shown in eslint because it is a comment.

dteviot:
@gamebeaker 
I figured that's what you'd done for the eslint error.  
I was more interested in why querySelector() followed by querySelectorAll()?

And I'll fix the template.

gamebeaker:
@dteviot My thought process was like ok i now have my html box with div and .genres-content and now i want to have all genre with rel=tag from inside this box i simple didn't thought of using only one querySelectorAll() with all statements together. I am pretty new to js, html, programming... maybe that's the reason why i simple haven't fought of solving the problem more efficient and easier.
Thanks for pointing this out. I appreciate it.

PR #530, opened 2021-05-08
Add site freewebnovel.com

sickan90:
I'm not a expert on javascript so had a hard time getting the lint to work properly. When i tried to do the 'npm install' on the project I got an error ' No matching version found for qunit@1.19.0'. So this code is not lint checked as i could not get it to work. I have however gotten the novel I wanted to epub to work. So the code works as intended.

### DIFF
--- a/package.json
+++ b/package.json
@@ -42,7 +42,8 @@
     { "name": "JimmXinu"},
     { "name": "gamebeaker"},
     { "name": "Mathnerd314"},
-    { "name": "Kondeeza"}
+    { "name": "Kondeeza"},
+    { "name": "Sickan90"}
   ],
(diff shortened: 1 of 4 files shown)


sickan90:
Ok, found out how to do the lint and fixed a lint error. Now all test are green and no lint errors

dteviot:
Nice.  Thanks very much.

dteviot:
@sickan90 
Please see https://github.com/dteviot/WebToEpub/commit/ee1d39e4f8a36e4c3236a210acb3d12db21d1429
I had to adjust getUrlsOfTocPages, it didn't work for https://freewebnovel.com/monster-integration.html
And I've also set it to add an information page to the epub.

PR #533, opened 2021-05-10
Fix site freewebnovel.com for novels with less than 40 chapters

sickan90:
Fix two things
1) `let lastUrl = dom.querySelectorAll(".page a.lastTocIndex-container-btn")[3];` -> `dom.querySelectorAll(".page a.index-container-btn")[3]` was an error in the previous commit where i did a IDE refactoring of a variable named `index` to `lastTocIndex` which changed the css selector as well. 
2) As I was checking i did some more testing and found that if there are less than 40 chapter it will generate an error. So I did a fix for that as well.

### DIFF
--- a/plugin/js/parsers/FreeWebNovelParser.js
+++ b/plugin/js/parsers/FreeWebNovelParser.js
@@ -18,21 +18,22 @@ class FreeWebNovelParser extends Parser {
 
     static getUrlsOfTocPages(dom) {
         // lastUrl should be example https://freewebnovel.com/<some-novel-name>/<index>.html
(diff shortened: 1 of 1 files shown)


sickan90:
Novel with <40 chapters nobunagas-imouto-is-my-wife

dteviot:
@sickan90 
I decided to do it a little differently. See: https://github.com/dteviot/WebToEpub/commit/a0f95a0dbcb9bacaa92f6a3e7ad7cd5c7e3a2ed5
FYI.  you might want to brush up on await/async/return rules.  https://jakearchibald.com/2017/await-vs-return-vs-return-await/

PR #577, opened 2021-08-18
Update LightNovelWorldParser.js to function with new site structure

Kiradien:
The site configuration for LightNovelWorld and LightNovelPub has changed and requires a separate pageload to show chapter information. Chapters are no longer determined through searchParams and need a defined path.
POST methodology no longer works to request paging, however default setting GET requests work fine.

Changed cover search logic to default to og:image meta data for book cover before the standard utility "FirstImgSrc".

isWatermark logic changed to search for classname since the site formatting only uses classes inside the content area for watermarks and Advertisements. Old Inclusion logic often failed when spaces were applied to the watermark (Fairly common on several hundred chapter stories.) This can be safely/easily reverted and still work most of the time.

### DIFF
--- a/plugin/js/parsers/LightNovelWorldParser.js
+++ b/plugin/js/parsers/LightNovelWorldParser.js
@@ -18,12 +18,6 @@ class LightNovelWorldParser extends Parser{
 
         chapterUrlsUI.showTocProgress(chapters);
         let options = {
-            method: "POST",
-            credentials: "include",
-            headers: {
-                "x-requested-with": "XMLHttpRequest",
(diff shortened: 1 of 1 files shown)


dteviot:
@Kiradien
Thank you.  See https://github.com/dteviot/WebToEpub/commit/a520c3d147b8d57d79ee978006b62adbb3e8fbed for assorted fixes.

dteviot:
@Kiradien
Updated version (0.0.0.136) has been submitted to Firefox and Chrome stores.
Firefox version is available now.
Chrome might be available in 1 to 3 weeks.

PR #585, opened 2021-09-03
Fix error caused by novels on Madara without description

Kiradien:
Novels using this parser where the site has failed to enter any description caused a failure which prevented further generation. Simply defaulted summary to return empty when no description is detected.

Considering how simple this pull request is, feel free to just copy/paste or whatever is simpler rather than merge.

### DIFF
--- a/plugin/js/parsers/MadaraParser.js
+++ b/plugin/js/parsers/MadaraParser.js
@@ -56,7 +56,12 @@ class MadaraParser extends WordpressBaseParser{
     }
 
     extractDescription(dom) {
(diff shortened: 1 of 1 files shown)


dteviot:
@Kiradien 
I'd use a ternary operator.
e.g.
```javascript
        return selectedElement ? selectedElement.textContent.trim() : "";
```
It's going to be easier for me to just do the fix.

dteviot:
@Kiradien
Updated version (0.0.0.139) has been submitted to Firefox and Chrome stores.
Firefox version is available now.
Chrome might be available in 1 to 3 weeks.

PR #595, opened 2021-09-15
Npm install

Dimava:
Simplify installation for possible developers

`npm install` to install
`npm test` to test (with qunit)
`npm run lint` to lint
`npm run build` to pack (without linting)


Dimava:
Note:
JSZip may be replace with npm version
Readme.md may need updates
opening tests may be simplified with http-server

dteviot:
1. I think you need to put jszip back in package.json's dependencies.  (It's used to tell GitHub the project's dependencies, so that Git Hub will automatically notify me if jszip needs to be updated.)
2. I don't fully understand what http-server is doing for testing.  I assume it's https://www.npmjs.com/package/http-server, and is being used to allow browsers access to local files without needing to change their settings.  However, I'd like the notes on configuring the browsers left in please.  (You are welcome to prefix them with a note about running the tests WITHOUT using running http-server.)

dteviot:
And thank you very much for your efforts.

Dimava:
> put jszip back in package (for notifications)

Will do
BTW maybe include a copy of `jszip.min.js` to remove "download jszip zip and unpack it" requirement?
(it is actually a dependency on git, so it ma be replaced with "run `git submodule update`", or with downloading repo through GithubDesktop)

> I don't fully understand what http-server is doing for testing. I assume it's https://www.npmjs.com/package/http-server, and is being used to allow browsers access to local files without needing to change their settings. However, I'd like the notes on configuring the browsers left in please. (You are welcome to prefix them with a note about running the tests WITHOUT using running http-server.)

It just serves local files on `localhost`. No matter which server is used, `http-server`, [chrome Web Server app](https://chrome.google.com/webstore/detail/web-server-for-chrome/ofhbbkphhbklhfoeikjpcbhemlocgigb), some php server, <https://raw.githack.com/>, whatever, it is just used to load files over `http:` instead of `file:` (which is unsafe, browsers have those limitations on `file:` for a reason)

Dimava:
added `npm run up` that updates jszip to latest

dteviot:
@Dimava  
Thank you very much.  This was educational.

PR #632, opened 2021-11-11
Fix downloads when using Firefox for Android

### LINKED ISSUE #480: Fix Addon for Firefox Fenix
Background
-Firefox Android changed from Fenic to Fenix in August 2020, which disabled all except a limited whitelist of addons
-However, you can now still install addons on Firefox Fenix in the Firefox Nightlys through the collection feature, and Webtoepub does install (how to here: https://www.ghacks.net/2020/10/01/you-can-now-install-any-add-on-in-firefox-nightly-for-android-but-it-is-complicated/)

Issue
-While webtoepub runs, and the status bar shows all the chapters being completed, the final step of downloading the file does not happen.
-Storage permissions for Firefox Nightly are on and I have not seen any settings in the Firefox Nightly app itself that might cause this issue and tested that I am able to download files generally in Firefox Nightly.

Requesting a fix so the addon continues to work.  Thanks.

Lej77:
This change allows using the extension on Firefox Fenix (the newer "Firefox for Android" browser). 

It seems like the `browser.downloads.download` API isn't implemented yet in Fenix so when it is called nothing happens. No error is thrown but the promise never resolves either. As a workaround I used a temp `a` tag to initiate the download as suggested at [Download File Using JavaScript/jQuery - Stack Overflow](https://stackoverflow.com/questions/3749231/download-file-using-javascript-jquery/9834261#9834261).

I also improved the Android detection using the [`browser.runtime.getPlatformInfo`](https://developer.mozilla.org/docs/Mozilla/Add-ons/WebExtensions/API/runtime/getPlatformInfo) API. This isn't required but it is very useful since it works even when the extension page is opened with the "Reqeust Desktop Site" toggle enabled. When I tested my changes with a story from https://www.fanfiction.net/ I noticed that the addon would try to download chapters from https://www.m.fanfiction.net/ (notice the m) even though I specified URLs without the "m". It [seems that the server is looking at the `User-Agent` header](https://stackoverflow.com/questions/11702453/how-does-chromes-request-desktop-site-option-work) and returning a redirect response (to the `fetch` method). If the extension web page is also opened with the "Reqeust Desktop Site" toggle enabled then the `User-Agent` for the fetch requests is overridden and they are no longer redirected. I suppose one could maybe [manually override the `User-Agent`](https://stackoverflow.com/questions/42815087/sending-a-custom-user-agent-string-along-with-my-headers-fetch) in this extension's code but I think it is simpler to leave that work to the browser. It would be nice with some better error messages for these cases (tell the user to try the "Reqeust Desktop Site" toggle for the extension's web page and warn about URLs that are redirected to a mobile site).

Fixes #480.

### DIFF
--- a/plugin/js/Download.js
+++ b/plugin/js/Download.js
@@ -64,14 +64,44 @@ class Download {
     }
 
     static saveOnFirefox(options, cleanup) {
-        if (Download.isAndroid()) {
-            options.saveAs = false;
-        }
-        return browser.downloads.download(options).then(
-            // on Firefox, resolves when "Save As" dialog CLOSES, so no
-            // need to delay past this point.
-            downloadId => Download.onDownloadStarted(downloadId, cleanup)
-        ).catch(cleanup);
+        return browser.runtime.getPlatformInfo().then(platformInfo => {
+            // If the extension's web page is opened with the
+            // "Reqeust Desktop Site" toggle enabled then only the second check
+            // will correctly detect android.
+            //
+            // This is important to support since fetch requests can be
+            // redirected to mobile versions of web sites when the toggle is
+            // disabled.
+            if (Download.isAndroid() || platformInfo.os.toLowerCase().includes("android")) {
+                try {
+                    options.saveAs = false;
+
+                    // `browser.downloads.download` isn't implemented in
+                    // "Firefox for Android" yet, so we starts downloads
+                    // the same way any normal web page would do it:
+                    const link = document.createElement("a");
+                    link.style.display = "hidden";
+
+                    link.href = options.url;
+                    link.download = options.filename;
+
+                    document.body.appendChild(link);
+                    try {
+                        link.click();
+                    } finally {
+                        document.body.removeChild(link);
+                    }
+                } finally {
+                    cleanup();
+                }
+            } else {
+                return browser.downloads.download(options).then(
+                    // on Firefox, resolves when "Save As" dialog CLOSES, so no
+                    // need to delay past this point.
+                    downloadId => Download.onDownloadStarted(downloadId, cleanup)
+                );
+            }
+        }).catch(cleanup);
     }
 
     static isAndroid() {
--- a/plugin/js/Download.js
+++ b/plugin/js/Download.js
@@ -73,27 +73,24 @@ class Download {
             // redirected to mobile versions of web sites when the toggle is
             // disabled.
             if (Download.isAndroid() || platformInfo.os.toLowerCase().includes("android")) {
+                options.saveAs = false;
+
+                // `browser.downloads.download` isn't implemented in
+                // "Firefox for Android" yet, so we starts downloads
+                // the same way any normal web page would do it:
+                const link = document.createElement("a");
+                link.style.display = "hidden";
+
+                link.href = options.url;
+                link.download = options.filename;
+
+                document.body.appendChild(link);
                 try {
-                    options.saveAs = false;
-
-                    // `browser.downloads.download` isn't implemented in
-                    // "Firefox for Android" yet, so we starts downloads
-                    // the same way any normal web page would do it:
-                    const link = document.createElement("a");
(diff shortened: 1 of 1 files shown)


dteviot:
I'll say thank you very much.
There's a few minor points:
1. Should commit against the ExperimentalTabBranch (that's the development branch.  master tracks releases.)
2. You didn't add yourself to the credits
3. Could probably remove isAndroid(), as the PlatformInfo replaces it.
4. I personally would pull the Android download logic into it's own function, instead of nesting inside the existing function.

However, I'll accept with thanks and clean up myself

dteviot:
@Lej77 
These are the amendments. https://github.com/dteviot/WebToEpub/commit/909f7749e25f4f4728d2dd35c2bce475bdb8a540
If you can confirm I haven't broken Firefox for Android (as I'm not set up for testing), I'll make a new build and submit to the Firefox store.

Lej77:
@dteviot Oops, should have read that `CONTRIBUTING.md` file. Thanks for fixing those things for me! I also think the code looks nicer after your changes.

I also tested the changes on the `ExperimentalTabBranch` and downloads seems to be working without any problems.

I did my testing by:
- install `adb` via `scoop` (or some other way I guess)
- install `web-ext` globally via `npm install --global web-ext`
- activate USB-debugging on my Android Phone
- activate USB-debugging on Firefox Nightly's setting page on my phone
- enable the file permission for Firefox Nightly on my phone
- connect my phone to my PC via a USB cable
  - Then press ok on the popup that showed up on my phone to ensure I was authorized
- cloned your project
- run `npm install`
- run `npm build`
- unpack `./eslint/WebToEpub.VERSION.xpi` via `7zip` to a new folder.
- cd a shell into that folder
- run `web-ext run -t firefox-android --adb-device XXX --firefox-apk org.mozilla.fenix` (from [here](https://extensionworkshop.com/documentation/develop/developing-extensions-for-firefox-for-android/#install-and-run-your-extension-in-firefox-for-android)) and replaced `XXX` with my phone's id which I got by running `adb devices`.
- Firefox Nightly opened on my Phone and I was free to test the extension
- I could also debug it via my PC by opening Firefox Nightly and going to the `about:debugging` page, then enabling USB features on that page and lastly clicking the connect button next to the device listed in the left sidebar.

dteviot:
@Lej77
FYI, updated build is now on Mozilla store.

Lej77:
Thanks for letting me know!

I followed the instructions at [Expanded extension support in Firefox for Android Nightly | Mozilla Add-ons Community Blog](https://blog.mozilla.org/addons/2020/09/29/expanded-extension-support-in-firefox-for-android-nightly/) and managed to install the version in the store on Firefox Nightly for my phone and the addon worked when I used it. 

I did have to toggle the web site ([`fanfiction.net`](https://www.fanfiction.net/)) to `Reqeust Desktop Site` for the default parser to work (since otherwise the URL had the [`m.`](https://m.fanfiction.net/) at the start) and then I used the same toggle on the opened WebToEpub page so that it could download the correct chapter URLs. But that is probably the best solution anyway since the desktop versions of web pages will be the most tested for the parsers so I don't think this needs fixing.

dteviot:
@Lej77 
> I did have to toggle the web site (fanfiction.net) to Reqeust Desktop Site for the default parser to work 

I forgot about that.  unfortunately, the m. version of site is a pain.  It doesn't seem to include a list of all chapters for a story.

Lej77:
I just wanted to mention it but really I think the best solution is to just use the desktop versions of web pages since most people will use the extension from their PC and therefore the default parsers will work much better for those versions. Its not really hard to workaround with the `Reqeust Desktop Site` toggle as long as you know that it exists.

crypticc23:
@Lej77 Thanks for fixing this!

Lej77:
Your welcome! I wanted this for my own use anyway, but its good that there are more people that can take advantage of it!

PR #636, opened 2021-11-16
Move Tab Logic to Background Script

Synteresis:
This is more of a proof of concept than an actual pull request. 

There are some leftover functions that I accidentally left in from trying to maneuver around CORS on IOS Safari. I also accidentally left in "incognito": "spanning" for testing on Firefox. 

Is there a reason this extension doesn't use something like webpack with its crossbrowser support? More of a rhetorical question, I haven't learned enough to understand the difference that it would make. I only know that trying to import different node modules didn't work out so smoothly because of it.

If you are fine with moving the tab logic to the background script, this fixes a bug on Safari MacOS and IOS that causes it to either flicker on startup and not work, or open a new tab while opening the popup on top of the extension tab.

### DIFF
--- a/plugin/background.js
+++ b/plugin/background.js
@@ -1,0 +1,44 @@
+function getActiveTab() {
+    return new Promise(function (resolve, reject) {
+        chrome.tabs.query({currentWindow: true, active: true}, function (tabs) {
+            if ((tabs != null) && (0 < tabs.length)) {
+                resolve(tabs[0].id);
+            } else {
+                reject();
+            }
+        });
+    });
+}
+
+function openTabWindow() {
+    // open new tab window, passing ID of open tab with content to convert to epub as query parameter.
+    getActiveTab().then(function (tabId) {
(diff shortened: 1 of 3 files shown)


plugin/background.js
```diff
@@ -0,0 +1,44 @@
...
+    });
+}
+
+chrome.browserAction.onClicked.addListener(
+    function () {
+        openTabWindow();
+    }
+);
+
+chrome.runtime.onMessage.addListener(function (request, sender, sendResponse) {
```
Synteresis, line 29: Used to have a runtime sendmessage in the downloads that wrapped the fetch call to go through here and return the response. The header ended up being null everytime.

dteviot:
@Synteresis 
> Is there a reason this extension doesn't use something like webpack with its crossbrowser support
If I use webpack, the extension requires a more extensive (and time consuming) audit process before Mozilla will allow it to be in the store.

PR #638, opened 2021-11-16
Simple Proof of Concept for Saving on Safari

Synteresis:
Unfortunately, I don't believe there is any support for a location/folder picker as of yet. Many different libraries all had the same result of just saving through legacy fallback on Safari.

There was also some sort of popup blocker that didn't let "_blank" do its job and kept reloading the page causing the blob to get collected and for the download to fail. I thought it was CORS for a good while until I finally realized.

util.isFirefox() is true on both Firefox and Safari. Only Firefox supports and has browser.runtime.getBrowserInfo() so if other browsers start to pick it up, it'll also become a false positive.

It's a bit late for me right now, so I don't have the time to check, but I may or may not have tested the new browser detection logic in the other two browsers. I wish there was an automatic way to compile directly to safari extension instead of having to use xcode.

### DIFF
--- a/plugin/js/Download.js
+++ b/plugin/js/Download.js
@@ -1,14 +1,16 @@
 "use strict";
 
 class Download {
-    constructor () {
+    constructor() {
     }
 
     static init() {
         Download.saveOn = util.isFirefox() ? Download.saveOnFirefox : Download.saveOnChrome;
-        if (util.isFirefox()) {
+        if (typeof browser.runtime.getBrowserInfo === "function") {
             Download.saveOn = Download.saveOnFirefox;
             browser.downloads.onChanged.addListener(Download.onChanged);
+        } else if (util.isFirefox()) {
+            Download.saveOn = Download.saveOnSafari;
         } else {
             Download.saveOn = Download.saveOnChrome;
             chrome.downloads.onChanged.addListener(Download.onChanged);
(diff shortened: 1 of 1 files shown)


dteviot:
A couple of points.
1. I'm not sure this code is correct.

2. You've reformatted a number of lines, but they're functionally unchanged.  e.g. line 4, you've changed the spacing.
Please don't do that.  It makes it more difficult to figure out what the change really is.
And I won't accept Pull requests that do that.

Synteresis:
> @Synteresis A couple of points.
> 
> 1. I'm not sure this code is correct.
> 
> ```js
> } else if (util.isFirefox()) {
>             Download.saveOn = Download.saveOnSafari;
> ```
> 
>  1. You've reformatted a number of lines, but they're functionally unchanged.  e.g. line 4, you've changed the spacing.
>    Please don't do that.  It makes it more difficult to figure out what the change really is.
>    And I won't accept Pull requests that do that.

Sorry about the reformat, I probably hit the shortcut by habit. For the Firefox function, I'm not sure why but it consistently works on all Safari browsers for me, both Mac and IOS.

dteviot:
@Synteresis 

```javascript
} else if (util.isFirefox()) {
            Download.saveOn = Download.saveOnSafari;
```

I think you misunderstood me.
The logic seems strange if I read the code.
Seems to say "If this is firefox, use safari".

Synteresis:
> @Synteresis
> 
> ```js
> } else if (util.isFirefox()) {
>             Download.saveOn = Download.saveOnSafari;
> ```
> 
> I think you misunderstood me. The logic seems strange if I read the code. Seems to say "If this is firefox, use safari".

Ah, I understand you now. I will find a more specific method of detecting Safari and add it to the ever-growing util functions.

PR #644, opened 2021-11-20
Implement pagestage.kakao.com Parser

### LINKED ISSUE #639: Please add site https://pagestage.kakao.com/
I tried using the Default Parser for the site. But I don't have any basic knowledge of JavaScript and HTML, So i failed

I tried
https://pagestage.kakao.com/novels/22514217?page=5

* URL of first chapter: https://pagestage.kakao.com/novels/22514217/episodes/484
* CSS selector for element holding content to put into EPUB: div.sc-bdnylx.eUOjDh
* CSS selector for element holding Title of Chapter: h2.sc-bdnylx.djRsjE
* CSS selector for element(s) to remove

It didn't work, and i don't know what to do.

Can you add site 'https://pagestage.kakao.com/' ?

Synteresis:
This adds the requested parser and closes #639 

Finally had some time. 

Wondering about the alternative to a problem I had. Because all the data was passed to the page through GET calls, I wanted to modify the TOC dom so that when get author, genre, and such was called, it would be able to access the information directly from the api. I didn't see a way to modify that dom. Is there currently no method or is there an alternative I didn't think of? I'll look through the other parsers to see what they did.

This time, the page luckily came with the data pre-javascript execution so I was able to get the data from there.

### DIFF
--- a/plugin/js/parsers/KakaoParser.js
+++ b/plugin/js/parsers/KakaoParser.js
@@ -1,0 +1,125 @@
+"use strict";
+
+parserFactory.registerUrlRule(
+    url => KakaoParser.isValidUrl(url),
+    () => new KakaoParser()
+);
+
+class KakaoParser extends Parser {
+    constructor() {
+        super();
+    }
+
+    static isValidUrl(url) {
+        // Return true when is on the table of contents page.
+        // Perhaps narrow this down so it can't execute on the chapters, but needs
+        // to execute on the chapter api
+        if (util.extractHostName(url).includes("api-pagestage.kakao.com")) {
+            return true;
+        }
+        if (util.extractHostName(url).includes("pagestage.kakao.com")) {
+            if(!url.includes("episode")) {
+                if(url.includes("novel")) {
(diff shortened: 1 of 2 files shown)


dteviot:
Thanks very much
A couple of minor points.
1. I needed to fix eslint warnings.  See: https://github.com/dteviot/WebToEpub/commit/d90e6f15c2c3d18373923dd79addd4e332de49e1
2. lines 20 to 27 could be simplified.  e.g.
``` javascript
return util.extractHostName(url).includes("pagestage.kakao.com"))
    && !url.includes("episode")
    && url.includes("novel")
```
3. My preference, when doing a commit for a requested site:
I find a better title is "Add site XXXX", rather than name of parser, and
I like to include a link to the request issue in the details notes.

Synteresis:
1. Oh! I was on stackoverflow for finding how querySelector works. I didn't know that HTML also accepted both single and double quotes.
2. Haha, I didn't think of this. Thank you.
3. I will refer to this when I make a commit in the future.

PR #654, opened 2021-11-28
Update XbiqugeParser.js

Hogesyx:
The site has change main domain to .so extension, .cc is now just a forwarder.


dteviot:
Thank you.

PR #659, opened 2021-11-30
Update to H1 H2 CSS

Hogesyx:
After recent updates to iOS and Android, page-break-before causes some weird behavior in some ebook reader.

Specifically, it render an entire blank page on some reader(iOS Book), or break text to speech on some other reader(eink android reader, unknown engine).

I have tested and confirm that it is indeed this tag that causes the issue using Calibre to rebuild the epub.

In most cases, this tag is redundant and alternatively we can try out other method if page break is really required(eg page-break-after on p tag)


dteviot:
Thanks.  I'm not sure why the page break was there in the first place.  I didn't create the CSS.

PR #661, opened 2021-12-01
Add support for PG-15 chapters from Kakao.com

### LINKED ISSUE #648: Kakaoparser doesn't work in PG 15 novels.
**Describe the bug**
'https://pagestage.kakao.com/'
I tried to extract the epub file from the above site with the added version of kakaoparser, but found that the program error occurred in the PG 15 novels on the above site.

**To Reproduce**
1. Go to 'https://pagestage.kakao.com/novels/95593310'
2. In order to read PG 15 novels, you must log in to the site's Adult Verified ID. So I made a temporary ID for you.

3. Launch WebToEpub.
4. Error.

Error: Fetch of URL 'https://api-pagestage.kakao.com/novels/16625636' failed with network error 403.
at FetchErrorHandler.onResponseError (chrome-extension://knkobmaemfjhaigiabijmncandjdblli/js/HttpClient.js:32:25)
at Function.checkResponseAndGetData (chrome-extension://knkobmaemfjhaigiabijmncandjdblli/js/HttpClient.js:166:45)
at chrome-extension://knkobmaemfjhaigiabijmncandjdblli/js/HttpClient.js:160:31
at async KakaoParser.getChapterUrls (chrome-extension://knkobmaemfjhaigiabijmncandjdblli/js/parsers/KakaoParser.js:82:21)

'https://pagestage.kakao.com/novels/16625636'
Since the same error occurs at the above address, I am guessing that an error appears in all PG 15 novels.

**Desktop (please complete the following information):**
- OS: Windows
- Browser : Chrome
- Version : 0.0.0.144

Synteresis:
I was unfortunately forced to learn promises and could no longer chain awaits 😢.

I messed up my git completely and was stuck to commit this fork a few commits behind. For some reason, rebasing, refetching, remerging, hard undoing commits, etc, would not correct the git history to match with ExperimentalTabMode. Thankfully, since it is behind, I was able to pull request just the two commits.

Required to be logged in with a verified account to download PG-15 books.

Closes #648

-----

KNOWN BUGS:
Start on the main page when using WebToEpub. Having ?page=[num] in the url can and will break the rest of the code.
Adding "&& !url.includes("page")" will cause it to break as well. Requires specific if and else cases within wrapFetch to clean URLs. Ugly. Much better to start on main page.

Written on a MacBook. I ran "npm run lint -- --fix" and it fixed all the line ending errors, but if it occurs, this is why.

### DIFF
--- a/eslint/.eslintrc.js
+++ b/eslint/.eslintrc.js
@@ -2,7 +2,7 @@ module.exports = {
     "env": {
         "browser": true,
         "es6": true,
-		"webextensions": true,
+        "webextensions": true,
     },
     "parserOptions": {
         "ecmaVersion": 2020
--- a/plugin/js/parsers/KakaoParser.js
+++ b/plugin/js/parsers/KakaoParser.js
@@ -10,6 +10,35 @@ class KakaoParser extends Parser {
         super();
     }
 
+    async setAuthorizationToken() {
+        if(typeof this.token === "undefined") {
+            const jsonUrl = "https://api-pagestage.kakao.com/users/login";
+
+            const fetchOptions = {
+                method: "POST",
+                credentials: "include"
+            }
+
+            return HttpClient.fetchJson(jsonUrl, fetchOptions);
+        }
+    }
+
+    async wrapFetch(jsonUrl){
+        return this.setAuthorizationToken().then(jsonResponse => {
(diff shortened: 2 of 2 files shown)


dteviot:
Thanks

PR #913, opened 2023-02-08
Open tab as child of current tab

### LINKED ISSUE #702: Open tab as child of current one
**Is your feature request related to a problem? Please describe.**
I habitually have dozens or hundreds of tabs open (using Tree-Style Tab). When I open WebToEpub on a given site, it's opened at the end of the tab list, making it annoying to switch back and forth between the WebToEpub tab and the original.

**Describe the solution you'd like**
It would be good to open the WebToEpub tab as a child of the current tab (i.e. the same way it behaves when you open a link in a new tab). In Tree-Style Tab, this makes the new tab a literal child of the current one; in the default tab interface, it just opens the tab directly adjacent to the current one rather than at the end of the tab list. Either way makes it much easier to handle WebToEpub.

alethiophile:
Fixes #702.

This simply makes the web2epub tab open as a child of the current one, which means it's directly after the current tab, rather than at the end of the tab list.

### DIFF
--- a/plugin/js/main.js
+++ b/plugin/js/main.js
@@ -281,7 +281,7 @@ var main = (function () {
         getActiveTab().then(function (tabId) {
             let url = chrome.runtime.getURL("popup.html") + "?id=";
             url += tabId;
-            chrome.tabs.create({ url: url });
+            chrome.tabs.create({ url: url, openerTabId: tabId });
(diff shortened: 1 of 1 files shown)


dteviot:
Next time, commit to experimentalTab branch, not master.

Issue #936, opened 2023-02-28
0.0.0.152 appears to be completely nonfunctional on Fennec 110.0.1


LongFluffDragon:
**Describe the bug**
Opening the addon from any table of contents page fails to load. The addon is unresponsive, shows only placeholder labels, with __MSG_Searching_For_URLs_Please_Wait__ displayed. It never times out and all buttons are nonfunctional.

**To Reproduce**
Open https://www.royalroad.com/fiction/31429/cinnamon-bun or https://wanderinginn.com/table-of-contents/
Open Addons -> WebToEpub
Observe failure to load

**Expected behavior**
Less failure to load. Both sites worked fine with this addon in the same browser prior to the update

**Screenshots**
Resulting page: https://i.imgur.com/jiEGtXY.png

**Smartphone (please complete the following information):**
 - Device: Pixel 3A XL
 - OS: Android 11
 - Browser Fennec F-Droid
 - Version 110.0.1

**Additional context**
The latest version of the addon continues to work fine in the latest desktop firefox. Unable to test on tablet devices due to browser UI issues.

bootless-corpse:
Same issue using Firefox nightly

dteviot:
@bootless-corpse 
Is this desktop or mobile Firefox?

bootless-corpse:
Android Firefox

gamebeaker:
@bootless-corpse can you please share the "Custom Add-on collection" you use.

bootless-corpse:
I'm not sure 100% how to do that. I made my own collection with just the web to epub add-on. I think it's might be this 17305934 but let me know if I'm wrong lol.

gamebeaker:
@dteviot i was able to debug it on firefox android. If the commit fe8e8cc is reverted it works again. 
@bootless-corpse @LongFluffDragon if this doesn't get fixed i would recomend using kiwi browser or other chromium based mobile browser.

dteviot:
@gamebeaker 
Thanks for the fixes.  

@bootless-corpse @LongFluffDragon 
Updated version (0.0.0.153) has been submitted to Firefox store and is available now.

Issue #933, opened 2023-02-26
Fix warnings raised by Mozilla


dteviot:
Unsafe assignment to innerHTML
Warning: Due to both security and performance concerns, this may not be set using dynamic values which have not been adequately sanitized. This can lead to security issues or fairly serious performance degradation.

js/Library.js line 324 column 13
Unsafe assignment to innerHTML
Warning: Due to both security and performance concerns, this may not be set using dynamic values which have not been adequately sanitized. This can lead to security issues or fairly serious performance degradation.

js/Library.js line 436 column 9
Unsafe assignment to innerHTML
Warning: Due to both security and performance concerns, this may not be set using dynamic values which have not been adequately sanitized. This can lead to security issues or fairly serious performance degradation.

js/Library.js line 533 column 9
Unsafe assignment to innerHTML
Warning: Due to both security and performance concerns, this may not be set using dynamic values which have not been adequately sanitized. This can lead to security issues or fairly serious performance degradation.

js/Library.js line 650 column 9

dteviot:
@gamebeaker 
Can you look into these?  Mozilla hasn't rejected it outright yet, but it may not get approved with them.

dteviot:
@gamebeaker 
Mozilla have approved, but I'd like it fixed for next commit.  Basically, need to build the XML using createElement() and similar calls.  See [o EpubPacker.js](https://github.com/dteviot/WebToEpub/blob/master/plugin/js/EpubPacker.js) for examples

gamebeaker:
@dteviot i am going to look into it.

gamebeaker:
@dteviot how were you able the get these warnings? If i load WebToEpub in firefox for debugging i don't see warnings. I just want to be able to check if my fix works.

gamebeaker:
@dteviot i was able to get the warnings with the command "web-ext lint". How to install web-ext is described here https://extensionworkshop.com/documentation/develop/getting-started-with-web-ext/

dteviot:
@gamebeaker 
Thank you for that

PR #937, opened 2023-03-02
Fix firefox adroid bug and mozilla warnings 

gamebeaker:
Fix firefox android bug #936 
Fix mozilla warnings #933

### DIFF
--- a/plugin/js/Library.js
+++ b/plugin/js/Library.js
@@ -255,6 +255,7 @@ class Library {
             let LibTemplateMergeUploadButton = "";
             let LibTemplateEditMetadataButton = "";
 
+            LibRenderString += "<div class='LibDivRenderWraper'>";
             if (!util.isFirefox()) {
                 let LibTemplateLibraryUses = document.getElementById("LibTemplateLibraryUses").innerHTML;
                 LibRenderString += LibTemplateLibraryUses;
@@ -318,10 +319,11 @@ class Library {
                 LibRenderString += "</tbody>";
                 LibRenderString += "</table>";
                 if (ShowAdvancedOptions) {
-                    LibRenderString += "<table id='LibRenderMetadata"+CurrentLibKeys[i]+"'></table>";
+                    LibRenderString += "<div id='LibRenderMetadata"+CurrentLibKeys[i]+"'></div>";
                 }
             }
-            LibRenderResult.innerHTML = LibRenderString;
+            LibRenderString += "</div>";
+            Library.AppendHtmlInDiv(LibRenderString, LibRenderResult, "LibDivRenderWraper");
             if (ShowAdvancedOptions) {
                 document.getElementById("libdeleteall").addEventListener("click", function(){Library.Libdeleteall()});
                 document.getElementById("LibUploadEpubLabel").addEventListener("mouseover", function(){Library.LibMouseoverButtonUpload(this)});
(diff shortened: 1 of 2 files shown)


dteviot (commented):
I think the following might be a cleaner solution

getActiveTab().then(function (tabId) {
    let args = {url: chrome.runtime.getURL("popup.html") + "?id=" + tabId};
    if (!util.isFirefox()) {
        args.openerTabId = tabId;
    }
    chrome.tabs.create(args);
     window.close();
});

gamebeaker:
@dteviot your version would treat firefox desktop different than chrome desktop in tabs.create(). Firefox desktop doesn't have a problem with tabs.create(). 
Only firefox android does have the problem with the second arg in tabs.create(). Something like util.is.FirefoxAndroid() is needed but i didn't find a difference in js between firefox desktop and firefox android.

dteviot:
@gamebeaker 
I wrote the code in haste.  Should have been
```javascript
        getActiveTab().then(function (tabId) {
            let createProperties = { url: chrome.runtime.getURL("popup.html") + "?id=" + tabId };
            if (!util.isFirefox() || !Download.isAndroid(platformInfo)) {
                createProperties.openerTabId = tabId;
            }
            chrome.tabs.create(createProperties);
            window.close();
        });
```

dteviot:
As regards the second fix, I think you're violating the intent of the check.
Which is to avoid getting malicious HTML injected into into your document, similar to a SQL injection attack.
I don't see your code checking things like
```javascript
LibSaveMetadataString += "<dc:title>"+LibTitleInput+"</dc:title>";
```
To make sure LibTitleInput doesn't contain anything that might cause problems.  e.g. HTML fragments.
I suspect it's probably safe, as the input is expected to be an EPUB that was already created by WebToEpub.
But what if it's not?
To be fair, invalid content will probably just crash WebToEpub, and I have trouble seeing how someone might craft HTML to exploit what's being done.  So, as I said, it's probably safe.  But I'm not willing to spend the time to minutely examine the code to re-assure myself of this.

On the third hand, I accepted this code in the first place.  And the change shuts up the Mozilla warning.
So, I'll accept it as is for the moment.  And rewrite it to something I'm comfortable with when I have time and inclination.  i.e. Probably never.

So, thanks again for your efforts.

gamebeaker:
@dteviot i tried js injection but couldn't get it to work.
You need an epub in the Library.
Library ->  check "Show Advanced Library Options2 -> Edit Metadata -> Paste in Tags textarea: </textarea><textarea> 
-> Save Metadata -> wait until finished -> Edit Metadata  -> console throws error "popup.html:1 Refused to execute inline event handler because it violates the following Content Security Policy directive: "script-src 'self'". Either the 'unsafe-inline' keyword, a hash ('sha256-...'), or a nonce ('nonce-...') is required to enable inline execution. Note that hashes do not apply to event handlers, style attributes and javascript: navigations unless the 'unsafe-hashes' keyword is present."
and
"Refused to execute inline event handler because it violates the following Content Security Policy directive: "script-src 'self' 'wasm-unsafe-eval' 'inline-speculation-rules' http://localhost:* http://127.0.0.1:*". Either the 'unsafe-inline' keyword, a hash ('sha256-...'), or a nonce ('nonce-...') is required to enable inline execution. Note that hashes do not apply to event handlers, style attributes and javascript: navigations unless the 'unsafe-hashes' keyword is present."
If you paste in Tags textarea: </textarea><iframe src="https://www.example.com"></iframe><textarea>
an iframe is loaded.
Regardless if it is a secrurity problem or not i am going to sanitize it because if you change the Autor or any other input field to: Web'To'Epub ->only Web is shown in the field.(dont' forget to rerender if you change filname or Story URL)

gamebeaker:
@dteviot i have a question about your example 
`LibSaveMetadataString += "<dc:title>"+LibTitleInput+"</dc:title>";`
This code snippet is part of a function which gets written in the epub. Do you mean it should be checked what is saved in an epub? If thats the case is such a check implemented for the normal parser?

dteviot:
@gamebeaker 
> if that's the case is such a check implemented for the normal parser?

https://github.com/dteviot/WebToEpub/blob/ea41bfe8e927d84182930196d573858bc4f8aaca/plugin/js/EpubPacker.js#L97

https://github.com/dteviot/WebToEpub/blob/ea41bfe8e927d84182930196d573858bc4f8aaca/plugin/js/EpubPacker.js#L359-L366

PR #939, opened 2023-03-04
Sanitize HTML generation from Library

gamebeaker:
Sanitize HTML generation like mentioned in #937.


gamebeaker:
@dteviot i try to sanitize epub creation.

PR #985, opened 2023-06-06
Parser for Novelusb.com

ImLJS:
Created parser for novelusb.com in plugins->js->parsers
Updated popup.html with link
Updated readme with contributers name


xRahul:
no need for separate parser here as the code is shared with novelbin

PR #1046, opened 2023-08-19
Update PtwxzParser.js

Hogesyx:
Domain change update. Not sure if we should just drop this js and create a new js or update existing js.


dteviot:
Thank you

PR #1052, opened 2023-08-22
Add author extract for mtl novel

xRahul:
en website has the id


package.json
dteviot, line 1: Sorry to be picky, but do you want to fix the closing square bracket you seem to have moved by mistake?

xRahul: Done

PR #1074, opened 2023-09-13
Fix: fixLazyLoadImageSource when image source in data-src attribute

toptaran:
Some websites have image path in data-src attribute.
Example:
``


dteviot:
Thank you very much.

PR #1108, opened 2023-11-04
Update KemonopartyParser.js

solively:
url was changed from kemono.party to kemono.su, any kemono.party link will automatically redirect to kemono.su, so updated the code accordingly.


dteviot:
@solively 
You should update package.json and readme.md with your name, so you get credit for the work.
But I can't be bothered re-doing it for one line.  So, I've just stolen your work and committed it.
And updated test build at https://drive.google.com/drive/folders/1B_X2WcsaI_eg9yA-5bHJb8VeTZGKExl8?usp=sharing

PR #1119, opened 2023-11-23
Fix #1118 - Update FanFictionParser.js

### LINKED ISSUE #1118: Inconsistent caching issue with Fanfiction.Net
As of an apparent server maintenance a few weeks ago, Fanfiction.net has had inconsistent caching failures resulting in a "Chapter not found" error. An alternative link exists to access data when the cache is not loaded, however it can cause issues if constantly accessed directly.

- Consistent reproduction is difficult, as this is a caching issue on the server end. Suggested fix using shortened Uri forces cache to engage within 15 minutes.
- Random novels will be "missing" a chapter when stepping through.

**Additional context**
Replacing url with /(https?:\/\/(?:www\.)?\w+\.\w+\/s\/\d+\/\d+)\/[a-z\-0-9]+/i[1] resolves to base page and updates server cache. Constant replacement of url for every page load puts extra load on server.

Chapters added when Fanfiction.net was having server issues appear to have a higher consistency of cache failures. Example chapter: https://www.fanfiction.net/s/14128289/69/A-Darker-Path
Appearances may be deceiving, not 100% confident on this issue.

Have a fix at https://github.com/dteviot/WebToEpub/commit/123e3e3563cc63da34eed996beab81b2164b0294 - still devtesting. It worked without the rateLimitDelay repeat, however have been unable to get a consistent second test yet.

Kiradien:
Fix for loading inconsistent caching issue on ff.net; automatically checks secondary link upon any failure to retrieve chapter or read chapter text. Separate link calls info directly without interacting with broken caching. Uses existing throttle logic for second call to avoid CloudFlare detection. 

Note: The detection could be narrowed down more to detect page layout and specific error message; This is not done because this method does actually resolve several other rare chapter load issues that have occurred in the last year.

### DIFF
--- a/plugin/js/parsers/FanFictionParser.js
+++ b/plugin/js/parsers/FanFictionParser.js
@@ -61,6 +61,33 @@ class FanFictionParser extends Parser {
         return dom;
     }
 
+    async fetchWebPageContent(webPage)
+    {
+        try
+        {
+            return await super.fetchWebPageContent(webPage);
+        }
+        catch(ex)
+        {
+            //Determine if path contains extra parameters. Immediately fail if already shortened.
+            //Shortened URI is not always ideal solution; apparently related to caching on server. 
+            let regex = /(https?:\/\/(?:www\.)?\w+\.\w+\/s\/\d+\/\d+)\/[a-z\-0-9]+/i
+            let shortUri = regex.exec(webPage.sourceUrl);
+            if (shortUri)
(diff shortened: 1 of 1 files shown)


Kiradien:
This is a fix for bug https://github.com/dteviot/WebToEpub/issues/1118 - not sure why it didn't link correctly; must have been a check-in issue.

PR #1171, opened 2024-01-20
KemonopartyParser don't download unwanted images

### LINKED ISSUE #1168: Unwanted images still downloaded and added into epub
When you use `Parser.removeUnwantedElementsFromContentElement()` on any element containing an image, the image still gets downloaded and added into the epub, even though it's not shown in any pages.

**To Reproduce**
1. Go to https://kemono.su/patreon/user/22614979
2. Run WebToEpub
3. Set only chapter "Chapter bundles" ( https://kemono.su/patreon/user/22614979/post/93789209 )
    or download last 20 chapters, which includes that chapter
5. Click "Pack epub"

**Expected behavior**
In resulting epub, "Chapter bundles" doesn't show the image, and the image is not in the epub either. This is correct because the entire 'div.ad-container' is removed, as per KemonopartyParser.js.

**Actual behavior**
In resulting epub, "Chapter bundles" doesn't show the image. 
BUT if you look into the epub, you will find the image in the Images folder, and a mention of said image in content.opf.

**Additional context**
As far as I understand, this happens because [fetchWebPageContent](https://github.com/dteviot/WebToEpub/blob/acee3bf4bf7fcccf309f648e4528956b62ecf635/plugin/js/Parser.js#L503) fetches both the content and then the images, before [convertRawDomToContent](https://github.com/dteviot/WebToEpub/blob/acee3bf4bf7fcccf309f648e4528956b62ecf635/plugin/js/Parser.js#L90) calls `removeUnwantedElementsFromContentElement`, which removes the element where the image is shown.

`fetchWebPageContent` actually conveniently already removes some elements via `removeUnusedElementsToReduceMemoryConsumption` before the images are fetched.
I tried to solve this myself by simply calling `removeUnwantedElementsFromContentElement` right after `removeUnusedElementsToReduceMemoryConsumption`, but 
I'm running into `contentElement.ownerDocument is null` error at [utils.removeEventHandlers](https://github.com/dteviot/WebToEpub/blob/acee3bf4bf7fcccf309f648e4528956b62ecf635/plugin/js/Util.js#L241), which is a function that has a @todo and is way beyond me.

**Solution**
There is a possible immediate hotfix for KemonopartyParser, which would override `removeUnusedElementsToReduceMemoryConsumption` and remove the image in there, but this would still leave this problem in all the other parsers that use just `removeUnwantedElementsFromContentElement`. Or maybe I was supposed to use it like this when changing the KemonopartyParser in the first place?
I think the solution is either somehow include the `removeUnwantedElementsFromContentElement` in `removeUnusedElementsToReduceMemoryConsumption`, or (the worse option) keep both as they are and erase all unused images at the very end (which would still download the images, though ...).

I hope I explained it clearly.

perishableloc:
Fixes #1168 for KemonopartyParser

### DIFF
--- a/plugin/js/parsers/KemonopartyParser.js
+++ b/plugin/js/parsers/KemonopartyParser.js
@@ -39,9 +39,8 @@ class KemonopartyParser extends Parser{
         }));
     }
 
-    removeUnwantedElementsFromContentElement(element) {
-        util.removeChildElementsMatchingCss(element, ".ad-container");
-        super.removeUnwantedElementsFromContentElement(element);
+    preprocessRawDom(webPageDom) {
+        util.removeChildElementsMatchingCss(webPageDom, ".ad-container");
     }
 
     findContent(dom) {


dteviot:
Thanks.

PR #1172, opened 2024-01-21
Remove Royalroad watermark

perishableloc:
Royal road recently added watermarks - in every chapter, they insert a single `<p>` element.
This element is then set `display: none` by internal css.
You can check this easily by running following code in any chapter:
`[...document.querySelectorAll('p')].filter(p => getComputedStyle(p).display == 'none');`

As WebToEpub doesn't grab css, the text is shown in final epub.

This PR grabs the internal css, looks for elements that are set `display: none` and removes them.

### DIFF
--- a/plugin/js/parsers/RoyalRoadParser.js
+++ b/plugin/js/parsers/RoyalRoadParser.js
@@ -32,6 +32,27 @@ class RoyalRoadParser extends Parser{
         return content || dom.querySelector(".page-content-wrapper");
     }
 
+    preprocessRawDom(webPageDom) { 
+        this.removeWatermarks(webPageDom);
+    }
+
+    //watermarks are regular <p> elements set to "display: none" by internal css
+    removeWatermarks(webPageDom) {
+        //href === null means it's internal css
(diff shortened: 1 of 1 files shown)


dteviot:
@perishableloc 
Have you tested this works as intended, for both Firefox and Chrome?
My experience with getComputedStyle() is that it doesn't work for WebToEpub.

dteviot:
@perishableloc 
Never mind, I see you didn't use getComputedStyle() in the end.

dteviot:
@perishableloc 
Nope.  Doesn't work.  Try this.
```javascript
        let internalStyles = [...webPageDom.querySelectorAll("style")]
            .map(style => style.sheet?.rules);
        let allCssRules = [];
        for(let ruleList of internalStyles) {
            for(let rule of ruleList) {
                allCssRules.push(rule);
            }
        }
        for(let rule of allCssRules.filter(s => s.style.display == "none")) {
            webPageDom.querySelector(rule.selectorText)?.remove();
        } 
    ```
    
    For my notes: 107 minutes work

dteviot:
Got tired of waiting.  Committed my updated version.

PR #1341, opened 2024-06-08
Update WetriedParser

ImLJS:


dteviot:
@ImLJS 

Looks like needs an update to find the series_id.  Is returning wrong value for https://wetriedtls.com/series/infinite-mage
Works for:
* https://wetriedtls.com/series/sss-class-hunter-who-dies-to-live
* https://wetriedtls.com/series/a-knight-who-eternally-regresses
* https://wetriedtls.com/series/my-platinum-blonde-childhood-friend-is-too-pretty
* https://wetriedtls.com/series/hard-carry-support

Actually, looks like there's only 14 stores on the site, and only infinite mage doesn't work.  (Typical, the one I test is the one that's broken.)

Could be lazy and just hard code the special case
in findSeriesId()

Never mind, I'll do that.

Issue #1342, opened 2024-06-10
webnovel.com (Qidian) all fanfic chapters missing title


seba11998:
**Describe the bug**
It seems like something changed in webnovel's code because now no chapter has a title and its full of `<div>` and `<i>`.

**To Reproduce**
1. Go to: [Book example](https://www.webnovel.com/book/creating-manga-of-one-piece-in-pirates-world-become-my-ultimate-goal!_21965064805459105)
2. Download book

**Screenshots**
Before:

![Screenshot 2024-06-10 013524](https://github.com/dteviot/WebToEpub/assets/14128134/779807c9-1c1f-4606-916b-d1838bf02075)

After:

![Screenshot 2024-06-10 013623](https://github.com/dteviot/WebToEpub/assets/14128134/204619cf-f960-4092-8d17-5f7d97d80713)

Kiradien:
Yeah, this bug has been going on for a few days along with a few intermittent bugs. They changed the page style when viewing and the chapter title is no longer within the div.page h1 it was originally in. 

I've got a fix sitting in a branch somewhere, I'll cherry-pick the part you need and add a pull request. Webnovel's been a bit weird since these changes went in, sometimes throwing random errors like HTTP 445, so I have a bunch of fixes that may or may not come up in the future.

dteviot:
@seba11998 

Test versions for Firefox and Chrome have been uploaded to https://drive.google.com/drive/folders/1B_X2WcsaI_eg9yA-5bHJb8VeTZGKExl8?usp=sharing. Pick the one suitable for you, follow the "How to install from Source (for people who are not developers)" instructions at https://github.com/dteviot/WebToEpub/tree/ExperimentalTabMode#user-content-how-to-install-from-source-for-people-who-are-not-developers and let me know how it goes.
Tested with:
* https://www.webnovel.com/book/king-of-gods_9017100806001205, chapter 1

For my notes: 31 minutes work

seba11998:
Thank you, it works. The xhtml still has a bunch of extra `<i>` and `<div>` but the main problem is solved. Thank you so much

dteviot:
@seba11998 @Kiradien 

Updated version (0.0.0.160) has been submitted to Firefox and Chrome stores.
Firefox version is available now.
Chrome might be available in a few hours to 21 days.

PR #1343, opened 2024-06-10
#1342 - Update QidianParser.js for new page style

Kiradien:
Re-added webnovel chapter title. Removed excess links added at bottom of page.

### DIFF
--- a/plugin/js/parsers/QidianParser.js
+++ b/plugin/js/parsers/QidianParser.js
@@ -40,7 +40,7 @@ class QidianParser extends Parser{
     }
 
     findContent(dom) {
-        return dom.querySelector("div.cha-content");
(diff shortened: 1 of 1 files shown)


dteviot:
@Kiradien 
1. Are you sure that "div.cha-content" should be removed?  When I look at https://www.webnovel.com/book/endless-path-infinite-cosmos_11766562205519505/the-beginning-of-the-end.-part-1-2_31586142793028464, this element is wrapping the wanted content, and also inside div.chapter_content.  Removing this will remove the wanted content.
2. I think you missed updating line 56.  I think you need to change it to
```javascript
        content.className = "chapter_content";
```

dteviot:
@Kiradien 

D'oh!  Thanks.

PR #1352, opened 2024-06-18
Handle qidian HTTP 445. Effectively acts the same as 500 error.

Kiradien:
This HTTP Error code from Qidian/Webnovel went into hibernation for a while, but it's baaack. Offering handling.


dteviot:
Thank you for your hard work

Issue #1365, opened 2024-06-30
Removing epub from Library does not reset include status


Kiradien:
**Describe the bug**
When adding a chapter to an epub, it automatically registers the story in your Reading List. Deleting the epub (often to resolve generation issues) does not reset this status, leaving the chapters forever unchecked.

**To Reproduce**
1. Add a story to the library
2. Remove the story from the library
3. Open the page again, epub is gone but still in reading list.

**Expected behavior**
Removal from library also removes from reading list.

**Desktop (please complete the following information):**
 - OS: Windows
 - Browser Chrome
 - Version [WebToEpub 0.0.0.160] [Chrome 126.0.6478.127]

**Additional context**
A pull request for this fix will be in place soon. This was a complex but low priority problem that has existed for a fairly long time and just annoyed me enough today to dig into it. Complex enough that it absolutely needs a ticket.

dteviot:
Test versions for Firefox and Chrome have been uploaded to https://drive.google.com/drive/folders/1B_X2WcsaI_eg9yA-5bHJb8VeTZGKExl8?usp=sharing.

dteviot:
Updated version (0.0.0.167) has been submitted to Firefox and Chrome stores.
Firefox version is available now.
Chrome might be available in a few hours to 21 days.

PR #1366, opened 2024-06-30
#1365 - Update readinglist when removing epub

Kiradien:
- Added userpreferences to Library
- Added try-delete to readingList with boolean return value should status be required.
- called readingList through userPreferences object in Library


dteviot:
@Kiradien 

Thank you.

Issue #1368, opened 2024-07-01
Tapas.io error 


Wilhelmia:
**Describe the bug**
When I go to Tapas.io and try to get an epub from their novel page I get an error. Until last week it worked very well and I could fetch some epubs, now the error looks like this:

Error: Fetch of URL 'https://tapas.io/series/266599/episodes?page=1&sort=OLDEST&max_limit=9999' failed with network error 500.
    at FetchErrorHandler.onResponseError (chrome-extension://akiljllkbielkidmammnifcnibaigelm/js/HttpClient.js:32:25)
    at HttpClient.checkResponseAndGetData (chrome-extension://akiljllkbielkidmammnifcnibaigelm/js/HttpClient.js:192:45)
    at HttpClient.wrapFetchImpl (chrome-extension://akiljllkbielkidmammnifcnibaigelm/js/HttpClient.js:182:31)
    at async TapasParser.getChapterUrls (chrome-extension://akiljllkbielkidmammnifcnibaigelm/js/parsers/TapasParser.js:14:21)

**Screenshots**
![image](https://github.com/dteviot/WebToEpub/assets/174390898/fa3525ec-5ad0-44b8-abad-755a8638e7b4)

**Desktop (please complete the following information):**
 - OS: [Windows 10]
 - Browser [Vivaldi]
 - Version [6.9.3392.10]

Kiradien:
Hmm, it looks like the api changed to not accept a max_limit greater than 20.
https://tapas.io/series/266599/episodes?page=1&sort=OLDEST&max_limit=20 works, but 21 does not.
The parser needs to be shifted to a more standardized pagenation format with multiple requests.

I may do a deeper peek into it a bit later, but it doesn't look like too bad to fix. Still, it's annoying that they gave a 500 error on that.

Wilhelmia:
> Hmm, might be more of a revamp than originally expected. It looks like most of the selectors across the parser might be invalid now, but I'm not sure I'm running it from the right view.
> 
> @Wilhelmia - can you confirm the url you opened WebToEpub from? Does the url match the following: https://tapas.io/series/villains-are-destined-to-die-novel ? Or was there a specific "table of contents" page?

Hello!
Thank you for looking into it. I open the special table of contents page which is this link: [https://tapas.io/series/villains-are-destined-to-die-novel/info](https://tapas.io/series/villains-are-destined-to-die-novel/info)

[...]

PR #1371, opened 2024-07-06
#1368 - Update TapasParser for multi-request

Kiradien:
Changed parse format to json data instead of json body to best access pagination data


dteviot:
@Kiradien 
Thanks

Issue #1373, opened 2024-07-07
ScribbleHub extractor seems to be broke


fake-name:
**Describe the bug**
Trying to convert a scribblehub page to epub fails with an error, generating a ~4kb epub file.

```
Error: Could not find content element for web page 'https://www.scribblehub.com/series/997267/maou-shoujo-magical-chaos/'.
    at chrome-extension://akiljllkbielkidmammnifcnibaigelm/js/Parser.js:501:23
    at async Promise.all (index 0)
    at async ScribblehubParser.fetchWebPages (chrome-extension://akiljllkbielkidmammnifcnibaigelm/js/Parser.js:470:17)
```

**To Reproduce**
1. Go to a series page 'https://www.scribblehub.com/series/997267/maou-shoujo-magical-chaos/'
2. Try to save as epub
3. If you include the info page (the first page the chapter list extractor finds), that page fails to extract, generating the above exception.

**Expected behavior**
Previously, this stuck the series info page as the first "chapter" in the book, which was super handy.
 
**Additional context**

I _think_ this can be fixed as easily as changing the content selector to `div.fic_row, div#chp_raw`.

fake-name:
Ok, I managed to load the unpacked extension and can confirm that change resolves the issue. See https://github.com/dteviot/WebToEpub/pull/1374

dteviot:
@fake-name 

Test versions for Firefox and Chrome have been uploaded to https://drive.google.com/drive/folders/1B_X2WcsaI_eg9yA-5bHJb8VeTZGKExl8?usp=sharing.

fake-name:
Test version for chrome at least seems to be working.

PR #1374, opened 2024-07-07
Fix information page extraction. 

fake-name:
Fixes https://github.com/dteviot/WebToEpub/issues/1373


dteviot:
@fake-name 
Normally I'd request you add yourself to package.json and the readme before I'd accept the PR.  But as this is a one line change, I'll just accept it as is.  If you want to post an update so you can get credit, I'll accept later.

Anyway, thanks.

fake-name:
That's fine. This is such a small change I don't think it's worth bothering with credit or w/e.

Issue #1405, opened 2024-08-07
1 Sec/Chapter


Iadders:
I keep getting banned from https://69shuba.cx/index.html because the download speed is too fast. And 3 Secs/Chapter takes too long.

gamebeaker:
@Iadders you can set [Max web pages to fetch simultaneously](https://github.com/dteviot/WebToEpub/wiki/Advanced-Options#max-web-pages-to-fetch-simultaneously) to 2 or 4.

Iadders:
> @Iadders you can set [Max web pages to fetch simultaneously](https://github.com/dteviot/WebToEpub/wiki/Advanced-Options#max-web-pages-to-fetch-simultaneously) to 2 or 4.

Huh, that works. 
Still pretty slow, but it will do. Thanks.

gamebeaker:
@Iadders as a warning you could have this problem #1359 i am not sure if these sites are the same.

gamebeaker:
#1094 would also solve it.

Charity29614:
make it 1.5 seconds, one second will probably be too short

Kiradien:
For the time being, I'm marking this as completed - but not yet deployed. The dev version of WebToEpub has 69shuaba's parser set to default 1 second minimum as mentioned in #1378 - if it becomes a problem in the future with this version in use, we can change that minimum period.

Note: the 1-second minimum is designed to work with default settings, and may require rollback on simultaneous settings.

gamebeaker:
@Kiradien why can't you run lint locally?

gamebeaker:
You can download the latest compiled versions here: [https://github.com/dteviot/WebToEpub/actions](https://github.com/dteviot/WebToEpub/actions)
For example the latest commit is this [https://github.com/dteviot/WebToEpub/actions/runs/10300694773](https://github.com/dteviot/WebToEpub/actions/runs/10300694773) and under Artifacts is a zip wich contains both compiled versions [https://github.com/dteviot/WebToEpub/actions/runs/10300694773/artifacts/1789920348](https://github.com/dteviot/WebToEpub/actions/runs/10300694773/artifacts/1789920348)

dteviot:
@Iadders 

Updated version (0.0.0.167) has been submitted to Firefox and Chrome stores.
Firefox version is available now.
Chrome might be available in a few hours to 21 days.

Issue #1378, opened 2024-07-11
Set 1 sec/chapter as default download rate


dteviot:
From @gamebeaker 

> as an idea maybe select a value for [Manual Throttle|Delay per chapter](https://github.com/dteviot/WebToEpub/wiki/Advanced-Options#manual-throttledelay-per-chapter) in the "default" configuration meaning if someone installs the extension for the first time. The reason is that a lot of sites have a cloudflare protection or something similar. I also know a few pages that if your IP gets on a blocklist they don't send a 4xx code but an empty document. And to get unblocked you have to wait.
If you want to implement it I would set 1 sec/chapter as default.

Kiradien:
Just want to tack an addendum on here, it might be worth setting the rate parameter as a parser variable; different sites are 'good' at different speeds, CloudFlare pretty much demands 3-5 seconds depending on the domain.

dteviot:
@Kiradien 
I'm going to assume you mean "save value on a per site basis".  If I'm wrong please correct.

Kiradien:
Wow, I'm getting back to this message late, I blame travelling.
https://github.com/Kiradien/WebToEpub/commit/a28c61285020a0a11a0f997754cf1717625868a0 - an example of what I was referring to. We just set that object variable in each parser where speed is a known issue.

We can also, if needed, set the global default minimum of Parser.js to 1000 for 1 second, if needed, but it does seem like a more tailored option might be better.

It's tested and works for ScribbleHub and Fanfiction.net in the above changeset, but those are just the ones I remember using Cloudflare.

The question is: Do we want to go with this approach?

dteviot:
Test versions for Firefox and Chrome have been uploaded to https://drive.google.com/drive/folders/1B_X2WcsaI_eg9yA-5bHJb8VeTZGKExl8?usp=sharing. Pick the one suitable for you, follow the "How to install from Source (for people who are not developers)" instructions at https://github.com/dteviot/WebToEpub/tree/ExperimentalTabMode#user-content-how-to-install-from-source-for-people-who-are-not-developers and let me know how it goes.

dteviot:
Updated version (0.0.0.167) has been submitted to Firefox and Chrome stores.
Firefox version is available now.
Chrome might be available in a few hours to 21 days.

Issue #1377, opened 2024-07-10
Web2Epub Error : Failed to fetch


NightHawk007:
Hi i tried to download epub of a novel Absolute Resonance & shadow slave on different sites, but all stopped after some chapters maybe 1-129,1-200 or 1-3,1-817 etc, tried on the same and different sites multiple times and the number always varies and stops and not able to get the full download of all chapters tried on these sites

https://ranobes.top/novels/1072291-absolute-resonance-v741610.html
https://ranobes.top/novels/1205249-shadow-slave-v741610.html
https://novelfull.com/shadow-slave.html
https://novelfull.net/shadow-slave.html
https://novelbin.me/novel-book/shadow-slave#tab-chapters-title
https://freewebnovel.com/shadow-slave.html

Also tried the experimental version, install from source too (https://drive.google.com/drive/folders/1B_X2WcsaI_eg9yA-5bHJb8VeTZGKExl8?usp=sharing) in firefox it also stopped after some chapters.

Ranobes
TypeError: Failed to fetch

freewebnovel
TypeError: Failed to fetch
Warning, unable to convert chapter 'Chapter 431 Fall From Grace' from 'https://freewebnovel.com/shadow-slave/chapter-431.html' to valid XHTML. Your epub viewer may fail when viewing that chapter. You may need to fix the chapter manually with Calibre. 
Conversion error message was: (This page contains the following errors:error on line 1 at column 3635: Specification mandates value for attribute freewebn
Below is a rendering of the page up to the first error.)
Warning, unable to convert chapter 'Chapter 432 First Day of Nothing' from 'https://freewebnovel.com/shadow-slave/chapter-432.html' to valid XHTML. Your epub viewer may fail when viewing that chapter. You may need to fix the chapter manually with Calibre. 
Conversion error message was: (This page contains the following errors:error on line 1 at column 1893: error parsing attribute name
Below is a rendering of the page up to the first error.)

gamebeaker:
@NightHawk007 do you have [Manual Throttle|Delay per chapter](https://github.com/dteviot/WebToEpub/wiki/Advanced-Options#manual-throttledelay-per-chapter) active? I am at Chapter 600 on both with 4 Max web pages to fetch simultaneously and 3 Manual Throttle|Delay per chapter.

gamebeaker:
I was able to download both without a problem (Besides one image res. warning for absolute resonance and the two valid xhtml for shadow slave).

NightHawk007:
[...]

PR #1407, opened 2024-08-07
#1378   create default parser rate

Kiradien:
Implemented logic for a default minimum rate threshold for requests.

Base Parser.js minimum threshold set at null/0ms. 

Implemented default values for Fanfiction, Scribblehub and 69shuParser

As this is a **non-standard** fix across a multitude of files, I will not close this pull request without external approval.

Affected Tickets:
#1377 #1378 #1405 

Ticket numbers are lost in squashes.

### DIFF
--- a/plugin/_locales/en/messages.json
+++ b/plugin/_locales/en/messages.json
@@ -311,6 +311,10 @@
 		"message": "Additional Metadata URL",
 		"description": "Additional Metadata URL"
 	},
+	"__MSG_label_Override_Default_Minimum_Delay__": {
+		"message": "Override Minimum Throttle Value (where applicable)",
+		"description": "Override Minimum Throttle Value (where applicable)"
+	},
 	"__MSG_label_Overwrite_Epub_When_Filename_Duplicte__": {
 		"message": "Overwrite existing EPUB file with same name",
 		"description": "Label for checkbox to toggle between 'overwrite exisitng Epub files' and 'make names unique'"
--- a/plugin/js/Parser.js
+++ b/plugin/js/Parser.js
@@ -51,8 +51,9 @@ class ParserState {
     }
 }
 
-class Parser {
+class Parser {    
     constructor(imageCollector) {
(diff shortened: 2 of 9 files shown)


gamebeaker:
Works for me i would add this.minimumThrottle = 0; to Template.js so that if someone new tries to write a parser he knows this option exists. Mabe with a short comment like "If the sites throttles requests you can slow them down with minimumThrottle (wait time is in ms)..."

Kiradien:
> Works for me i would add this.minimumThrottle = 0; to Template.js so that if someone new tries to write a parser he knows this option exists. Mabe with a short comment like "If the sites throttles requests you can slow them down with minimumThrottle (wait time is in ms)..."

Hmm, that's a good point. That said, however, we probably wouldn't want every parser to set minimumThrottle unless they need it, else the 'default' setting in the base parser loses its worth. Given the structure of the template, perhaps marking it as an optional parameter is the best method? e.g. Sharing it with the default cloudflare value within a comment like the following:

```
    constructor() {
        super();
        //Optional Parameters:

        /*
        // Minimum delay (in ms) between page requests. 
        // If the sites this parser accesses throttles requests or uses cloudflare, it is recommended to set this.
        this.minimumThrottle = 3000;
        */
    }
```

gamebeaker:
You are right. When i think back to my first parser i remember how i simply copied the template without understanding what is hapening in the background. Slow downs weren't a problem at this time but as a lot of sites use cloudflare now it is the best not to set it to 0 so that it is working "out of the box" even if it isn't understood what it does. Could you add an additional line to the comment?
`//403 error prevention`
I think that helps non technical people if they try to troubleshoot.

Kiradien:
Yeah, that's probably the most important chunk to include for that. Pushed it up to git.

dteviot:
@Kiradien 
Thank you for your efforts.

This is much simpler than I was thinking.  Which was allowing the user to set the rate on a per-site basis.

PR #1419, opened 2024-08-11
walkTocPages() add slow down

gamebeaker:
Example: https://www.scribblehub.com/series/16599/bl-oh-my-fate/
Has 90 requests for Searching for URLs at the moment they are fetcht without slow down.


gamebeaker:
I am not sure if this should be merged, because i haven't run into a problem where i got 403 during "Searching for URLs".

dteviot:
@gamebeaker Yes, it should be there.

Issue #1410, opened 2024-08-08
Create new github release with each pull


gamebeaker:
In github actions is already a workflow that automatically builds the addon. The idea is that this build is automatically released as a new version on github so that people can get the latest version on github. It looks like there is a pre-release flag that could be associated with such releases.
![grafik](https://github.com/user-attachments/assets/4d1778c8-d6dc-4f8e-b1f4-3dfdb295ac2e)

@dteviot doesn't have to upload the latest release to google drive anymore.

Related to #41 how should the versoning be handeled?
0.0.0.160-beta.01 and 01 increases with each pull?

Another option would be to override the latest pre release binarys with the new ones.
0.0.0.160-beta 
Problem: If someone installs this version it could be unclear to him that a few weeks later the version with the same name is another version.

I don't know how github actions work so help is appreciated.

Kiradien:
Hmm, this is something that both Azure and Jenkins provide in their workflows, but I'm not entirely sure on the implementation for GitHub. 
A lot would be based on implementing it into the existing workflow: https://github.com/dteviot/WebToEpub/blob/ExperimentalTabMode/.github/workflows/node.js.yml

An existing Workflow which seems to do this is here: https://github.com/github/safe-settings/blob/main-enterprise/.github/workflows/create-pre-release.yml

Some documentation & general information
https://blog.kubesimplify.com/automated-github-releases-with-github-actions-and-conventional-commits
https://github.com/marketplace/actions/github-releases-for-automated-package-publishing

I haven't read it all yet myself, but it should make for some interesting reading...

Kiradien:
Random note from another dev issue, don't want to spam that thread:
> @Kiradien why can't you run lint locally?
![image](https://github.com/user-attachments/assets/ecd07062-9423-49e6-b54e-d405845209c0)
I've fixed the issue locally a few times, but it always ends up being a pain when running my Windows install.

gamebeaker:
@Kiradien do you run windows or linux?

Kiradien:
Both - Linux for work, Windows for relaxation. However, I tend to be booted into Windows when I work on and test WebToEpub.

gamebeaker:
If you use vscode in the right bottom corner is a button where you can change lf to crlf.
![grafik](https://github.com/user-attachments/assets/8b726434-8d16-4662-8075-5297a24373d6)

Kiradien:
[...]

PR #1429, opened 2024-08-19
Added auto release dev-build

gamebeaker:
Related to #1410 i don't know how to test github actions so i push it to test.


gamebeaker:
i can test it in my own repo so that's what i am going to do (i think it is a permission problem)

PR #1445, opened 2024-08-25
fix_www_subdomain_cookie_issue

### LINKED ISSUE #1444: Chrome 403 errors on Cloudflare - Cont.
This is a continuation of #1439 - said bug is fixed for mtlarchive, but other sites are still problematic. As the main site mentioned has been fixed, I am raising a separate issue for other domains still having said issues.

Domains in question:
Fanfiction.net
Scribblehub.com - Actually still generates, but still missing cf_clearance, it will likely  fail during high load periods.

Chromium Ticket: https://issues.chromium.org/issues/361862746

Note, the current fix also results in a variety of errors showing up in the dev console, however these are relatively easy to fix once we have a permanent solution:
image

@gamebeaker - poking you on this since you found the initial solution and while MTLArchive is fixed, issues persist. 

Feel free to add other domains where this issue persists in the comments, I will try to keep the above list up to date.

gamebeaker:
I hope this fixes #1444
Problem:
On fanfiction.net
parsedUrl.hostname is www.fanfiction.net but the "domain" is .fanfiction.net
if www. is striped from the hostname it matches the cookies.

### DIFF
--- a/plugin/js/HttpClient.js
+++ b/plugin/js/HttpClient.js
@@ -217,7 +217,7 @@ class HttpClient {
                 //  get all cookie from the site which use the partitionKey (e.g. cloudflare)
                 //keep old code for reference in case it changes again
                 //let cookies = await chrome.cookies.getAll({partitionKey: {topLevelSite: topLevelSite}});
-                let cookies = await chrome.cookies.getAll({domain: parsedUrl.hostname,partitionKey: {}});
+                let cookies = await chrome.cookies.getAll({domain: parsedUrl.hostname.replace(/^www./, ""),partitionKey: {}});
 
                 //create new cookies for the site without the partitionKey
                 //cookies without the partitionKey get sent with fetch


dteviot:
@gamebeaker 
I'm wondering if should try fetching cookies twice.
1. With "www." prefix if there is one.
2. Without the "www." prefix

gamebeaker:
@dteviot i don't think this is necessary. I haven't seen a cookie where there is a leading www. in the domain section. 
Do you have an example for such a cookie?

gamebeaker:
@dteviot It is not necessary, because all subdomains are included
example:
{domain: example.com,partitionKey: {}}
matches cookies with domain set to:
www.example.com or subdomain.example.com or example.com
image
https://developer.chrome.com/docs/extensions/reference/api/cookies#method-getAll
I tested it and it looks like the docs are correct.

dteviot:
@gamebeaker 

I stand corrected.
Thank you for your hard work.

PR #1450, opened 2024-08-28
Fix Patreon Collections

Tyderion:
Patreon collections changed again (I guess...).
This collects all loaded post links from a collection using a nested h3 as discriminator (this is the title of the card), there's no data-tag attributes anymore on the collection page


dteviot:
@Tyderion 
Thank you for your hard work.

Issue #1423, opened 2024-08-14
Please add site https://genesistudio.com 


nandakishore2009:
Hi.
https://genesistls.com is changed to https://genesistudio.com
The site is fully revamped and the extension not working.
Is there a chance of supporting it again?
Thanks

dteviot:
@nandakishore2009 
Sorry, site is doing things to disable the site when it detects dev tools being used.  This makes it difficult to write a parser for the site.  I could probably work around by using Fiddler, but to me it's not worth the effort.

Time taken: 14 minutes

gamebeaker:
It looks like they use this here to prevent it: [https://github.com/theajack/disable-devtool](https://github.com/theajack/disable-devtool)
To prevent this js from loading you can block it with uBlock Origin. I blocked this url and it looks like it works: `https://genesistudio.com/_app/immutable/chunks/disable-devtool*`

dteviot:
@gamebeaker 

Thank you for your efforts.  Unfortunately, the Google rules are that Chrome extensions are not allowed to circumvent copyright protection measures. And I suspect the site is doing this for at least partly this reason.

So, not going to support this site in version released to stores.  But might do it for the experimental builds.  Will take a look at the site for curiosity's sake.

dteviot:
@gamebeaker 

For my future reference, how did you figure this out?

dteviot:
Looking at the network traffic now.
To get content for a chapter, is POSTing a REST call, to https://analytics.genesistudio.com/api/send with URL in the JSON "/viewer/723", which is same as the hyperlink for chapters.  But there's other stuff in the payload.  But it looks like the response is encrypted.  So, this is another site where you'd probably need to load each page normally, let the browser unscramble the text and then extract it.

So, not going be able to handle this site.  It's job for a different tool.

gamebeaker:
@dteviot  I first tried to open dev tools -> that didn't work (I used the shortcut CTRL+Shift+E to open them).  I googled how to detect dev tools open and found this [https://stackoverflow.com/questions/7798748/find-out-whether-chrome-console-is-open/24327093#24327093](https://stackoverflow.com/questions/7798748/find-out-whether-chrome-console-is-open/24327093#24327093) answer. 
I realized that
To test if this method is used i prevented the console from clearing in firefox 
![image](https://github.com/user-attachments/assets/150dd05e-cf23-44df-82dd-c9c64590df95)
After opening the page the console looked like this.
[...]

PR #1466, opened 2024-09-07
Add genesi studio parser

gamebeaker:
@dteviot reference: #1423

I have a problem:
I set the Chapterlist links to the api links e.g.: [https://genesistudio.com/viewer/6213/__data.json](https://genesistudio.com/viewer/6213/__data.json)
and in findContent() i want to extract the content from the json response but at the moment it fails i think the problem is that it is not valid html is there a solution for this problem?
A "hack" would be i set the links in Chapterlist to a valid HTML page e.g. [https://genesistudio.com/viewer/6213](https://genesistudio.com/viewer/6213) and than in the function findContent() 
`let data = (await HttpClient.fetchJson(dom.baseURI + "/__data.json")).json;`
but i don't think this is the right approach.

### DIFF
--- a/plugin/js/parsers/GenesiStudioParser.js
+++ b/plugin/js/parsers/GenesiStudioParser.js
@@ -8,33 +8,73 @@ parserFactory.register("genesistudio.com", () => new GenesiStudioParser());
 class GenesiStudioParser extends Parser{
     constructor() {
         super();
-        //Optional Parameters:
-
-        /*
-        // Minimum delay (in ms) between page requests. Useful for 403 error prevention.
-        // If the sites this parser accesses throttles requests or uses cloudflare, it is recommended to set this.
         this.minimumThrottle = 3000;
-        */
     }
 
-    // returns promise with the URLs of the chapters to fetch
-    // promise is used because may need to fetch the list of URLs from internet
-    /*
-        https://genesistudio.com/novels/yar
-
-    */
-
-    async getChapterUrls(dom, chapterUrlsUI) {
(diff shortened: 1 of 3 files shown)


gamebeaker:
@dteviot i think i got it i overload `async fetchChapter(url)` i got inspiration from here: [https://github.com/dteviot/WebToEpub/blob/ExperimentalTabMode/plugin/js/parsers/InoveltranslationParser.js#L53](https://github.com/dteviot/WebToEpub/blob/ExperimentalTabMode/plugin/js/parsers/InoveltranslationParser.js#L53)

gamebeaker:
It works :)
Tested with: 
https://genesistudio.com/novels/yar
https://genesistudio.com/novels/nw

dteviot:
@gamebeaker 

Wow.  Thank you for your hard work.

> @dteviot i think i got it i overload async fetchChapter(url) i got inspiration from here: https://github.com/dteviot/WebToEpub/blob/ExperimentalTabMode/plugin/js/parsers/InoveltranslationParser.js#L53

That's the way I would do it.

Please note, you can't assign to innerHTML in a Firefox extension.  The automated checker won't accept it, because it can't validate the HTML is safe.  I've pushed a correction.  And taken the liberty to refactor titleFormJson() and findCoverImageUrl()
https://github.com/dteviot/WebToEpub/commit/cf458e01ebafea699237d6e3732077978909a656

Tested with: 
* https://genesistudio.com/novels/yar, prolog and chapter 1

Issue #1454, opened 2024-09-03
Enhance QuestionableQuesting: Export all posts (including non-threadmark posts) to epub


xypha:
**Problem**
Non-threadmark posts cannot be exported.

Steps to replicate:
* Open 'Story Only' thread of [With This Ring](https://forum.questionablequesting.com/threads/with-this-ring-young-justice-si-story-only.8961/#post-2408762) in browser tab.
* **Scenario 1**: Run `WebToEpub` from toolbar icon → only 1 chapter is loaded. None of the threadmarks on the page are seen.
* In the browser tab containing QQ post, click on "Threadmarks" button and select "View all 148 threadmarks" option.
* **Scenario 2**: Run `WebToEpub` from toolbar icon → only 25 chapters are loaded.
* In the browser tab containing QQ post, on the "Threadmarks" overlay, change the "Per page:" option to maximum (as of 2024.09.03, it is 400).
* **Scenario 3**: Run `WebToEpub` from toolbar icon → all 148 chapters are loaded, but non-threadmark posts cannot be exported.

![WebtoEpub issue 1](https://github.com/user-attachments/assets/cc23f760-a2b9-48b3-9d7a-96dc83f6299d)
![WebtoEpub issue 2](https://github.com/user-attachments/assets/106857f1-242a-4804-9c34-5f4bdbd41945)
![WebtoEpub issue 3](https://github.com/user-attachments/assets/c1e95a39-b4de-403e-b91d-0e594719d56e)

**Describe the solution you'd like**

Possible solution to **Scenario 1 and 2**:
* In `WebToEpub` popup tab,  add warning text (maybe above the Chapters Count), telling users to ensure all threadmarks are loaded before export.
* Add section in Wiki on how to add all threadmarks to chapter link this on QQ and similar sites (and add a link in the warning text if above solution is implemented).

Possible solution to **Scenario 3**:
* Add advanced option to export all posts from all thread pages (in this case, using page navigation instead of threadmarks - pages 1 to 135, as of 2024.09.03).

**Describe alternatives you've considered**
Exporting through [FicHub](https://fichub.net/) (also on [GitHub](https://github.com/FicHub/fichub.net)) solves **Scenario 1 and 2** - but fails to export images (which is a deal breaker).
For **Scenario 3**, the problem persists. Non-threadmark posts cannot be exported.

**Additional context**
Current version: 0.0.0.167
Browser: Firefox 129.0.2 (64-bit)
OS: Windows 11 23H2

[...]

PR #1481, opened 2024-09-10
Reduce http header cookie clutter

Kiradien:
Previous logic attempted to duplicate cookies which were being set by default within chrome's request object. Cookies at issue had specialized settings of "Session" or no_restriction on SameSite.

@gamebeaker - I'd like your opinion on the HttpClient change before pushing it through; under the current build, if you open the chrome dev console while running on a site like fanfiction.net, you can see the number of errors popping up. The errors don't actually break anything, but they do make debugging other things a tad annoying. 
I'm fairly confident it covers all situations for your former fix, but I'd like to know if there's any glaring issues or if you think the filter should be handled from a different angle.

To no one in particular: SpacebattlesParser change is more of an irregularity than anything, I'm fine pushing it but the reasons for the throttle are a bit unusual: By default, cloudflare isn't actually an issue on the Xenforo forums, but when requesting 150+ chapters too quickly (Inconsistent number due to caching) they DO request throttling and eventually temp ban. I have not been able to replicate the error since setting to 50ms, however it is trial&error. It's mainly included here because I forgot to check it in when I changed it.

Now I'll get back to looking at 1454 without these errors. *whistles innocently*

### DIFF
--- a/plugin/js/HttpClient.js
+++ b/plugin/js/HttpClient.js
@@ -222,6 +222,7 @@ class HttpClient {
             let cookies = "";
             if (!util.isFirefox()) {
                 cookies = await chrome.cookies.getAll({domain: urlparts[urlparts.length-2]+"."+urlparts[urlparts.length-1],partitionKey: {}});
+                cookies = cookies.filter(item => item.session || item.sameSite == "no_restriction");
             }else{
                 cookies = await browser.cookies.getAll({domain: urlparts[urlparts.length-2]+"."+urlparts[urlparts.length-1],partitionKey: {}});
             }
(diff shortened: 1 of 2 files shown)


gamebeaker:
@Kiradien i was able to reproduce the error on reader-hub.com, on fanfiction.net i got no error.

I don't know why your fix works but it does.
My two cents:
As the function name is setPartitionCookies() and the original idea was, that we have to set them manually because only they would not be sent with a fetch request the filter could be something like this:
`cookies = cookies.filter(item => item.partitionKey != undefined);`
with this only the cookies with a partitionKey get set and not the other ones which we wouldn't have to set anyway.

What is 1454?

dteviot:
@gamebeaker 
>What is 1454?

Probably https://github.com/dteviot/WebToEpub/issues/1454

gamebeaker (approved):

dteviot:
@Kiradien @gamebeaker 
You don't need my approval to merge changes.  If you're both happy, that's more than enough for me.

PR #1490, opened 2024-09-15
Library export import function

gamebeaker:
@Kiradien related to #1365 #1365
I changed how the Readinglist gets updated if the book is deleted. 
Why?
If you open WebToEpub from a not valid page example in 
chrome: `chrome://extensions/`
firefox: `about:addons`
It wasn't possible to delete a epub.
Additionally i added the deletion of the reading list of the saved books if the Library gets cleared with "Clear Library".
Is that ok or did you have other reason to use .addObserver(library)?

Now the original commit:
Allow the user to export/ import the current Library with reading status as json file.
minor fix: the save as dialog got triggered if "Add to Library" is used on a new url -> now only triggers if the Option "Download Epubs after new Chapter got added automatically" is checked.

### DIFF
--- a/plugin/_locales/en/messages.json
+++ b/plugin/_locales/en/messages.json
@@ -23,6 +23,14 @@
 		"message": "Clear Library",
 		"description": "Label on button to clear Local Storage"
 	},
+	"__MSG_button_Lib_Template_Export_Library__": {
+		"message": "Export Library",
+		"description": "Label on button to export Library in json format"
+	},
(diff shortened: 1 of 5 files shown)


Kiradien:
Definitely a more complete / better way of handling it; I went with AddObserver because it was the first method that came to mind to access the object, but considering the frequency "Deletion" will actually be called, reading from storage on call is a far better solution than the constant overhead.

Issue #1491, opened 2024-09-16
Add option to skip 403 retries, for content that is legitimately not available


fake-name:
Basically, I have a number of stories where LARGE swathes content has gone inaccessible. For example, there are a number of series where the author has (rather foolishly) used direct links to their patreon-hosted images in a publicly available story.

Downloading these is extremely annoying, because you get a "you might need to log in to a site" prompt for EVERY ONE of the broken links. You can wind up literally having to click "Yes, I know" more then a hundred times for a single series.

This is extremely annoying. 

It seems to me, given there is already a facility for deciding whether to ask the user (`wrapOptions.retry.promptUser`), that an option that just globally forces that to `false` would be very nice from a user standpoint.

fake-name:
For the moment, I've kind of "redneck patched" the issue by just removing all `alert()` calls in `getAutomaticRetryBehaviourForStatus()`, and changing all `promptUser` values to `false`, but that's not a general fix. Ideally this would be something exposed through the UI, but I'm not smart enough to make that set of changes myself.

gamebeaker:
@fake-name do you have an example story with problematic content?

fake-name:
Try https://www.scribblehub.com/series/945582/the-heavenly-martial-empress-returns-an-op-xianxia-returnee-litrpg/

gamebeaker:
https://www.scribblehub.com/series/945582/the-heavenly-martial-empress-returns-an-op-xianxia-returnee-litrpg/
Chapter 36 is problematic.

gamebeaker:
The current flowchart for HTTP 403:
![Neues draw io Diagramm-93867 drawio](https://github.com/user-attachments/assets/cd5cf352-5713-4900-8d52-7be9e6930d68)

gamebeaker:
@fake-name I changed how 403 gets handled is the new solution ok?
Test versions for Firefox and Chrome have been uploaded to https://github.com/dteviot/WebToEpub/releases/tag/developer-build. Pick the one suitable for you, follow the "How to install from Source (for people who are not developers)" instructions at https://github.com/dteviot/WebToEpub/tree/ExperimentalTabMode#user-content-how-to-install-from-source-for-people-who-are-not-developers and let me know how it goes.
Tested with:
* https://www.scribblehub.com/series/945582/the-heavenly-martial-empress-returns-an-op-xianxia-returnee-litrpg/

dteviot:
@fake-name 
[...]

Issue #1488, opened 2024-09-14
Blacklist for links to and from images?


T-Man2000:
Since images (and other) from (presumably) Patreon are repeatedly displayed on some pages, WTE gives the following error message:

_WARNING: Site 'c10.patreonusercontent.com' has sent an Access Denied (403) error.  You may need to logon to site, or browse site normally until you get a Cloudflare "Are you a human" page or satisfy some other CAPTCHA before WebToEpub can continue. Open Page?_

Can I fill a blacklist somewhere in WTE that generally prevents downloads from these pages from being attempted? Or can a blacklist be built in?

gamebeaker:
@T-Man2000 I only know of the option [Skip Images](https://github.com/dteviot/WebToEpub/wiki/Advanced-Options#skip-images)

T-Man2000:
I would like to integrate the images in general. Just not images from specific pages. This often makes no sense, as you have to be logged in to access the images there. Patreon, Discord, etc. are examples.

gamebeaker:
@T-Man2000 I changed how 403 gets handled is the new solution ok?
Test versions for Firefox and Chrome have been uploaded to https://github.com/dteviot/WebToEpub/releases/tag/developer-build. Pick the one suitable for you, follow the "How to install from Source (for people who are not developers)" instructions at https://github.com/dteviot/WebToEpub/tree/ExperimentalTabMode#user-content-how-to-install-from-source-for-people-who-are-not-developers and let me know how it goes.
Tested with:
* https://www.scribblehub.com/series/945582/the-heavenly-martial-empress-returns-an-op-xianxia-returnee-litrpg/

dteviot:
@T-Man2000 
Updated version (1.0.0.0) has been submitted to Firefox and Chrome stores.
Firefox version is available now.
Chrome might be available in a few hours to 21 days.

Issue #1361, opened 2024-06-29
Automatically handle 403 errors when try to get image from different host


bookimp:
**Describe the bug**

**To Reproduce**
1. Go to https://libread.com/libread/the-rebirth-of-an-ill-fated-consort-65376
2. Download epub with defaul settings other than 3 second delay
3. Get 403 error about logging in (site has no accounts) or solving CAPTCHA
4. You also get a billion failed to dl image errors but I'm less concerned about that.

**Expected behavior**
Epub to download without this popping up a dozen times

**Screenshots**
Attached

**Desktop (please complete the following information):**
 - OS: Windows 11 Home 64bit
 - Browser: Opera LVL 5 (core: 109.0.5097.130)

**WebToEpub Version**
0.0.0.159

**Additional context**
It did allow me to just hit ok on the alert, but I had to be on the download or it stalled. 

![Screenshot 2024-06-29 000104](https://github.com/dteviot/WebToEpub/assets/20101896/8df34f6c-ab91-434b-9cb0-7025baf01675)
[The Rebirth of an Ill-Fated Consort Chapter 1-207.zip](https://github.com/user-attachments/files/16039140/The.Rebirth.of.an.Ill-Fated.Consort.Chapter.1-207.zip)
Just in case you want to see the file without having to download the whole thing.

gamebeaker:
@bookimp try to update to version 0.0.0.160 it looks like this #1306 issue.

gamebeaker:
My estimate was wrong. The problem is that the Autor embeded pictures witch get blocked with 403 error code. Example: [https://freewebnovel.comenovel.com/the-rebirth-of-an-ill-fated-consort/chapter-106](https://freewebnovel.comenovel.com/the-rebirth-of-an-ill-fated-consort/chapter-106) image on this site with error: [https://lh6.googleusercontent.com/xuxiYqUG6e6F7FurcfYQJWCAqmNYLQpD1HINXhnGhjGC6AXkZekhcTSn0Kzx-TFTuTDU3XtWGEPGh5dIyLFlaPk3d1aw0t79jlYc1iFZ](https://lh6.googleusercontent.com/xuxiYqUG6e6F7FurcfYQJWCAqmNYLQpD1HINXhnGhjGC6AXkZekhcTSn0Kzx-TFTuTDU3XtWGEPGh5dIyLFlaPk3d1aw0t79jlYc1iFZ)
I don't know what a solution would be like.
[...]

PR #1496, opened 2024-09-20
Change 403 user options/ handling

gamebeaker:
reference: #1491, #1488, #1361
Added option to block URL.hostname from an 403 error URL. 
This can reduce the ammount of errors an user has to skip if the images on the website are now 403 like on patreon.

Changed 403 handling for the user.
Old:
Neues draw io Diagramm-93867 drawio
New:
Only one error page.
image

### DIFF
--- a/plugin/_locales/en/messages.json
+++ b/plugin/_locales/en/messages.json
@@ -115,6 +115,14 @@
 		"message": "Retry",
 		"description": "Label on 'Retry' button when show error"
 	},
+	"__MSG_button_error_Open_URL__": {
+		"message": "Open URL for Captcha",
+		"description": "Label on 'Open URL for Captcha' button when show error HTTP 403"
+	},
+	"__MSG_button_error_Block_URL__": {
+		"message": "Block Website for future requests",
+		"description": "Label on 'Block Website for future requests' button when show error HTTP 403"
+	},
 	"__MSG_button_finished_default_parser__": {
 		"message": "Apply",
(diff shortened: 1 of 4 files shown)


dteviot:
@gamebeaker 

Did you intend to leave these lines commented out?
https://github.com/dteviot/WebToEpub/blob/957d6989b18b50b02efcffbb330f26e6e883e512/plugin/js/HttpClient.js#L85-L92

gamebeaker:
@dteviot no i was debugging and forgot.

gamebeaker:
@dteviot i pushed a clean up commit here:https://github.com/dteviot/WebToEpub/commit/97a5a407b3b46f8407fcb6f6939357b0ae76571a
What should be the workflow if i want to push code?
1. Should i directly push it in https://github.com/dteviot/WebToEpub/tree/ExperimentalTabMode directly without a pull request
Con:
I think it is harder to keep track of changes if they get directly merged
Pro:
Only 2 Commits in commit history, example:
image
2. Create a branch in my fork and create a pull request for this fork in ExperimentalTabMode.
Con:
Creates 3 Commits in commit history even for small changes (parser), example:
image
Pro:
I feel like it is easier to keep track of changes especially if they are new functionality/ changes in the code (not parser)

dteviot:
@gamebeaker 

Do whichever you are more comfortable with.  I'm good either way.

Also, thank you very much for your work.

PR #1505, opened 2024-09-24
Add xyParser

gamebeaker:
reference: #issue_id
tested with:
url1
url2


gamebeaker:
I am creating a tutorial for contributing and this is a example pull request.

Issue #1455, opened 2024-09-03
Problem converting tables from Royal Road


bardokhofmeier896:
**Describe the bug**
I read a lot of LitRpg and they startet adding <table> for ingame status messages. These messages, if to long, don't convert properly and can't be read on my reader.

**To Reproduce**
1. Go to 'https://www.royalroad.com/fiction/84708/the-gate-traveler-slice-of-life-litrpg'
2. Convert to epub
3. Copy to Reader (I have a Tolino Shine 3)
4. Go to [Chapter 8: My Second Gate]
5. There is a lenghty explanation of the other world that is cut.

**Expected behavior**
I want to read the whole text of the box.

**Screenshots**
Page with the table:
![2024-09-03-08-40-30-897](https://github.com/user-attachments/assets/0d0de907-4f0a-4f00-9991-3138e007d1cf)

Next page: 
![2024-09-03-08-40-37-493](https://github.com/user-attachments/assets/212b1c24-f792-46e1-9da6-d3467c74d5d3)

Actual content of the box:
![image](https://github.com/user-attachments/assets/9aca7a24-cc53-4a65-9161-b43eff2b0860)

If i open the book in calibre it's also cut at the same point.

**Desktop (please complete the following information):**
 - OS: Windows 10
 - Browser Vivaldi (Chromium), also with Firefox 129.0.2 (64-Bit)
 - Version 6.9.3447.37 (Stable channel) (64-Bit)

dteviot:
@bardokhofmeier896 

I'm not sure I can help you.  I just tried to download an view that chapter.
The epub file created by WebToEpub does have all that text.  You could unzip the epub and look at the chapter file with a text viewer, or a browser and you should see all the text is there.

I have seen similar problems (as your Tolino Shine 3) with the table being cut in Edge, when it use to support epubs.  So, I think it's a flaw in your viewer's handling of a table, it only shows as much of the table as will fit on a page.

I tried viewing it in Calibre, and all the text appears.
I don't know why it doesn't appear for you, but I would guess that you've asked Calibre to convert to some format that results in the tables being stripped.  You could try this epub reader.  https://www.epubread.com/en/
![image](https://github.com/user-attachments/assets/4402be0a-6851-4ef4-8d93-7b7dcd0bd94b)

gamebeaker:
[...]

PR #1516, opened 2024-09-26
Create convertTableToDiv to fix royalroad tables on tolino

gamebeaker:
reference: #1455
To use the change add this to the Stylesheet:
```
.WebToEpub-table {
    border-collapse: collapse;
    width: 100%;
}
.WebToEpub-row {
    display: flex;
    border: 1px solid black;
}
.WebToEpub-cell {
    flex: 1;
    padding: 5px;
    border: 1px solid black;
}
```
@dteviot as this is really obscure i would appreciate other ideas to enable this option.
Maybe a new Checkbox "tolino compatibility"?.
Why not always convert tables to div?
It looks like only tolino has the problem and not other reader. As the table looks better than the div version other user shouldn't be impacted from the tolino fix.

Tested with:
https://www.royalroad.com/fiction/22546/inexorable-chaos-complete/chapter/324087/chapter-four-abnormal-summoning
There is a table in a table and the table is to long for tolino.

### DIFF
--- a/plugin/js/Parser.js
+++ b/plugin/js/Parser.js
@@ -89,6 +89,7 @@ class Parser {
         util.decodeCloudflareProtectedEmails(content);
         this.removeNextAndPreviousChapterHyperlinks(webPage, content);
         this.removeUnwantedElementsFromContentElement(content);
+        this.convertTabletoDiv(content);
         this.addTitleToContent(webPage, content);
         util.fixBlockTagsNestedInInlineTags(content);
         this.imageCollector.replaceImageTags(content);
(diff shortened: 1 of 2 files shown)


gamebeaker:
@dteviot 
> @gamebeaker I'm going to say, this isn't something for WebToEpub to do. Instead, should provide a script in EpubEditor.
_Originally posted by @dteviot in https://github.com/dteviot/WebToEpub/issues/1455#issuecomment-2377793706_

Can i leave it in WebToEpub as a "hidden" feature?
`.WebToEpub-table` needs to be in the Stylesheet or it is skipped.
Or should i revert the commit?

dteviot:
@gamebeaker 
Have to go to work.  Will get back to you.

Issue #1452, opened 2024-08-31
URL parser improvements


Darthagnon:
I currently use an AI-generated janky Python script to convert a list of URLs into an HTML-formatted list for use with WebToEpub: https://github.com/Darthagnon/web2epub-tidy-script

It works to solve the workflow problems I have with this extension, as explained in a previous issue (quoted below). 

Would it be possible to adapt the URL parser to automatically do what I currently use my external script to do?

Darthagnon:
Apologies, my explanation was rather confusing. 

"Edit chapter URLs" is the view I use to collect chapter lists to convert to EPUB, because
- the Wizards website is broken/useless/missing chapters, so there is no auto-parser that could work (*EDIT: without too much work*). An auto-parser would need to process `https://magic.wizards.com/en/news/archive` (2024), `https://web.archive.org/web/2023mmddetc/https://magic.wizards.com/en/articles/archive` (unreliable infinite scroller) and `https://web.archive.org/web/2020mmddetc/http://www.wizards.com/Magic/Magazine/Archive.aspx`  (paginated, mostly 404s), `https://web.archive.org/web/2014mmddetc/http://www.wizards.com/Magic/Magazine/Article.aspx`
- a lot of chapters are not story-related, so less useful for EPUB.

## Questions
1. **Parser:** Is it possible for a single parser to target multiple domains (e.g. archive.org, multiple versions of wizards.com) with separate filters on a per-chapter basis? Or are there any example site templates that are similar to this use case?
2. **Extension:** Would it be possible to (optionally) open "Edit chapter URLs" by default, rather than auto-parsing all chapters? For the default parser and manual listing on weird random websites, auto-parsing doesn't seem to work very well, gathering lots of irrelevant navbar links. It would help my workflow if I could paste story links directly, rather than fiddling with the UI to eliminate irrelevant chapters.
3. **Extension:** Edit chapter URLs requires specifying the title to use per chapter, as well as html tags, eg. `a href="">Title here</a>` - could it be changed to just take a list of URLs? e.g. instead of 
```
<a href="https://magic.wizards.com/en/news/magic-story/cowardice-hero-2014-01-22">18 Cowardice of the Hero</a>
<a href="https://magic.wizards.com/en/news/magic-story/emonberry-red-2014-02-05">19 Emonberry Red</a>
<a href="https://magic.wizards.com/en/news/magic-story/kioras-followers-2014-02-14">20 Kiora's Followers</a>
```
we could have
```
[...]

PR #1526, opened 2024-09-28
Improve edit chapter urls

gamebeaker:
reference:  #1452
"Edit Chapter URLs" accepts now pure URLs (one URL per line)
image
The title name is `[placeholder]` if the parser has implemented `findChapterTitle()` this name is used if `findChapterTitle()` isn't implemented `dom.title` gets used.
Problem:
You have to know that you can simple paste an URLs list. (no hint etc,)
@Darthagnon any ideas how to inform the users?

### DIFF
--- a/plugin/_locales/en/messages.json
+++ b/plugin/_locales/en/messages.json
@@ -255,6 +255,10 @@
 		"message": "CSS selector for element(s) to remove:",
 		"description": "Label in front input for CCS selector for elements to remove"
 	},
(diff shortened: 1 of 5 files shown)


Darthagnon:
A way to inform users could use overwritable background hint text (similar to that used in search fields) in the URL list edit view:
chrome_241020_164
(5min photo edit)

gamebeaker:
@Darthagnon thanks for the idea at the moment it looks like this:
image

PR #1558, opened 2024-10-25
Added MTLNovels support

### LINKED ISSUE #1545: mtlnovel.com url change
mtlnovel.com >> mtlnovels.com
ui looks similar to what it was before.
but just changing url in parser didn't seem to work

maforn:
Closes #1545 
I've added a new script and changed the way it fetches the chapters from MTLNovels, as of now they are not giving back the urls in a json anymore. Now the scripts scrapes the chapter list url and gets all the chapters from it.  
I've tested it and it works (for me), as specified in contributing all the tests pass, I've also added my name to the contributors list.  
Let me know if there are any further changes or tests that I need to make. By the way, thank you for your work, it's an amazing idea.

### DIFF
--- a/package.json
+++ b/package.json
@@ -72,7 +72,8 @@
     { "name": "nozwock"},
     { "name": "Darthagnon"},
     { "name": "LucasFreitaslpf1"},
-    { "name": "Jemeni11"}
+    { "name": "Jemeni11"},
+    { "name": "maforn"}
   ],
   "license": "GPL-3.0-only",
   "bugs": {
--- a/plugin/js/parsers/MtlnovelsParser.js
+++ b/plugin/js/parsers/MtlnovelsParser.js
@@ -1,11 +1,16 @@
 "use strict";
-parserFactory.register("mtlnovel.com", () => new MtlnovelParser());
+parserFactory.register("mtlnovels.com", () => new MtlnovelsParser());
+parserFactory.register("mtlnovel.com", () => new MtlnovelsParser());
+parserFactory.registerUrlRule(
+    url => (util.extractHostName(url).endsWith(".mtlnovels.com")),
+    () => new MtlnovelsParser()
+);
 parserFactory.registerUrlRule(
     url => (util.extractHostName(url).endsWith(".mtlnovel.com")),
-    () => new MtlnovelParser()
+    () => new MtlnovelsParser()
 );
 
-class MtlnovelParser extends Parser{
+class MtlnovelsParser extends Parser{
     constructor() {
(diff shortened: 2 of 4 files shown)


gamebeaker:
@maforn can you please remove the changes made to plugin/jszip/dist/jszip.min.js?

maforn:
Done, it updated automatically

dteviot:
@maforn 

You should remove the existing file MtlnovelParser.js and its reference in popup.html.  Either that, or edit MtlnovelParser.js and remove MtlnovelsParser.js.  Either is fine with me.
Also, if you can provide a commit that doesn't modify and then revert jszip.min.js that would be much appreciated.  Maybe its my OCD, but I dislike bloating Git with binary file changes.

dteviot:
@maforn 
And thank you for your hard work

maforn:
@dteviot 
I've already removed the jszip file, maybe this rule should be added to the .gitignore? I did not remove MTLNovelParser.js because it still works for mtlnovel.com ([example](https://es.mtlnovel.com/hello-changan/)). Do I still have to remove or rename it?

dteviot:
@maforn 

> I've already removed the jszip file changes,

You reverted the change.  There are two commits, one with the initial change, the second with jszip being reverted.  I'm suggesting using a --force-with-lease to cleanup the commit.  Or do a squash.

> maybe this rule should be added to the .gitignore

It's complicated.  How to handle the library goes round and round.
Personally, as a library, I think it should be either a sub-module or brought in via packages.  But that makes it difficult for newbies to build.

> I did not remove MTLNovelParser.js because it still works for mtlnovel.com

Ah.  My mistake, I changed MTLNovelParser.js, but did not commit it due to issues with site.
Note, that *.mtlnovel.com and *.mtlnovels.com seem to be the same site, just different aliases.  So, should be just one parser for both.

Also, site seems to still be broken.  If I browse to https://es.mtlnovel.com/hello-changan/, the chapters 1-250, 251-500 and 501-670 don't work.  (Clicking on them to expand shows nothing.)  And there's a comment dated two days ago complaining the site doesn't work.

maforn:
@dteviot
I've made all the changes that you requested, renaming the parser and removing the reverting commit. The only problem is that some novels in the website have the chapter list section duplicated, so sometimes it does show double the number of chapters. There is nothing we can do about that, it's just their website that is a bit broken.

dteviot:
@maforn 

Thank you again for your efforts.

maforn:
You're welcome! And thank you for your hard work.

Issue #1568, opened 2024-11-09
Please add site https://www.mvlempyr.com


Tenome:
### Provide URL for web page that contains Table of Contents (list of chapters) of a typical story on the site
https://www.mvlempyr.com/novel/my-house-of-horrors

### Did you try using the Default Parser for the site?  If not, why not?
Yes, but the manual scraper isn't grabbing the URLs for some reason despite showing in the preview. Something with the site blocking it?

### What settings did you use?  What didn't work?
* [Table of Contents](https://www.mvlempyr.com/novel/my-house-of-horrors)
* [URL of first chapter](https://www.mvlempyr.com/chapter/755-1)
* #chapter-content
* #span-28-164
* #image-2836-10

Leifman35:
I would really love it if this website was somehow supported by the scrapper. but it just gives you a "No parser found for this URL. Default Parser will be used. Please specify how to obtain content from each web page."

was going to open a 'site request' for it but searched beforehand and saw this was already requested. so i definitely would wanna join in this request

Leifman35:
Totally randomly, as i am quite clueless in 'how to make a parser' for this website (be it via clicking the bottom and following the instructions as the person that made this post originally or just in general) i think i may have done something that might help with making this website supported.

What i did is went on another supported website (novelbin) and specifically here: https://novelbin.com/b/bro-im-not-an-undead/ and after clicking the web2epub icon i clicked "Edit Chapter URL" and basically removed everything except the very first chapter url, and edited that one with the first chapter url from https://www.mvlempyr.com/novel/bro-im-not-an-undead (the site that we want supported) and basically just put 'https://www.mvlempyr.com/chapter/5384-1' into the <a href="https://novelbin.com/b/bro-im-not-an-undead/chapter-1-dead-forever">Chapter 1 Dead Forever</a> instead of https://novelbin.com/b/bro-im-not-an-undead/chapter-1-dead-forever.

It did managed to scrap the chapter from mvlempyr that way, only additionally adding some weird .gif at the end of the page/chapter. ill attach the .epub that it created but this sorta gives me hope that it is possible? maybe if you somehow configure the mvlempyr site paraser to act similarly to the novelbin one, with basically just somehow to make it grab the first/last chapter and what's between them, while also removing that .gif in the end of each chapter? not sure if it makes sense, but to my small clueless brain... it gives hope? lol. well, hopefully this is helpful and i didnt just ramble with nonsense

[...]

PR #1600, opened 2024-12-15
Add Site mvlempyr

ImLJS:
Solves Issue #1568 

One of the weirdest sites.

Notes:

1. Each Novel has a slug which is used to get the Novel Id, Chapter Count but don't know from where slug is being created. Had to take the slug from image source. This api call sometimes fail. So, had to use again chapter count div to get it
2. API call to get Chapter URL always give 404 for some reason. No time to check.

API Call to get ID : 	https://chp.mvlempyr.net/wp-json/wp/v2/tags?slug=6062 => {"id": 2021, "count": 190} and more
API Call to get Chapters : https://chp.mvlempyr.net/wp-json/wp/v2/posts?tags=2021&per_page=500&page=1 => Gives JSON error

@dteviot can you check

Code is working fine but is there any better implementations?

### DIFF
--- a/plugin/js/parsers/MvlempyrParser.js
+++ b/plugin/js/parsers/MvlempyrParser.js
@@ -1,0 +1,54 @@
+"use strict";
+
+parserFactory.register("mvlempyr.com", () => new MvlempyrParser());
+
+class MvlempyrParser extends Parser {
+    constructor() {
+        super();
+    }
+
+    async getChapterUrls(dom) {
+        let imgLink = dom.querySelector("div.novel-image-wrapper img").src;
+        let slug = imgLink.split("/").pop().split(".")[0];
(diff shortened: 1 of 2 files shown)


gamebeaker:
@ImLJS The "problem" is that the index Chapter names are Chapter 1/2/... and not the real name.
Example: https://www.mvlempyr.com/novel/my-house-of-horrors
image

https://github.com/dteviot/WebToEpub/blob/cadde6852c3e7d1026dba31634cb7eb36ff59efa/plugin/js/parsers/MvlempyrParser.js#L17-L23
If you change line 21 to: 
`title: "[placeholder]"`
WebToEpub will change the [placeholder] to what findChapterTitle() found.
You can test it manually:
image
It looks like this:
image
@ImLJS  your decision if you want to change it.

PR #1608, opened 2024-12-20
Update Inoveltranslation

Inareous:
The site doesn't use different api path anymore. For this purpose I've removed all locked chapters as well.

Tested with https://inoveltranslation.com/novels/852e979c-fcf0-486c-aa9d-7ec39571a305

### DIFF
--- a/plugin/js/parsers/InoveltranslationParser.js
+++ b/plugin/js/parsers/InoveltranslationParser.js
@@ -8,38 +8,25 @@ class InoveltranslationParser extends Parser{
     }
 
     async getChapterUrls(dom) {
-        let novelId = dom.baseURI.split("/").pop();
-        let tocUrl = `https://api.inoveltranslation.com/novels/${novelId}/feed?limit=100000`;
-        return (await HttpClient.fetchJson(tocUrl)).json.chapters
-            .sort((a, b) => a.volume === b.volume ? a.chapter - b.chapter : a.volume - b.volume)
-            .map(c => this.jsonChapterToChapter(c));
+        return [...dom.querySelectorAll("section[class^='styles_chapter_list'] div:has(>a):not(:has(> div))")]
+            .map(link => this.linkToChapter(link)).reverse();
     }
 
-    jsonChapterToChapter(jsonChapter) {
-        // return wrong hostname, to avoid additional parserFactory registration issues
-        // fixup in fetchChapter()
+    linkToChapter(link) {
+        let a = link.querySelector("a");
+
         return ({
-            title: this.titleFromJson(jsonChapter),
-            sourceUrl: `https://inoveltranslation.com/chapters/${jsonChapter.id}`
+            sourceUrl: a.href,
+            title: a.textContent,
         });
     }
 
-    titleFromJson(jsonChapter) {
-        let title = "";
-        title += util.isNullOrEmpty(jsonChapter.volume)
-            ? ""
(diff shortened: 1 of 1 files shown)


ImLJS:
@Inareous After going through your PR, slight changes are required.

1. getChapterUrls Function

`return [...dom.querySelectorAll("section[class^='styles_chapter_list'] a")].map(link => this.linkToChapter(link));`

`linkToChapter(link) {
        return ({
            sourceUrl:  link.href,
            title: link.textContent,
        });
    }`

It is much cleaner this way

2. FindContent Function

`    findContent(dom) {
        return (
            dom.querySelector("section[data-sentry-component='RichText']")
        );
    }`

Again cleaner option

3. Add getInformationEpubItemChildNodes function at the end

`getInformationEpubItemChildNodes(dom) {
        return [...dom.querySelectorAll("section[class^='styles_details_container'] dl:last-child")];
    }`

After making changes, I can merge your code

Inareous:
@ImLJS

Thanks on the css selector and the information func, I've slightly adjusted to still remove the locked chapters ya.

I think the content preserve the styling as well now, which seems to be a little bit off for Calibre but looks fine on my Kindle. Should be of no issue.

ImLJS:
@Inareous Code looks fine.

After Testing, found that in some chapter layouts it fails

- https://inoveltranslation.com/novels/51cfa3b4-934d-4e28-a4e1-5eeaf9e68da4 - Passes when all chapter are present 
- https://inoveltranslation.com/novels/98cd9efb-7f7c-41e9-b395-adc8c17ff92a - Fails when Chapters are divided into volumes

Can you look into it. 

Also add your name in credit section of readme.

Inareous:
The chapters layout seems to be lazy loaded on click for volumes, where the data is saved in a script tag.

From what I see on
1. https://edspencer.net/2024/7/1/decoding-react-server-component-payloads
2. https://github.com/vercel/next.js/discussions/42170#discussioncomment-8137079

The best action would be to create a RSC payload parser and get the Object content properly, which would be outside my skills.

This PR would keep the volume parsing broken as I'm don't know how to fix it, will close the PR in a day or two if its not acceptable to parse only the latest volume ya.

A simple hack like a regex would suffice but it would be not proper and can break anytime. Sample in the branch https://github.com/Inareous/WebToEpub/tree/ExperimentalTabMode-InovelTranslation-RCS_Payload.

```js
    async getChapterUrls(dom) {
        let chapters = []
        let chapterRegex = /\\"href\\":\\"(.+?)\\".*?\\"children\\":\\"(.+?)\\"/g
        let text = [...dom.scripts].filter(x => x.text.includes("ChapterList"))?.[0]?.text

        let match;
        while ((match = chapterRegex.exec(text)) !== null) {
            chapters.push({
                sourceUrl: match[1],
                title: match[2],
            })
        }

        return chapters.reverse()
    }
```

ImLJS:
@Inareous I looked that it for some time, it is complicated as heck. So, we can just take the latest volume. 

Also one eslint error is there ->         
`let a = link.querySelector('a');`  -> Turn single quotes to double quotes

Also add your name in credit section of readme. Then I can merge your PR.

Thanks for your Hard Work.

Inareous:
Updated ya, thank you @ImLJS, would it be okay to omit the credit section? I'll update later when I got the full parser working.

Thanks for your hard work as well.

dteviot:
@Inareous @ImLJS 
I thank both of you for your hard work

Issue #1619, opened 2024-12-30
Multi-page support? Please add site https://www.wxscs.com/book/1597/


phazei:
Please note, I'm basically the only developer working on WebToEpub, and I'm not paid for doing this.  (WebToEpub is completely free, and generates no money.) By asking to add a site, you're asking me to give you some of my limited free time. So, I think it's not unreasonable for me to ask you to do as much as you can to help me.

### Did you try using the Default Parser for the site?  If not, why not?
I did use the default parser, it sort of works.

### What settings did you use?  What didn't work?
* URL of first chapter
** https://www.wxscs.com/book/1597/222174.html
* CSS selector for element holding content to put into EPUB
** "#cont-body"
* CSS selector for element holding Title of Chapter
** "h1.cont-title"
* CSS selector for element(s) to remove
** "script"

### If the Default Parser did not work, if you have developer skills, did you try writing a new parser?
I have lots of dev skills

So, adding a parser that follows the sites structure is generally simple, but this site decided to put everything on 2 pages, so each chapter that's clicked, you need to click page 2.  I don't know how common that is or if there are any other parsers that do that.  If there is a parser that manages that, then I could use that as a template to create a new one.  If not, then it'd be a new feature that probably isn't as clear cut with the built in page loop.  Since the index doesn't have the pages, I'd guess that the extra URL to parse would have to be added on the fly which could be difficult, I haven't looked at the code.

For this site, the second url could be found via the css selector "ul.pagination li:nth-child(2) a" but if you're already on the second page, then we'd want to skip it.  It would probably be easier if we knew for certain if there's only ever 2 pages, ~I haven't seen one with 3 pages, but I guess there could be.~ https://www.wxscs.com/book/15207/5177007.html has 3 pages

So, on that particular site, the pages are always rootpage_{#}.html, so:
https://www.wxscs.com/book/15207/5177007.html
https://www.wxscs.com/book/15207/5177007_2.html
https://www.wxscs.com/book/15207/5177007_3.html
So it could always check if _2 exists, and if it's not a 404, add it to the current chapter.

Really I'd just find a different site with the series, but I can't find one.

dteviot:
@phazei 

> So, adding a parser that follows the sites structure is generally simple, but this site decided to put everything on 2 pages, so each chapter that's clicked, you need to click page 2. I don't know how common that is or if there are any other parsers that do that.

[...]

PR #1620, opened 2024-12-30
Add Site wxscs.com

phazei:
This addresses this issue:
https://github.com/dteviot/WebToEpub/issues/1619

It adds support for https://www.wxscs.com with multi-page support.

I used WattpadParser as a reference since it was the only one I could find that supported multi-page.

I'm continually impressed with Claude Sonnet, I gave it about 2 dozen parsers, top/bottom 12 based on size, asked it if anything had multi-page support, had a discussion on how it worked, gave it the html for the index and a random 3 page chapter, and it one shot a completely working extension.  I've been a dev for 20 years, and it blows my mind every time I do something like this.

I tested it out on a few books on the site, worked great.

### DIFF
--- a/package.json
+++ b/package.json
@@ -73,7 +73,8 @@
     { "name": "Darthagnon"},
     { "name": "LucasFreitaslpf1"},
     { "name": "Jemeni11"},
-    { "name": "maforn"}
+    { "name": "maforn"},
+    { "name": "phazei"}
   ],
   "license": "GPL-3.0-only",
   "bugs": {
--- a/plugin/js/parsers/WxscsPaser.js
+++ b/plugin/js/parsers/WxscsPaser.js
@@ -1,0 +1,195 @@
+"use strict";
+
+parserFactory.register("wxscs.com", () => new WxscsParser());
+
+class WxscsParser extends Parser {
+    constructor() {
+        super();
+        this.state.indexDom = null;
+        this.state.indexUrl = null;
+    }
+
+    async onLoadFirstPage(url, firstPageDom) {
+        // If we're on a chapter page, fetch and cache the index page
+        if (!firstPageDom.querySelector("div#all-chapter")) {
+            let bookId = this.extractBookId(firstPageDom);
+            if (bookId) {
+                this.state.indexUrl = `https://www.wxscs.com/book/${bookId}/`;
+                this.state.indexDom = (await HttpClient.wrapFetch(this.state.indexUrl)).responseXML;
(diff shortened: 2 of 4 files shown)


phazei:
I noticed before it only worked when I clicked the plugin when on the index page.  I fixed it so it works on the chapter page too.

dteviot:
@phazei 
Thank you very much for your hard work.

It's missing an implementation of: 
* getInformationEpubItemChildNodes()  => Provides summary/description on information page
* findCoverImageUrl()

Also, please add your name to readme.md and package.json and squash the commits.

> I'm continually impressed with Claude Sonnet, I
Can you provide the script?  Or better yet, add a page to the wiki describing what you did?  If it can help others, that could save me and the others a lot of work.

dteviot:
@phazei 
> used WattpadParser as a reference since it was the only one I could find that supported multi-page.
See:
https://github.com/dteviot/WebToEpub/wiki/FAQ#solutions-to-site-issues-that-require-special-coding

phazei:
I updated the requested changes and squashed the commits.

I wanted to note that the getFirstImgSrc helper doesn't work if the selector is the img element itself.

Also, there are some things that aren't available on anything but the index page, like the book cover, and description.  If someone clicks the plugin while on a chapter page, it mostly works, but it tries to grab stuff from the wrong pages.  For the chapters, it's done async, so I can go and fetch whatever.  But everything else is synchronous.  I did what I could to check if it's the index page and if not, grab it and cache the DOM of the index page so the items that are on it can be grabbed without grabbing it over and over again.  But the onLoadFirstPage isn't called via "await".  I think if it were, it would allow the parsers to have more time to set up anything unique that might be on a site.  Right now that can only be done when grabbing the chapters, but by then everything else has already been called.

dteviot:
@phazei 

Thank you for your efforts.

> I wanted to note that the getFirstImgSrc helper doesn't work if the selector is the img element itself.

Please raise an issue

>Also, there are some things that aren't available on anything but the index page, like the book cover, and description. If someone clicks the plugin while on a chapter page, 

I have stopped trying to make WebToEpub work when a ToC is not supplied.  (IMHO Not worth the effort to write, test and maintain.)  If it worries you, I'd suggest checking WebToEpub has been given a ToC in getChapterUrls(), and if not, give user a warning message.

dteviot:
@phazei 

Additional, there were 2 eslint warnings in the text you supplied.  I've pushed a fix.

Issue #215, opened 2018-10-21
Improvement: RAM Allocation


lehur7:
I using webtopeub on chrome (latest version) for lnmtl.com for 3 or 4 novels at the same time and then it become like snail (and come out warning for wait or kill tab option), albeit chrome's other tabs work normal because my RAM still 40% usage.
*Perhaps chrome maintains the extension for RAM usage as well as RAM each tabs used, could you bypass it and used all RAM directly

3dycosmo:
have been implemented on #304 
and webtoepub also adjusting to delay packaging an epub if many webtoepub runs so not to make packaging epubs at the same time for multiple epubs

dteviot:
With regards to
> could you bypass it and used all RAM directly

No. Chrome extensions have no control over memory allocation.  In fact, they can't even tell how much memory is available or being used.  Probably for security reasons.  (Actually, this isn't quite true.  If you're running the developer version of Chrome there's an API, but most people don't run that version of chrome.  And it's not available at all on Firefox.)  Note, you can see how much memory WehToEpub is using in the Chrome Task Manager.  See: https://www.howtogeek.com/437681/how-to-use-chromes-built-in-task-manager/  Note, if you open multiple copies of WebToEpub, each one is allocated its own block of memory.

dteviot:
Looks like I'm wrong again.  At least on Chrome I can find out how much heap memory is (a) being used and (b) is available. https://developer.mozilla.org/en-US/docs/Web/API/Performance/memory

dteviot:
Unfortunately, Performance.Memory is pretty much useless for my purposes.  From the code 
https://bugs.webkit.org/attachment.cgi?id=154876&action=prettypatch
>  Web pages can only learn new data every 20 minutes

dteviot:
There's a new API that might be useful performance.measureMemory().
Should investigate further.
https://github.com/WICG/performance-measure-memory
https://web.dev/monitor-total-page-memory-usage/

Dimava:
I think there is one and only one thing that can reduce memory consumption in HTML 
*Removing the required node from the document* (with `el.remove()`)
(then the #document will be completely garbaged if not keeped elsewhere)
well, or cloning the required node with `el.cloneNode(true)` and garbaging document yet again

dteviot:
@Dimava  
It's more an issue of the extension not being aware that it's running out of memory, so can halt with a useful message.
Instead, it just crashes, with no warning to user.
[...]

Issue #1665, opened 2025-02-03
Royal road - download issues


shreyassv:
**Describe the bug**
Sometimes , even though there are no errors , and the plugin shows that all chapters have been downloaded, the file is not saved.

**To Reproduce**
1. Go to [royalroad fic'](https://www.royalroad.com/fiction/82881/madman-apocalypse)
2. Use webtoepub to create ebook.
3. after all chapters are worked on, the file is not created- webtoepub appears to be stuck in a loop.
4. inspect shows the following error-uncaught in promise - range error maximum call stack size exceeded

**Expected behavior**

**Screenshots**

**Desktop (please complete the following information):**
 - OS: windows 11
 - Browser chrome

dteviot:
@shreyassv 

Can you figure out the conditions needed to consistently reproduce the problem?

shreyassv:
Selecting skip images and deselecting the other two image related checkbox solves the problem.

This issue has occurred with images downloaded for 2 royalroad stories for me. 
The one linked in the issue is one. I don’t remember the other. I’ll find and upload the other later.

The issue is that most stories have the image as part of the story(status etc)

shreyassv:
@dteviot 

Selecting skip images and deselecting the other two image related checkbox solves the problem.

This issue has occurred with images downloaded for 2 royalroad stories for me.
The one linked in the issue is one. I don’t remember the other. I’ll find and upload the other later.

The problem is that most stories have the image as part of the story(status etc)

gamebeaker:
I can reproduce the error:

![Image](https://github.com/user-attachments/assets/7be3abc2-a007-4123-b15e-1cf1b615950c)

I think there exists another issue with the same problem but i can't find it.

The Network tab:

![Image](https://github.com/user-attachments/assets/09f34313-c42d-4da7-8aac-b3158bb82aef)

dteviot:
@shreyassv 

> The problem is that most stories have the image as part of the story(

Sounds like the problem is the images are taking up too much memory. 
Possible solution would be for you to download the story in blocks of 50 chapters or so at a time.  Then merge the chapters together.

gamebeaker:
@shreyassv the problem is that there is a gif that is 48 MB big and this is to big in the first Chapter at the bottom.

dteviot:
Note to self, look at replacing jszip with archiver https://npm-compare.com/adm-zip,archiver,jszip,zip-local

gamebeaker:
[...]

Issue #1514, opened 2024-09-26
Fix code that generates warning when submit to Mozilla.


dteviot:
> The Function constructor is eval.
Warning: Evaluation of strings as code can lead to security vulnerabilities and performance issues, even in the most innocuous of circumstances. Please avoid using `eval` and the `Function` constructor when at all possible.

jszip/dist/jszip.min.js line 13 column 96498

Has been seen elsewhere.  https://github.com/FooSoft/yomichan/issues/828#issuecomment-695272233
Not sure why it's showing now.

gamebeaker:
@dteviot as a "hack" you can just delete the offending part in jszip.min.js and it still works. (More tests needed)
Why do i think this is a valid solution? In my opinion the browser would block the js execution with the content_security_policy if WebToEpub should try to execute it as https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/manifest.json/content_security_policy is not set to 'unsafe-eval'.
That is why i think this part of the code is never used in WebToEpub and can be deleted.
Replace in jszip.min.js `&&(e=new function(""+e))` with nothing.
![image](https://github.com/user-attachments/assets/11c9fd00-6051-4371-bc58-4ebc1b7abdd8)
![image](https://github.com/user-attachments/assets/29c13347-8f35-45a7-960d-5c4e6c577e40)
save the file and try `web-ext lint` it should now have no warnings.

dteviot:
@gamebeaker 
> as a "hack" you can just delete the offending part in jszip.min.js and it still works.

That's not a solution.
The problem is, the code generates a warning message when file is submitted to Mozilla to go on the store.  (Later on, this might become, extension is rejected.  As it is, it's posting a warning the submission may be manually inspected and rejected.)

Modifying the jszip.min.js is not feasible because it's a minified file.  That's only acceptable by Mozilla because it's a copy of a known library.  Needed to specify the library so it can validate file is same.  If we modify the code, it will no longer match.

gamebeaker:
As jszip is no longer in active development(2 years since last update) an other zip packager is needed. Here is a comparison per downloads 
https://npmtrends.com/adm-zip-vs-archiver-vs-jszip-vs-node-zip-vs-zip

dteviot:
A third option, include source of jszip, rather than minified version.  And remove the offending bit.  But I'll take a look at the other zip packagers.

gamebeaker:
[...]

Issue #179, opened 2018-06-02
Out of Memory


minekasetsu:
I don't think this is what the extension is designed for, but I used it on lparchive.org, specifically https://lparchive.org/Legend-of-Mana-(by-Mega64)/ . The chapters was filled with a lot of gameplay pictures and animations, so they ended up being really big, but initially I found success making the first 18 chapters into a book (132mb). But when I tried to make the next 24 chapters into a book, It returned "out of memory", after all of the urls being processed. I tried to reduce the number of chapters to 18 so it's not much different from the previous one, but It still returned "out of memory".

I decided to manually end the process of that specific firefox tab (it took 300~500mb of memory even after the failure, according to task manager), closed the tab, and open it at a new tab. Then I find successs in making the book (18 chapters, 163mb).

I also found some other failure if I simply refreshed the tab instead of close and open it at a new tab, but I forgot what it said and I'm not willing to recreate the scenario. Sorry.

Additionally it might be nice if I can skip analyzing the starting url and immediately put the chapter urls. But I suspect my case was highly specific, and so the gain would be marginal at best.

dteviot:
@minekasetsu
I'm sorry, but I don't know what you're asking for.
If you're trying to say, "The extension can run out of memory and crash".  Then thank you, I'm aware of that.  Unfortunately, there's not a lot I can do about it.
If you're telling me that "Refreshing a tab does not release all memory".  Thank you.  I'll look into that.
If it's something else, please tell me what it is you'd like me to do.

Synteresis:
Note to Self

Look at https://github.com/jimmywarting/StreamSaver.js?

dteviot:
Note to self:  https://wicg.github.io/file-system-access

PR #1692, opened 2025-03-01
change jszip to zip.js

gamebeaker:
reference: #1665
there are other issues that are about changing jszip.

- [x] Pack EPUB works (tested with https://www.royalroad.com/fiction/82881/madman-apocalypse from #1665)
- [ ] Is npm correct? (Help wanted)
- [x] Libarary works
- [x] Update DarkNovelsParser.js to zip.js
- [x] Change test
- [x] fix eslint error (Help wanted)
- [x] squash commits
- [x] is zip compression working (i am not sure if deflate is currently used)
- [x] change jszip mentioning in readme

hint:
zip-no-worker.min.js terminates without error if not the option {useWebWorkers: false} is defined this is so that it doesn't work in local dev instead of working in dev env and not working if shipped.
Example:
If zip.min.js is used it is possible to write code that works in dev env but not in prod. see https://github.com/dteviot/WebToEpub/pull/1692#issuecomment-2692921796


gamebeaker:
image
oh no...

gamebeaker:
@dteviot this version works without jszip it uses zip.js instead.
Tested with:
https://www.royalroad.com/fiction/64916/hell-difficulty-tutorial
on:
Firefox-android
Firefox 135.0.1 (64-bit) 
Chrome 133.0.6943.142

Optimizations/ cleaner code in the Libarary are possible e.g. LibConvertDataUrlToBlob() should no longer be neccassary as the zip can be created from base64.

I don't really know node.js so maybe have a look there i left jszip as dependency to build the release zip files with pack.js

Is https://github.com/dteviot/WebToEpub/tree/ExperimentalTabMode/doc actively mantained or can this folder be deleted? I found these docs as they mention "jszip"

following are issues that get fixed with this pull request:
#1514
#179 (The generated epub >700 MB)

I guess this issue is related but not sure if they are fixed:
#215

gamebeaker:
The last commit fixes issues with the Library download if the epub has a large size.
I think that problem is that zip.js can generate larger epubs to save in the Library and if the base64 gets converted into a string with atob the max string size is exceeded and it terminates.

The alternative is pretty slow compared to atob, that's why it is in try {} catch {} to make it faster in "normal" cases.
Bug found and tested with: https://www.royalroad.com/fiction/82881/madman-apocalypse

gamebeaker:
@dteviot please do you merge this change.
I don't want to merge it as it changes a dependency and i don't know if i handled npm correct.

dteviot:
@gamebeaker 

Firstly, thank you for your very hard work.
Second, I didn't look at your changes to the Library.js file.  (Too big, and I don't know that code.)

Lastly, a few minor points.
1. I don't think jszip should be here
https://github.com/dteviot/WebToEpub/blob/b199e75052bdad156fa35b05a2f39c360e14473b/package.json#L5-L7

2. I think you might have committed https://github.com/dteviot/WebToEpub/pull/1692/commits/8ddc3e0ad89a163e8aeaa77b5ddad5edd9cd871e by mistake.  That's adding **jszip**, which you are removing.

3. Line 355 of this code might have a problem with images
https://github.com/dteviot/WebToEpub/blob/b199e75052bdad156fa35b05a2f39c360e14473b/plugin/js/EpubPacker.js#L352-L361

When the file is an image, content is an arrayBuffer.  It's a string when the file is an XML document.  e.g.
https://github.com/dteviot/WebToEpub/blob/b199e75052bdad156fa35b05a2f39c360e14473b/plugin/js/EpubItem.js#L50-L60

https://github.com/dteviot/WebToEpub/blob/b199e75052bdad156fa35b05a2f39c360e14473b/plugin/js/EpubItem.js#L197-L199

I should have renamed packXhtmlFiles().  When first written, it just did XML files.  Then the code got updated to handle images, but wasn't renamed.  

I'd suggest changing Lines 50 to 60 of EpubItem.js to return an arrayBuffer, then you can update packXhtmlFiles() to handle an arrayBuffer in all cases.

Other that those things "Looks good to me".  Go ahead and commit.

>I don't want to merge it as it changes a dependency and i don't know if i handled npm correct.

It builds for me, so I think you did it correctly.
I should say, I'm not totally sure how the NPM part works myself.  I originally used git submodules to bring in the jszip library.  But someone else wrote the NPM bit and removed the submodule, on the basis that the NPM would bring it in from the packages.json file.  So, don't duplicate things.  Then someone else went and committed the jszip library.  So, a bit of a painful mess.

Notes
Time taken: 164 minutes

gamebeaker:
>1. I don't think jszip should be here
>2. I think you might have committed https://github.com/dteviot/WebToEpub/commit/8ddc3e0ad89a163e8aeaa77b5ddad5edd9cd871e by mistake. That's adding jszip, which you are removing.

That is correct as jszip is needed for eslint to package the zip and xpi file. If i remove jszip as dependencie 'npm run lint' will not work after a fresh 'npm install'.

>3. Line 355 of this code might have a problem with images

I tested with images and gifs and didn't have any problem i would keep it like this as it is working. My concern with changing the text to an arraybuffer are that it could maybe destroy encoding as zip.js needs Uint8Array which is an array of 8bit and that is only ASCII. [UTF-8](https://en.wikipedia.org/wiki/UTF-8) uses [Variable-width encoding](https://en.wikipedia.org/wiki/Variable-width_encoding) and i fear that this will break.

gamebeaker:
I merged with the squash and merge button from github and the result looks good in the git history. 
image
For the future if there are pull requests with multiple commits they can be squashed with one click.
image

dteviot:
@gamebeaker 

Again, Thank you for your efforts.

>That is correct as jszip is needed for eslint to package the zip and xpi file. If i remove jszip as dependencie 'npm run lint' will not work after a fresh 'npm install'.

Interesting.

>I tested with images and gifs and didn't have any problem i would keep it like this as it is working. My concern with changing the text to an arraybuffer are that it could maybe destroy encoding as zip.js needs Uint8Array which is an array of 8bit and that is only ASCII. [UTF-8](https://en.wikipedia.org/wiki/UTF-8) uses [Variable-width encoding](https://en.wikipedia.org/wiki/Variable-width_encoding) and i fear that this will break.

If it works, it works.  That said, and thinking about it some more, there's not really any reason to try to compress images, as they're already compressed.  Trying to compress them any more doesn't really work.  So, images could be stored not compressed.  (Should make packing faster, and hopefully use less memory during the zip creation process.) So, change would be change fileContentForEpub() into packContentFileIntoEpub(), pass in the zipWritter, and make the two implementations pack differently.  I may experiment and see what happens.

Issue #1688, opened 2025-02-27
Docs missing: How to install from Source (for people who are not developers)


gamebeaker:
The link https://github.com/dteviot/WebToEpub/tree/ExperimentalTabMode#user-content-how-to-install-from-source-for-people-who-are-not-developers
is no longer working.
@dteviot Needs to be changed in future complete messages.

Needs to be changed here https://github.com/dteviot/WebToEpub/releases/tag/developer-build

dteviot:
Note
Corrected link is https://github.com/dteviot/WebToEpub/tree/ExperimentalTabMode#user-content-installation

dteviot:
Tagging so I know to fix this once zip.js has been merged

gamebeaker:
Test versions for Firefox and Chrome have been uploaded to https://github.com/dteviot/WebToEpub/releases/tag/developer-build. Pick the one suitable for you, follow the "How to install from Source" instructions at https://github.com/dteviot/WebToEpub/tree/ExperimentalTabMode#user-content-installation and let me know how it goes.

PR #1699, opened 2025-03-07
fix developer build instructions link

gamebeaker:
reference: #1688


dteviot:
@gamebeaker Thank you

Issue #1701, opened 2025-03-08
Fanfiction.net sometimes randomly fails to load chapters


Kiradien:
This bug report is more for archival purposes; the fix will be uploaded shortly.

Fanfiction.net has 3 scenarios where chapters fail to load, all of which occur server side - From what I can see, it appears to be an MVC controller forward failure, cache failure, server failure.

Two of these scenarios can be covered in error handling - current code only handles cache failure which just happens to cover most cases of controller failure as well. In the scenario where the server cache is too recent but controller forwarding is still failing to forward a specific chapter, the entire load can stall out.

The code attached to this bug report is to work around server side issues and can be removed if FFNet/GoDaddy fixes the source of the issues on their side.

PR #1702, opened 2025-03-08
#1701 Update FanFictionParser.js

Kiradien:
Updated minimal throttle to account for slightly faster computers & connections making requests at min accepted threshold. Updated failure regex to account for secondary failure condition Created secondary forwarding condition.
Updated fetchWebPageContent method to proper recursion instead of invoking super


dteviot:
@Kiradien 

Thank you for your work.

Issue #1581, opened 2024-11-26
Function to add multiple URL to 'Story URL' in Library


Demaar:
**Is your feature request related to a problem? Please describe.**
When I download stories from forums like SB many of them have Sidestories or Character Sheets that I might want in the Epub but aren't in the Threadmarks like the main story but in others like Sidestory, Apocrypha, Media, Informational, etc.
Or stories long enough that they are split among multiple Threads.
I normally solve this by opening the other Category in W2E and copying the URL over to the main story's.
These Extra Chapters, while saved in the library, will not be checked by the 'Search new Chapters' button.

**Describe the solution you'd like**
Some designated Symbol, maybe just a space, that can't be part of any URL that tells W2E that there is another URL to check.

**Additional context**

![{A6484677-A35D-4CD5-A9B5-28990D50F3E1}](https://github.com/user-attachments/assets/00fd8fc6-72a6-4563-8d80-7f45e896fba4)
Pressing the 'Search new Chapters' button gives this, even though the associated Epub has over 60 Chapters.

gamebeaker:
As i understand it you created 1 epub from multiple stories. Can't you just create multiple epubs -> multiple Library entries?

Demaar:
> As i understand it you created 1 epub from multiple stories. Can't you just create multiple epubs -> multiple Library entries?

More like that the story is split among multiple URL and I have to manually stitch it together.

gamebeaker:
@Demaar can you please share an example with link?

Demaar:
[Here](https://forums.spacebattles.com/threads/my-next-life-as-a-supervillain-all-routes-lead-to-doctor-doom-hamefura-mcu.1164297/) is the Story Page, [Here](https://forums.spacebattles.com/threads/my-next-life-as-a-supervillain-all-routes-lead-to-doctor-doom-hamefura-mcu.1164297/threadmarks) is the Main Chapters Page and [Here](https://forums.spacebattles.com/threads/my-next-life-as-a-supervillain-all-routes-lead-to-doctor-doom-hamefura-mcu.1164297/threadmarks?threadmark_category=13) is the SideStories Page.
The Goal is to be able to Save the chapters from the Main Chapter Page and the ones from the SideStories Page as one library entry so that the 'Search new Chapters' button scans both URL.

gamebeaker:
@Demaar 
I will tag it as Complete (It will be automatically closed in a few weeks).
[...]

PR #1711, opened 2025-03-12
Add Library update all

gamebeaker:
reference: #1581 
This adds an button which tries to update all books in the Library and adds an indication if new Chapter got downloaded.
grafik

To realize this i added Promises to multiple functions to know when the TOC is fully loaded/ The EPUB finished downloading.
@dteviot do you know of an more elegant way or is this ok?

I also changed an eslint rule "no-async-promise-executor":0

### DIFF
--- a/eslint/.eslintrc.js
+++ b/eslint/.eslintrc.js
@@ -28,6 +28,9 @@ module.exports = {
             "off",
             "never"
         ],
+        "no-async-promise-executor": [
+            0
+        ],
     },
 	"globals": {
 		"JSZip": "readable",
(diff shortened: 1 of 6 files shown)


dteviot:
@gamebeaker 
>do you know of an more elegant way

Declare the functions as async?

According to https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/async_function
>Each time when an async function is called, it returns a new [Promise](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise) which will be resolved with the value returned by the async function, or rejected with an exception uncaught within the async function.

Might need to refactor the internals to use await, and add try/catch handling.  But I think that would improve the readability of the functions.  I did have a thought in the back of my mind to replace promises with async/await at some time.

gamebeaker:
@dteviot i changed it to async/ await and it looks like it works.

Issue #1017, opened 2023-07-21
Library search new chapter not functioning in webnovel


Veaxale:
I had uploaded the epub file into the library.
When i clicked a search function its came out with nothing, i had already filled the url for it tho.

dteviot:
@Veaxale 
> When i clicked a search function

What search function?  Please describe

gamebeaker:
I think the problem is that if someone uses "Add to Library" with a novel from [https://www.webnovel.com](https://www.webnovel.com) like [https://www.webnovel.com/book/military-system-in-isekai-of-magic_22815989905323705](https://www.webnovel.com/book/military-system-in-isekai-of-magic_22815989905323705) and then tries to search for new Chapters no new Chapters get found.
Problem:
Go to: [https://www.webnovel.com/book/military-system-in-isekai-of-magic_22815989905323705](https://www.webnovel.com/book/military-system-in-isekai-of-magic_22815989905323705)
Open WebToEpub -> no Chapters found
Close WebToEpub
Click in novel Tab "Table of Contents"
Open WebToEpub -> Chapters are found
If you now use "Add to Library" the "Story URL" is [https://www.webnovel.com/book/military-system-in-isekai-of-magic_22815989905323705](https://www.webnovel.com/book/military-system-in-isekai-of-magic_22815989905323705) and if you click "Search new Chapters" the "Story URL" is used with "Load and Analyse" function and you can't find any Chapters like the first time WebToEpub got opend before "Table of Contents" got clicked on the website.

Solution:
Append "/catalog" on the link if it isn't there.
Change the "Story URL" to [https://www.webnovel.com/book/military-system-in-isekai-of-magic_22815989905323705/catalog](https://www.webnovel.com/book/military-system-in-isekai-of-magic_22815989905323705/catalog) and than click "Search new Chapters" now it works.

gamebeaker:
@Veaxale 

Test versions for Firefox and Chrome have been uploaded to https://github.com/dteviot/WebToEpub/releases/tag/developer-build. Pick the one suitable for you, follow the "How to install from Source" instructions at https://github.com/dteviot/WebToEpub/tree/ExperimentalTabMode#user-content-installation and let me know how it goes.

dteviot:
@Veaxale 

Updated version (1.0.4.0) has been submitted to Firefox and Chrome stores.
Firefox version is available now.
Chrome might be available in a few hours (typical) to 21 days.

Closing, will reopen if there are problems.

PR #1712, opened 2025-03-12
Update webnovel.com

gamebeaker:
reference: #1017 
fixes Library search function for webnovel.com


dteviot:
@gamebeaker 
Did you need to use regexs here?
I think string.endsWith https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/endsWith and concatenation would do the job.

gamebeaker:
@dteviot
> Did you need to use regexs here?

apparently not but i like regex and i automatically think: string manipulation => regex

Issue #1430, opened 2024-08-20
unable to DL some chapter https://wtr-lab.com/en/serie-8003/attributes-are-infinitely-increasing-i-dominate-multiple 


Finks82:
https://wtr-lab.com/en/serie-8003/attributes-are-infinitely-increasing-i-dominate-multiple 

ch 350+ unable to download

gamebeaker:
more context: 
![image](https://github.com/user-attachments/assets/b3e5e1ba-2a35-4b10-9d55-ea45a7ed4bdf)
It looks like this website translates the Chapter on the fly the first time someone opens it. After the first translation it gets cached on the server and WebToEpub can get them. The Chapter needs to be opened in a tab before WebToEpub can download it. I think after Chapter 360 you can see this problem.
Here are screenshots how it looks if you are the first person to open a Chapter:
![image](https://github.com/user-attachments/assets/078b3bd2-7fe5-4fb3-91f3-491ac96cf5a4)
![image](https://github.com/user-attachments/assets/dd68a7fc-d343-4452-8b3d-61bdf5dbd8d3)
![image](https://github.com/user-attachments/assets/32b30ed9-6fb0-4b64-87cb-705c1cbe7e07)

Phaethonix:
They're adding AI to this now. I don't think this problem can be fixed on the extension's end no more. It's good quality, sure, but man, I just want to listen to them instead of reading.

vigneshsarma:
Making this change lets the extensions still work to download epub. 
```diff
modified   plugin/js/parsers/WtrlabParser.js
@@ -10,7 +10,7 @@ class WtrlabParser extends Parser{
     async getChapterUrls(dom) {
         let items = [...dom.querySelectorAll("div.accordion a.chapter-item")];
         return items.map(a => ({
-            sourceUrl:  a.href,
+            sourceUrl:  a.href + '?service=google',
             title: this.formatTitle(a)
         }));
```

I dont thing this is the right solution though.

gamebeaker:
@vigneshsarma you can make a pull request if you have a solution.

gamebeaker:
@Finks82 problem got fixed. Each chapter that is new has a delay from 10s.
[...]

PR #1740, opened 2025-03-20
Update WtrlabParser.js to always use google for translation

vigneshsarma:
#1430 wtr-lab now has AI translation which is causing the download failure.

Issues not resolved by this PR:
- wtr-lab also support other free translation services like Bing and Sogou, a user might want the option to chose between these services. Allow this in advanced config?
- Since wtr-lab seems to do the translation on demand, if the chapter is not already translated  WebToEpub  would still fail.

### DIFF
--- a/plugin/js/parsers/WtrlabParser.js
+++ b/plugin/js/parsers/WtrlabParser.js
@@ -10,7 +10,7 @@ class WtrlabParser extends Parser{
     async getChapterUrls(dom) {
         let items = [...dom.querySelectorAll("div.accordion a.chapter-item")];
         return items.map(a => ({
-            sourceUrl:  a.href + '?service=google',
+            sourceUrl:  a.href + "?service=google",
             title: this.formatTitle(a)
         }));
     }
--- a/plugin/js/parsers/WtrlabParser.js
+++ b/plugin/js/parsers/WtrlabParser.js
@@ -10,7 +10,7 @@ class WtrlabParser extends Parser{
     async getChapterUrls(dom) {
         let items = [...dom.querySelectorAll("div.accordion a.chapter-item")];
         return items.map(a => ({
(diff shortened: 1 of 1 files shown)


dteviot:
@vigneshsarma 

I'm not sure asking for a translation for every call is the right approach.  It's seems abusive to the site, as Google translation can be chargeable.

I think it would be better to make the call without asking for a translation, look at the response to see if a translation is needed, and then ask for a Google translation.  e.g. Something like

```javascript
    async fetchChapter(url) {
        let dom =  (await HttpClient.fetchHtml(url)).responseXML;
        return this.needsGoogleTranslate(dom)
            ? (await HttpClient.wrapFetch(url + "?service=google" )).responseXML
            : dom
    }
```

gamebeaker:
@dteviot the server caches the response from google. If the response isn't cached WebToEpub fails. https://github.com/dteviot/WebToEpub/issues/1430#issuecomment-2298645876

dteviot:
@gamebeaker 

I was assuming that adding "?service=google" to the call results in the server calling Google and caching the response.  If I'm wrong, then the originally proposed fix won't work.

@vigneshsarma Also, the checks are failing because you've used single quotes, you need to use double quotes.

vigneshsarma:
@dteviot Adding `"?service=google"` to the call only sets the preferred translation service. If the chapter has already been translated using google, that would be returned, otherwise would trigger a request to translate. At least that's how I understand the behavior of the site. 

If a the chapter needs to be translated webtoepub would stop at the previous chapter.

Current behavior of the site that broke WebToEpub for me is that it(wrt-lab) started defaulting to AI translation which requires users to login. This pr specifically requests google translation thus avoid getting the following message and  WebToEpub failing. 
> AI translation only available for registered users! You can sign up for free. Or you can read with other translation services.

Issue #1519, opened 2024-09-27
Help in setting getInformationEpubItemChildNodes()


gamebeaker:
In a few parser i had the problem that the description of the book isn't in the dom but easily accessible.
Example: 
88xiaoshuo.net
URL of dom: https://m.88xiaoshuo.net/Partlist/98781/
URL of info: https://m.88xiaoshuo.net/book/98781/
genesistudio.com
In the api request to get all Chapter is a description for the book with author etc. in the json response.

How can i get the info for example from an api request?
I think a possible solution would be to make 
getInformationEpubItemChildNodes()
extractDescription()
async
to use
let data = (await HttpClient.fetchJson(dom.baseURI + "/__data.json")).json;
and extract the description.

dteviot:
@gamebeaker 
The downside of that is you'd need to update all existing implementations with an async.  It's just a search/replace exercise, but numerous hits.  Tedious, but not difficult.

The other obvious option (to me) is to create async versions of the two functions.  And only provide an implementation if required.
Which is how getInformationEpubItemChildNodes() is used.  E.g.

https://github.com/dteviot/WebToEpub/blob/8bd11887c8a4d913714aa8799cf89ca553f1d441/plugin/js/Parser.js#L326-L328

Less work now, but complicates things going forward.  So, I agree with making them async.

dteviot:
Note, most implementations of getInformationEpubItemChildNodes() just get the summary, so they're duplicating extractDescription().  And, in fact, the base implementation of extractDescription is now to delegate to getInformationEpubItemChildNodes().

gamebeaker:
@dteviot i just had the idea to switch the populateMetaInfo() with onLoadFirstPage() (switch second try with third try)
https://github.com/dteviot/WebToEpub/blob/afaadc55f0ad31790bba6f86bf8939f632a86cfe/plugin/js/main.js#L39-L61
If it is switched the Metadata can be obtained with async (await api call) during getChapterUrls() and saved in the parser as local variable and getEpubMetaInfo() just returns these local variables.

gamebeaker:
On second thought this is maybe not that good of an idea. Because Metadata retrieval gets mixed with the chapter retrieval.

I have created a new funtion called async loadEpubMetaInfo() and think this is better suited for this here is the commit: https://github.com/dteviot/WebToEpub/commit/7df5e98daf42471473ec18a13ea104bf192ffd23

@dteviot any thoughts?

dteviot:
[...]

PR #1762, opened 2025-04-05
Add async loadEpubMetaInfo()

gamebeaker:
reference: #1519
@dteviot any thoughts?


gamebeaker:
reference: #1761
Here is an example how it can be used to get the metadata info from an api.

PR #1779, opened 2025-04-08
potential fix #1775

### LINKED ISSUE #1775: Would it be possible to change how the chapters are numbered?
```
<a href="https://">1: Chapter 1: Kids, this isn't funny.</a>
<a href="https://">2: Chapter 2: The Announcement Letter</a>
<a href="https://">3: Chapter 3: Showtime</a>
```
This is an example of how titles are written(I removed a big part of the URL just to shorten it up).

The problem that I'm having is that when I use TTS to listen to the novel, It gets a bit annoying hearing it recite the number twice. Sometimes it's not even correct; Let's say there are extra chapters or non-canon omakes', so now it's no longer accurate. Leading to examples as “48 Chapter 44: Home”(Read as: Forty-eight Chapter Forty-four, Home) and just making it a bit confusing to listen too.

My fix for this issue is to clean up the file once I've downloaded it using regex. It removes the 1:, 2:, 3:… etc. from the title. It goes from:
```
1: Chapter 1: Kids, this isn't funny.
2: Chapter 2: The Announcement Letter
3: Chapter 3: Showtime
```
to this:
```
Chapter 1: Kids, this isn't funny.
Chapter 2: The Announcement Letter
Chapter 3: Showtime
```

I'm just wondering if this could be changed.

Happens on WebNovel.

gamebeaker:
@Kiradien in reference to #1778 and #1775

My idea is a little different from yours.
I added this Checkbox 
image
And my Idea is that the Chapter title in the content gets always replaced with the title from the ToC in case there is [placeholder] your fix from #1778 will be used.

I am going to be honest and say that i didn't really understand your regex i tried it here https://regex101.com/ and couldn't get it to work with problematic title (example: `Chapter 1: Chapter: 1`)

To be honest i am surprised that this isn't already the norm. Now that i think about it this would make `findChapterTitle()` for all parser irrelevant if always the ToC name would be used.

### DIFF
--- a/plugin/_locales/en/messages.json
+++ b/plugin/_locales/en/messages.json
@@ -379,6 +379,10 @@
 		"message": "Remove Author Notes",
 		"description": "Label for checkbox to remove Author Notes from chapters"
 	},
+	"__MSG_label_Remove_Chapter_Number__": {
+		"message": "Remove Chapter Numbers",
+		"description": "Label for checkbox to remove Chapter Numbers from chapter titles"
+	},
 	"__MSG_label_Remove_Original__": {
 		"message": "Remove Original/Raw Text",
 		"description": "Label for checkbox to remove Original text from chapters"
--- a/plugin/js/UserPreferences.js
+++ b/plugin/js/UserPreferences.js
@@ -94,6 +94,7 @@ class UserPreferences {
         this.addPreference("chaptersPageInChapterList", "chaptersPageInChapterListCheckbox", false);
         this.addPreference("autoSelectBTSeriesPage", "autoParserSelectIncludesBTSeriesPageCheckbox", false);
         this.addPreference("removeAuthorNotes", "removeAuthorNotesCheckbox", false);
+        this.addPreference("removeChapterNumber", "removeChapterNumberCheckbox", false);
         this.addPreference("removeOriginal", "removeOriginalCheckbox", true);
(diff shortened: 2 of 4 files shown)


Kiradien:
> am going to be honest and say that i didn't really understand your regex i tried it here https://regex101.com/ and couldn't get it to work with problematic title (example: `Chapter 1: Chapter: 1`)

I'm gonna clearly say this was my fuckup - I had it working for that, then I handled the more complex case of "Chapter 01: Chapter 01: The beginning" and broke it for that use case. I've pushed a better version of the regex to my branch, it's not actually that different
`(\\w+[\\s\\-]+\\d+):\\s*\\1:?(.*)"` instead of `(\\w+[\\s\\-]+\\d+:)\\s*\\1(.*)`

Sites like regex101 don't require regexes to be escaped, so just need to be added without so many double slashes `(\w+[\s\-]+\d+):\s*\1:?(.*)"`

Kiradien:
Alright, aside from the image saying "Remove Remove Chapter Numbers" for some reason, and the issue with my regex, this solution looks good. removeChapterNumberRow.hidden removes my main concern for extra options. 

Do you want to update the regex to `(\\w+[\\s\\-]+\\d+):\\s*\\1:?(.*)` & I'll close my existing pull request, or merge now and I'll fix that in my PR?

Issue #1840, opened 2025-04-27
Fix security warnings raised by Mozilla


dteviot:
Unsafe assignment to innerHTML
Warning: Due to both security and performance concerns, this may not be set using dynamic values which have not been adequately sanitized. This can lead to security issues or fairly serious performance degradation.
* js/Library.js line 400 column 17
* js/Library.js line 489 column 17
* js/Library.js line 495 column 21
* js/parsers/QidianParser.js line 101 column 13

PR #1842, opened 2025-04-27
fix firefox warning

gamebeaker:
reference: #1840
web-ext lint results in:
image


dteviot:
@gamebeaker 

Thank you.

Issue #1847, opened 2025-04-28
wtr-lab is not working anymore


Tyler-Burden:
<img width="1141" alt="Image" src="https://github.com/user-attachments/assets/9779b157-35f3-40c2-a907-2f14bf1f46f7" />

wtr-lab has stopped working for me, same issue is faced by a couple more people today on Reddit too. Tried expanding the table of contents too, it collects the chapter's links as expected (without the chapter numbers tho, just the titles). Once you click "Pack Epub", you get the Download confirmation dialog box at the first chapter, the epub downloaded contains no chapters. It creates this error. Also, this is the same when I am signed-in.

gamebeaker:
@Tyler-Burden i am looking into it. Care to share the reddit link?
Notes for me:
looks like the magickey is the problem
https://github.com/dteviot/WebToEpub/blob/1da8d724f1ebd3ac11af0568760574977c77462a/plugin/js/parsers/WtrlabParser.js#L65-L69

gamebeaker:
@Tyler-Burden i created a hot fix that will break again it isn't a real fix.
Test versions for Firefox and Chrome have been uploaded to https://github.com/dteviot/WebToEpub/releases/tag/developer-build. Pick the one suitable for you, follow the "How to install from Source" instructions at https://github.com/dteviot/WebToEpub/tree/ExperimentalTabMode#user-content-installation and let me know how it goes.

gamebeaker:
@Tyler-Burden i created a proper fix and you have now the option to add chapter number or not(after a change reload webtoepub):

![Image](https://github.com/user-attachments/assets/723e8484-93bf-4e8f-9e35-44745713ca53)
Test versions for Firefox and Chrome have been uploaded to https://github.com/dteviot/WebToEpub/releases/tag/developer-build. Pick the one suitable for you, follow the "How to install from Source" instructions at https://github.com/dteviot/WebToEpub/tree/ExperimentalTabMode#user-content-installation and let me know how it goes.

Tyler-Burden:
That worked! You've been a blessing. Thanks :))

Phaethonix:
can service=google be optional quality kinda ass tbh 

> [@Tyler-Burden](https://github.com/Tyler-Burden) i created a proper fix and you have now the option to add chapter number or not(after a change reload webtoepub):
> 
[...]

Issue #1770, opened 2025-04-07
Please add site https://estar.jp/


lemonskydesu:
Please note, I'm basically the only developer working on WebToEpub, and I'm not paid for doing this.  (WebToEpub is completely free, and generates no money.) By asking to add a site, you're asking me to give you some of my limited free time. So, I think it's not unreasonable for me to ask you to do as much as you can to help me.

### Did you try using the Default Parser for the site?  If not, why not?
Instructions for using the default parser can be found at https://github.com/dteviot/WebToEpub/wiki/FAQ#how-to-convert-a-new-site-using-the-default-parser

### What settings did you use?  What didn't work?
* URL of first chapter
* CSS selector for element holding content to put into EPUB
* CSS selector for element holding Title of Chapter
* CSS selector for element(s) to remove

### If the Default Parser did not work, if you have developer skills, did you try writing a new parser?
Instructions https://github.com/dteviot/WebToEpub/wiki/FAQ#how-to-write-a-new-parser

gamebeaker:
The problem is that in the fetch post request the origin header is set to the addon. If the origin header is not set or the website it works here is a curl command that works to get a ToC page
```
curl --header "accept-encoding: deflate" --header "content-type: application/json;charset=UTF-8" --header "x-from: https://estar.jp/novels/26264581/episodes?page=2" --header "User-agent: Mozilla/5.0 (Windows NT 10.0; Win64; x64; rv:137.0) Gecko/20100101 Firefox/137.0" --data "{\"query\":\"pages/novels/workId/episodes\",\"data\":{\"workId\":\"26264581\",\"first\":30,\"page\":3}}" https://estar.jp/api/graphql -o test.json
```

dteviot:
@lemonskydesu 

Updated version (1.0.6.0) has been submitted to Firefox and Chrome stores.
Firefox version is available now.
Chrome might be available in a few hours (typical) to 21 days.

PR #1857, opened 2025-04-29
Add declarativeNetRequest header handling

gamebeaker:
reference: #1770
@dteviot please confirm merge it could be that this triggers a manual review as a new permission is needed for the change
`declarativeNetRequest`

What is the problem?
Some sites check the headers they receive and only send content if the headers are correct.
estar.jp for example checks if the "origin" header is correct if it exists.
With this change i created a rule that removes the "origin" header from all requests.

This could also be used for other sites such as wtr-lab.com #1847 as they changed to a "POST" api which checks these headers.
I created it in HttpClient.js to try and abstract the security concerns.
declarativeNetRequest rules are pretty powerful and can compromise the security on other sites if they aren't bound to WebToEpub only.

### DIFF
--- a/plugin/js/HttpClient.js
+++ b/plugin/js/HttpClient.js
@@ -215,6 +215,37 @@ class HttpClient {
         }
     }
 
+    static async getActiveTab() {
+        return new Promise(function (resolve, reject) {
+            chrome.tabs.query({ currentWindow: true, active: true }, function (tabs) {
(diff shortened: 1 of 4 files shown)


dteviot:
@gamebeaker Should be fine.  Firefox is already using the more dangerous webRequest.
And I doubt Google will do a manual check.  I'll just have to provide the paper work for why we're making the call.
Which is: "Remove 'origin' headers for requests to site Estar

Issue #1862, opened 2025-04-30
TypeError: Cannot read properties of null (reading 'title')


narifin01:
**Describe the bug**
TypeError: Cannot read properties of null (reading 'title')
**To Reproduce**
1. Go to '(https://my-novel.online/series/making-honkai-impact-3rd-in-a-world-where-kallen-exists/)'
2. Click on 'webtoepub'
3. See error

**Screenshots**

![Image](https://github.com/user-attachments/assets/ba99c77d-5dbf-4aee-88a4-c2e42dfc160e)

**Desktop (please complete the following information):**
 - OS: [Window]
 - Browser [Microsoft Edge]
 - Version [135.0.3179.98]

gamebeaker:
@narifin01 you are to fast in the download. I will make a fix where it is slowed down.

gamebeaker:
@narifin01 i just found another error will take longer to fix

gamebeaker:
@narifin01 
Test versions for Firefox and Chrome have been uploaded to https://github.com/dteviot/WebToEpub/releases/tag/developer-build. Pick the one suitable for you, follow the "How to install from Source" instructions at https://github.com/dteviot/WebToEpub/tree/ExperimentalTabMode#user-content-installation and let me know how it goes.

narifin01:
thanks for the fix.

dteviot:
@narifin01 

Updated version (1.0.6.0) has been submitted to Firefox and Chrome stores.
Firefox version is available now.
Chrome might be available in a few hours (typical) to 21 days.

PR #1867, opened 2025-05-01
update MyNovelOnlineParser

gamebeaker:
reference: #1862
Website tries to slow down the user but it is wrong implement as they still send the content. Out of respect i slowed down for 3 seconds and clampSimultanousFetchSize to 1.


gamebeaker:
image

Issue #1468, opened 2024-09-07
Feature request: change clampSimultanousFetchSize()


gamebeaker:
Change `clampSimultanousFetchSize()` so that it is the same as `this.minimumThrottle = 3000;`
I think a variable is more elegant than a function and at the moment the user is unable to overwrite `clampSimultanousFetchSize()`.
Todo:
- [ ] Change `clampSimultanousFetchSize()`  to a variable
- [ ] If "Override Minimum Throttle Value (where applicable)" is checked allow overwrite 
- [ ] Rename "Override Minimum Throttle Value (where applicable)"

dteviot:
@gamebeaker 

User is not supposed to be able to override clampSimultanousFetchSize().  It's applied on a per parser (i.e. Site) basis, for sites that don't like rates of more than 1 page at a time.

gamebeaker:
@dteviot i would argue that this.minimumThrottle is the same as in that if it is defined in a parser it is normally required or you get rate limited. So user should not change minimumThrottle but are still able todo it if they want. 
I would be ok with either (the user is able/ unable to overwrite both values). I just want it to be consistent as both option influence fetches/ second.

dteviot:
@gamebeaker 
I agree the rate limiting is something a mess.
Perhaps we should take a step back, decide what the requirements/goals need to be, and proceed from there.

dteviot:
1. User should be able to set rate on a per-site/parser basis.
2. Default rate is 1 page at a time, 3 seconds per fetch.
3. Should be as easy for users to use as possible.
4. Some sites mandate 1 page at a time.  (e.g. Royal Road's TOS.)  For these sites, users can not override pages at a time.

gamebeaker:
@dteviot and we don't violate the TOS...
![image](https://github.com/user-attachments/assets/a4e34bd6-7eb3-4543-b1f8-d16837e758d2)

gamebeaker:
idea:
```
class Parser {    
    constructor(imageCollector) {
        this.minimumThrottle = 0;
        this.defaultThrottle = 3;
        this.maximumPagesToFetchSimultaneously = 8;
        this.defaultPagesToFetchSimultaneously = 1;
```
and then control logic like in `getRateLimit()` for maximumPagesToFetchSimultaneously in `getmaximumPagesToFetchSimultaneously()` where the user value gets ignored if it is less than minimumThrottle or more than maximumPagesToFetchSimultaneously

[...]

PR #1870, opened 2025-05-01
help needed Add SangtacvietParser

gamebeaker:
reference: #1477

The site has an custom rate limit.
As i access the "api" directly i get an http 200 response but the body says solve the captcha.

How can i trigger the 429 error response from the parser? (line 71)

I think that this problem will be more widespread soon because at the moment most api's have no rate limit i don't think that this will last.https://github.com/dteviot/WebToEpub/issues/1468#issuecomment-2843609231

### DIFF
--- a/plugin/js/parsers/SangtacvietParser.js
+++ b/plugin/js/parsers/SangtacvietParser.js
@@ -1,0 +1,100 @@
+"use strict";
+
+parserFactory.register("sangtacviet.com", () => new SangtacvietParser());
+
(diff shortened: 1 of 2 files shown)


dteviot:
@gamebeaker 

I think you'd have to extend the HttpClient class.  Add member to take a "custom check and retry" and set that in the parser.

gamebeaker:
I opened issue #1874 for this problem.
The current solution is too much in case it is needed for more parser. Refactoring needed.

PR #1871, opened 2025-05-01
update speed logic

gamebeaker:
reference: #1468
this change sets the clampSimultanousFetchSize() for all to 1 and removes the option to increase it from the user.
It sets minimumThrottle=500 in the Parser constructor for all parsers.
The ms delay is now a input field instead of a selection.

I am not sure if removing the option for simultanous fetch is the right move. Thoughts @dteviot @Kiradien?

### DIFF
--- a/plugin/_locales/en/messages.json
+++ b/plugin/_locales/en/messages.json
@@ -324,8 +324,8 @@
 		"description": "Show more Metadata options to user"
 	},
 	"__MSG_label_Manual_Delay_Per_Chapter__": {
-		"message": "Manual Throttle|Delay per chapter",
-		"description": "Manual Throttle|Delay per chapter"
+		"message": "Delay per chapter in ms",
+		"description": "Delay per chapter in ms"
 	},
 	"__MSG_label_Manually_Select_Parser__": {
 		"message": "Manually Select Parser:",
--- a/plugin/css/default.css
+++ b/plugin/css/default.css
@@ -103,7 +103,7 @@ button:disabled {
 }
 
 
-input[type=text], input[type=url], textarea, select {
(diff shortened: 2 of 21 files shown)


dteviot:
@gamebeaker 

You set this.maxSimultanousFetchSize = 1; in the constructor for the Parser class.
As you've removed the UI to change the value, I don't think you need to set it in any of the derived classes.
Or am I missing something?

> removing the option for simultanous fetch is the right move.

Given the number of complaints due to speed causing problems on sites, yes, removing it is probably the right thing to do.
Note, we will now get complaints about how slow it is.

Kiradien:
Hmmm, I do agree that removing simultaneous fetch is a good move, but I'm not sure about the 500ms minimum default for all parsers... It's a fairly safe value, but will slow down a lot of sites that have a much lower threshold - though by the same measure, those can be determined & set case by case.

That said, absolutely keep `this.addPreference("manualDelayPerChapter", "manualDelayPerChapterTag", "0");` as the default; the 'manual delay' value there becomes a hard minimum for ALL sites, which would block all sites that have been overridden to a lower number - such as qidian and spacebattles/Xenforo

gamebeaker:
> Or am I missing something?

I wanted to leave them there in case there is user backlash and they want the option back. (That would make it easier to revert back)

> That said, absolutely keep this.addPreference("manualDelayPerChapter", "manualDelayPerChapterTag", "0"); as the default; the 'manual delay' value there becomes a hard minimum for ALL sites

thanks for the catch

gamebeaker:
> I wanted to leave them there in case there is user backlash and they want the option back. (That would make it easier to revert back)

should there really so a huge backlash it is easy to find out. (just look which sites got changed in this commit)

Issue #1874, opened 2025-05-02
Add custom 403 retry error on http 200


gamebeaker:
Some websites send HTTP 200 with the wrong/ empty content.
If an API rate limits you it send a json file with empty/ wrong content.

As the original url is a api url there needs to be an option to convert the resturl to the "normal" url that gets opened if the "open url" in 403 dialog is clicked.

(clean up WtrlabParser/ SangtacvietParser)

PR #1876, opened 2025-05-02
Add custom retry error

gamebeaker:
reference: #1874
This adds two functions to the Parsers:
isCustomError()
take a look at the content and decide if it is really HTTP 200 (true=error; false=success)
setCustomErrorResponse() 
this function manipulates the response so that it is handled like an HTTP error
To make use of this the parser needs to be added to the "fetchOptions" or "wrapOptions".

I changed two parser that make use of this:
SangtacvietParser
It returns HTTP 200 but the json body doesn't contain the content.
WtrlabParser
If it is the first time someone requests a page the server returns HTTP 200 with missing content while the server is translating the content with the custom HTTP 999 error it is possible to set parser specific retry delay times.

### DIFF
--- a/plugin/js/HttpClient.js
+++ b/plugin/js/HttpClient.js
@@ -29,7 +29,7 @@ class FetchErrorHandler {
     }
 
     onResponseError(url, wrapOptions, response) {
-        let failError = new Error(this.makeFailMessage(url, response.status));
+        let failError = new Error(this.makeFailMessage(response.url, response.status));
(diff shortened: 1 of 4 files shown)


dteviot:
@gamebeaker 
Concept seems sound.

Note:  
Can simplifythis code in WtrlabParser.js
```javascript
        if (response.json.data?.data?.body?false:true) {
            return true;
        }
        return false;
```

to
```javascript
        return !(response.json.data?.data?.body)
```

gamebeaker:
I didn't realize that i closed this PR.
I am merging this.

PR #1893, opened 2025-05-07
Add parser for Fictioneer WordPress theme

crybx:
I've been using this parser in a private branch since July last year and recent changes introduced a LilyonthevalleyParser that mine actually covers, in addition to other sites that use this WordPress theme. I figured I'd finally offer my version as a PR so it covers more sites.

I also updated FictioneerParser a bit to make sure nothing from the LilyonthevalleyParser would be omitted (like not selecting scheduled chapters by default).


gamebeaker:
@crybx thank you for the PR.
Can please fix the eslint errors? 
https://github.com/dteviot/WebToEpub/actions/runs/14890620181/job/41824135855?pr=1893#step:5:1587

plugin/js/parsers/FictioneerParser.js
```diff
@@ -0,0 +1,128 @@
...
+                element.removeAttribute('data-paragraph-id');
+            }
+            // remove style attribute if style="font-weight: 400;" - it's just noise
+            if (element.hasAttribute('style') && element.getAttribute('style') === 'font-weight: 400;') {
+                element.removeAttribute('style');
+            }
+        });
+    }
+
+    removeUnwantedElementsFromContentElement(element) {
```
dteviot, line 115: Is there a reason you're not just calling 
```javascript
base.removeUnwantedElementsFromContentElement(element)
```
After your removeElements() line?

dteviot:
@crybx 

And to echo gamebeaker, thank you very much for your contribution.

plugin/js/parsers/FictioneerParser.js
crybx: Ah, no. I think I originally just copied the whole base method and then changed it a little. Updated.

dteviot:
@crybx 
I forgot to ask you to add yourself to the credits.
Never mind, I'll take care of it.

And thank you again.

PR #1907, opened 2025-05-10
Misc fixes

crybx:
Any change that isn't in TapasParser, MadaraParser, or .gitignore was to fix a failing unit test.

### DIFF
--- a/plugin/js/ChapterUrlsUI.js
+++ b/plugin/js/ChapterUrlsUI.js
@@ -7,10 +7,19 @@ class ChapterUrlsUI {
         ChapterUrlsUI.getPleaseWaitMessageRow().hidden = false;
         if (this.parser)
         {
-            document.getElementById("spanParserName").textContent = this.parser.constructor.name;
-            document.getElementById("spanDelayMs").textContent = `${this.parser.getRateLimit()} ms`;
+            let nameElement = document.getElementById("spanParserName");
+            if (nameElement) nameElement.textContent = this.parser.constructor.name;
+
+            let delayMsElement = document.getElementById("spanDelayMs");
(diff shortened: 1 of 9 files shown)


gamebeaker:
I can't find the successful https://github.com/dteviot/WebToEpub/actions/workflows/node.js.yml check for this PR.
image
Should this happen again maybe i am going to investigate why this isn't linked with this PR.

crybx:
@gamebeaker I had a failure at first, made a fix and force pushed the update. Here was the original failure (that I just HAPPENED to take a screenshot of)

failure

Last time I force push with an active PR on the branch it re-ran the build steps fine, I don't know how to kick it off again now 😭

crybx:
Well now I have an extra commit 'Merge branch 'ExperimentalTabMode' into fixes' but it ran

plugin/js/ChapterUrlsUI.js
gamebeaker, line 1: Is there a reason for the rewrite?
I don't see the reason for your change. 
As all ElementsbyId are static in popup.html the if conditions will always be true.

crybx: Yes, the reason is to pass the unit tests where those elements do not exist.
image

These tests could be made to pass by injecting an element with those ids into the document for the duration of the test but... the test is just trying to test the chaptersToHTML method, not the extra elements and happens to need the constructor. Maybe some future feature will want an instance of the ChapterUrlsUI and not use the popup.html. Maybe chaptersToHTML could be static, but then makeLink would need to be too and anything that called either would need to be updated. I thought about a lot of options, but it seemed most harmless to just make sure the elements exist.

Issue #1888, opened 2025-05-05
Make WebToEpub more conform with w3c


gamebeaker:
There is this program [epub-checker](https://pagina.gmbh/startseite/leistungen/publishing-softwareloesungen/epub-checker/)
If i throw in an epub created with WebToEpub i get a bunch of errors and warnings.
(Chapter 1 & 2 from https://www.royalroad.com/fiction/112549/a-modern-miners-cultivation-manual)
<details>
  <summary>Image</summary>

![Image](https://github.com/user-attachments/assets/2bb14cd8-bba1-40b2-bc7a-2689df44347f)

</details>

I know that some errors/ warnings will be unfixable but it could be made an effort.
.webp is now in the spec: https://www.w3.org/TR/epub-33/#sec-core-media-types (epub 3.3)

@dteviot should we keep supporting EPUB 2?

dteviot:
@gamebeaker 
>should we keep supporting EPUB 2

Yes.  Mostly because WebToEpub doesn't always create a valid EPUB 3 file.  I started down that path, but I found it required a lot more work to do it properly, in all cases, for little gain.

I believe https://www.w3.org/publishing/epubcheck/ is the official tool for checking compatibility.  It's something of a pain to use.

Looking at your screenshot, the errors with the information page are mostly due to that code doing a raw scoop of the HTML and stuffing it into a page.  (Specifically, the &lt;input&gt; control on the page.
I think I've also seen problems with image files where format doesn't match the filename, or the advertised mime.

We could add better support for webp, (Like, don't give warning if EPUB 3.3 selected) but I wonder how many older readers out there will support it?

I've spent a lot of time in the past trying to get to 100% EPUB compliance, but came to conclusion, what we have is good enough, and more effort is in the area of greatly diminishing returns.

gamebeaker:
@dteviot it uses https://www.w3.org/publishing/epubcheck/ / https://github.com/w3c/epubcheck as library and just provides an gui.

> I've spent a lot of time in the past trying to get to 100% EPUB compliance, but came to conclusion, what we have is good enough, and more effort is in the area of greatly diminishing returns.

Ok i will close this issue.

PR #1909, opened 2025-05-11
Various fixes plus add novelingua

crybx:
Sorry for the PR spam. This one has the changes from my other un-merged PR plus more. I just have a lot of fixes and changes in my private repo and I've been trying to clean up what might be good for everyone, including parsers for sites that are being requested lately (like Novelingua).

My other PR could be closed or merged first and I can also remove anything that's not wanted.

### DIFF
--- a/plugin/js/EpubPacker.js
+++ b/plugin/js/EpubPacker.js
@@ -112,7 +112,8 @@ class EpubPacker {
             this.addMetaProperty(metadata, translator, "role", "translator", "trl");
         }
 
-        let identifier = that.createAndAppendChildNS(metadata, dc_ns, "dc:identifier", that.metaInfo.uuid);
+        let idText = (this.version === EpubPacker.EPUB_VERSION_3 ? "uri:" : "") + that.metaInfo.uuid;
+        let identifier = that.createAndAppendChildNS(metadata, dc_ns, "dc:identifier", idText);
         identifier.setAttributeNS(null, "id", "BookId");
         if (this.version === EpubPacker.EPUB_VERSION_2) {
             identifier.setAttributeNS(opf_ns, "opf:scheme", "URI");
--- a/plugin/js/Util.js
+++ b/plugin/js/Util.js
@@ -925,7 +925,43 @@ var util = (function () {
                 }
             );
         });
-    }    
+    }
+
+    const replaceSemanticInlineStylesWithTags = function(element, removeLeftoverStyles = false) {
+        if (element.hasAttribute("style")) {
+            let styleText = element.getAttribute("style");
+
+            // Map of style patterns to their semantic HTML equivalents
+            const styleToTag = [
+                { regex: /font-style\s*:\s*(italic|oblique)\s*;/g, match: /italic|oblique/, tag: "i" },
+                { regex: /font-weight\s*:\s*(bold|[7-9]\d\d)\s*;/g, match: /bold|[7-9]\d\d/, tag: "b" },
+                { regex: /text-decoration\s*:\s*underline\s*;/g, match: /underline/, tag: "u" },
+                { regex: /text-decoration\s*:\s*line-through\s*;/g, match: /line-through/, tag: "s" }
+            ];
+
+            // Apply semantic tags and remove corresponding styles
+            for (const style of styleToTag) {
+                if (style.match.test(styleText)) {
+                    const wrapper = document.createElement(style.tag);
+                    while (element.firstChild) {
+                        wrapper.appendChild(element.firstChild);
+                    }
+                    element.appendChild(wrapper);
+                    styleText = styleText.replace(style.regex, "");
+                }
+            }
+
+            // Remove non-semantic font-weight
+            styleText = styleText.replace(/font-weight\s*:\s*(normal|[1-4]\d\d)\s*;/g, "");
+            styleText = styleText.trim();
+
+            if (styleText && (!removeLeftoverStyles || /italic|bold|font-weight|underline|line-through/.test(styleText))) {
+                element.setAttribute("style", styleText);
+            } else {
+                element.removeAttribute("style");
+            }
+        }
+    }
 
     return {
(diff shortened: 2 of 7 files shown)


plugin/js/Util.js
gamebeaker, line 1: This looks to me like it will violate the google addons store policy.

gamebeaker: If you have a parser for such a site you could add the decipher function in this projected: https://github.com/dteviot/EpubEditor

crybx: I wonder if there's a link to the policy details? A simple cipher is a very basic message coding technique that kids have solved in magazines before the internet existed. Having a general method for handling them that provides no cipher keys and is implemented on no websites that use this technique for "DRM" doesn't seem like it should be a violation of anything. The method could, if desired, be used to embed newly encoded messages in an epub for fun, or be used on websites that are encoding text for fun and not "DRM".

crybx: I removed the function, I don't want to hold anything up, but I'm still curious about the policy.

gamebeaker: https://developer.chrome.com/docs/webstore/program-policies
and there specific 
https://developer.chrome.com/docs/webstore/program-policies/malicious-and-prohibited
There could be made arguments that WebToEpub already violates these policy points.
We see it like this: 
The "normal" requesting/ parsing is ok as the user could do it manual. And there are a lot of video downloader that help to download not encrypted video. 
As soon as something is encrypted (or uses custom fonts to obfuscate the content like https://requiemtls.com/) we don't decrypt it in WebToEpub as this is an obvious attempt at preventing the copy/ download of the content.
Video downloader extension are also unable (unwilling) to download encrypted content.

plugin/js/EpubPacker.js
gamebeaker, line 1: Do you have a source for this? I am unable to find it in the spec: https://www.w3.org/TR/epub-33/#dfn-dc-identifier

plugin/js/Util.js
gamebeaker, line 1: @dteviot I think replaceSemanticInlineStylesWithTags can be a used for all sites in convertRawDomToContent()

plugin/js/EpubPacker.js
crybx: I don't think it's about the format specification. The spec just says they _may_ have multiple identifiers. My source is that without this fix, Calibre doesn't recognize the url as a valid identifier at all. Its list of identifiers is blank, when normally it would have a clickable url that shows up:

calibre-interface-epub3-no-url

Even though the url is there in opf, it's not in a format recognizable because it's missing any kind of prefix or scheme:

calibre-interface-epub3-no-url-opf

Now if I copy the correct url and use Calibre interface to click this button that will format the identifier for me, the clickable link appears

calibre-interface-epub3-added-url

And this is what's added to the opf:

calibre-interface-epub3-added-url-opf

NOW, with my fix in this PR, the url in the interface is populated with no manually editing by me:

epub3-no-edits-with-fix

And appears in the opf with a prefix and scheme:

epub3-no-edits-with-fix-opf

crybx: Also, in that last screenshot, Calibre saw `<dc:identifier id="BookId" opf:scheme="URI">https://tapas.io/series/TBATE</dc:identifier>`
and added `<dc:identifier>uri:https://tapas.io/series/TBATE</dc:identifier>` upon import because it could recognize the scheme. `<dc:identifier>uri:https://tapas.io/series/TBATE</dc:identifier>` isn't in the epub3 created by WebToEpub, in case that's confusing.

gamebeaker: I think it is a problem with calibre as if i run an epub checker that uses the w3c library it flags this as error.
image
reference: #1888

I know there are a lot of other errors. A lot of them are parsed content from the Internet so not really fixable.

crybx: Calibre even generates a uuid cause it doesn't think the one identifier counts as the required one.

I don't know where to try that checker in your screenshot, I can't find it. But does it complain about this? 
`<dc:identifier id="BookId">url:https://tapas.io/series/i-shall-master-this-family-novel</dc:identifier>`

Calibre would also work if that was used and I could fix this PR to do that (just adds the url: prefix before the https)

dteviot: @crybx @gamebeaker 

Looking at the Epub 3 spec https://idpf.org/epub/30/spec/epub30-publications.html#sec-opf-dcidentifier the "opf:scheme" attribute is not permitted.  However, a "uri" prefix on the value is probably OK.

For Epub 2 spec, https://idpf.org/epub/20/spec/OPF_2.0_final_spec.html, section 2.2.10, "opf:scheme" attribute is permitted.  It's not 100% clear if a uri prefex would be OK, but I think it almost certainly is.

So, I'm NOT OK with adding opf:scheme for Epub 3, but a "uri" prefix on the value is OK.

Hopefully that's clear

plugin/js/Util.js
dteviot: @gamebeaker 
Agreed

plugin/js/EpubPacker.js
gamebeaker: My proposal is to change line 115 To
````
let identifier = that.createAndAppendChildNS(metadata, dc_ns, "dc:identifier", ((this.version == EpubPacker.EPUB_VERSION_3)?"uri:":"")+that.metaInfo.uuid);
````
and revert the other changes.
I tested it and the result is an valid epub 2 and 3
Calibre recognizes the EPUB3 uri

crybx: @gamebeaker I pushed a change that uses your proposal, just breaks it into two lines for length.

Calibre is NOT okay with both `opf:scheme="URI"` and uri: prefix in EPUB 2. It ends up generating a `uri:uri:actual_url` entry and it doesn't even make a clickable link show up. So only appending for EPUB 3 is good.

dteviot:
@crybx 
Thank you for your hard work.

PR #1917, opened 2025-05-12
Fix NovelinguaParser

crybx:
Fixes issue with some chapters being empty.


crybx:
SIGH, don't merge this yet, I found another empty chapter case.

PR #1918, opened 2025-05-12
Fewer empty things in epub

crybx:
I'm know I'm an outlier who edits every epub I touch, but it's nice not to see (and clean up) a ton of completely pointless spans. I also used to see style="" or class="" sometimes.

Plus some other clean up, a couple new util functions that can reduce lines of code elsewhere, and improvements to the TapasParser as I've been using it.

### DIFF
--- a/plugin/js/Util.js
+++ b/plugin/js/Util.js
@@ -336,10 +336,9 @@ var util = (function () {
     }
 
     var moveChildElements = function(from, to) {
-        while (from.hasChildNodes()) {
-            let node = from.childNodes[0];
-            to.appendChild(node);
-        };
+        while (from.firstChild) {
+            to.appendChild(from.firstChild);
+        }
     }
 
     var copyAttributes = function(from, to) {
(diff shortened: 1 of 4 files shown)


plugin/js/Util.js
gamebeaker, line 1: Why are you so focused on Attributes? 
For example removeSpansWithNoAttributes() what is the reason for this?
Wouldn't it be better to check for empty textContent and than remove them even if they have Attributes? (I am not sure which site it was but it random parts of the content in spans without Attributes that why i ask)

crybx: removeSpansWithNoAttributes is not meant to remove stuff with no textContent, it's mean to unwrap exactly all the random parts of content in spans with  no attributes. So this:

```
<p><span><span>The three of them stood motionless, looking down in uneasy silence. What happened <span></span><span><span><span>to<span></span></span> Shifty didn't come as a shock, but it was still</span></span> a hard thing to digest. An ominous feeling settled in their hearts — seeing the broken body of their companion, it<span> was</span> too easy to imagine one of them sharing the same fate.</span></span></p>
```
becomes
```
<p>The three of them stood motionless, looking down in uneasy silence. What happened to Shifty didn't come as a shock, but it was still a hard thing to digest. An ominous feeling settled in their hearts — seeing the broken body of their companion, it was too easy to imagine one of them sharing the same fate.</p>
```
It's basically span stripping. It only does this if they're in p or div tags. Also, with the method added in Parser, the method to remove empty divs comes right after (empty divs do no harm, why remove them? for the same reason to remove pointless spans, to have cleaner content)

So this: 
```
<div><span></span><div>
```
would become this
```
<div></div>
```
which would then be completely removed.

This is a real example from Tapas that I pull before that change:
```
<p><span>Imagine how dumbfounded I felt. This man, who never even visited his ailing daughter, proceeded to say that “</span><span><i>I looked like I had been doing especially well for the past few days</i></span><span>.”</span></p>

<p><span>It’s not as if I was expecting anything as caring as “</span><span><i>I heard you were unwell. How are you doing?</i></span><span>” or “</span><span><i>Look how gaunt your face has become</i></span><span>.”</span></p>
```
It drives me nuts because I edit every epub I have.

dteviot:
@crybx 
>empty divs do no harm, why remove them?

Because they introduce a break or space in the text.  And there's a couple of sites where this results in a layout of lines of text, separated by multiple blank lines.  i.e. Page is 67% empty space.  Quite distracting to read.

dteviot:
@gamebeaker 

FYI.  You have full rights to approve any pull request you're OK with.  You don't need my OK.

dteviot:
@crybx 

I can understand your desire to remove unnecessary things from the epub.
However, I always have concerns that it might run into some corner case where it results in a problem:
1. Something that we want is removed, or
2. The HTML is malformed, and the code crashes, or
3. An unexpected interaction, or
4. It just complicates reading the code.

So, my general philosophy with WebToEpub is to keep it as simple and minimal as possible, and grab everything, even if not necessary.  (Provided it doesn't mess up the reading experience.)

Any custom cleaning should be delegated to the EpubEditor, which is applied to the collected epubs.  
Yes, I know it's an extra step, but I'd rather require users who want it, do a little extra work, as opposed to risk things for everyone else.

That said. I accept these changes.

gamebeaker:
@dteviot 

> FYI. You have full rights to approve any pull request you're OK with. You don't need my OK.

I know. 
I am uncomfortable to merge changes that touch more than the Library or specific parsers without you having a look at it to know what changed.
That is in case something breaks and i am too slow. You could have an idea which change could have broken it. 
In this case for example if we get complains that a lot of page layouts are broken all of a sudden you know where to look at. (I don't hope it happens)

dteviot:
@gamebeaker 

Some advice my boss gave me
> Not all decisions are equal. Some are “one-way doors” — once you walk through them, you can’t go back. Those deserve time and care. But most decisions are “two-way doors” — they’re reversible. You can make a call, try it, and change direction if needed. 

I trust your judgement.  That's why I gave you commit rights. Mistakes will be made, we're only human.  And I've made my fair share of them.  But worst case, we just roll back the changes.

gamebeaker:
@dteviot Maybe i am going to merge more future PRs.

Another question:
At the moment i am creating PRs instead of pushing directly in ExperimentalTabMode for clear documentation what i did.
Is this clutter in your inbox and i should push directly in ExperimentalTabMode my changes or should i keep doing it like i am doing it now?

dteviot:
@gamebeaker 
Pushing directly to ExperimentalTabMode is fine.

PR #1933, opened 2025-05-16
Update download state icons and add tooltips

crybx:
The main goal of this PR is to add an icon (and tooltip) to help understand what is going on with the reading list, library, why things start out unchecked when you use them and what is going to be updated if you click "Add to Library" again.

I very much want to cache chapters downloaded and I read in a comment on an issue that the library kind of serves that purpose. But I couldn't understand how the library worked until I actually read the code and learned there was a reading list option and it was involved.

I still think there's more I want to do with caching that the library doesn't quite cover, but adding the icon for Previously downloaded chapters feels like step in the right direction to know what is even in the Library books.

Other changes in this PR are just because of fixes for spacing, typos, and editor warnings being made by me and my IDE that happened to be in the files for this change. I couldn't bring myself to delete them all when they make the code style more consistent.

Here's what it looks like when a book in your reading list or library is loaded with this PR:

1-new-icon

And here's how it looks when adding a subset of chapters to an existing library book. The download success icon I updated to match the new previously downloaded icon and look a bit more modern, but it could be reverted to the previous checkmark if you feel strongly about it.

2-add-chapter-subset

Also, I removed the text shadow on the buttons, because it's painful to look at in dark mode. This is before the shadow was removed:

3-text-shadow-in-dark-mode-hurts

Here's it removed:

image

And here's what the tooltips look like if you hover over the new state icons. I figured people would wonder what the heck it meant if they never saw it before:

Screenshot 2025-05-16 161122
Screenshot 2025-05-16 161147

I wasn't sure what to do about languages other than English though.

### DIFF
--- a/plugin/js/ChapterUrlsUI.js
+++ b/plugin/js/ChapterUrlsUI.js
@@ -72,16 +72,52 @@ class ChapterUrlsUI {
 
     static showDownloadState(row, state) {
         if (row != null) {
-            let img = row.querySelector("img");
+            let downloadStateDiv = row.querySelector(".downloadStateDiv");
+            ChapterUrlsUI.updateDownloadStateImage(downloadStateDiv, state);
+        }
+    }
+
+    static updateDownloadStateImage(downloadStateDiv, state) {
+        let img = downloadStateDiv.querySelector("img");
+        if (img) {
             img.src = ChapterUrlsUI.ImageForState[state];
+
+            // Update tooltip
+            let tooltipText = ChapterUrlsUI.TooltipForSate[state];
+            let tooltipTextSpan = downloadStateDiv.querySelector(".tooltipText");
+
+            if (tooltipText && !tooltipTextSpan) {
+                tooltipTextSpan = document.createElement("span");
+                tooltipTextSpan.className = "tooltipText";
+                tooltipTextSpan.textContent = tooltipText;
+                downloadStateDiv.appendChild(tooltipTextSpan);
+            } else if (tooltipText) {
+                tooltipTextSpan.textContent = tooltipText;
+            } else if (tooltipTextSpan) {
+                // Remove tooltip text if there is no text to display
+                downloadStateDiv.removeChild(tooltipTextSpan);
+            }
         }
     }
 
     static resetDownloadStateImages() {
         let linksTable = ChapterUrlsUI.getChapterUrlsTable();
-        let imgSrc = ChapterUrlsUI.ImageForState[ChapterUrlsUI.DOWNLOAD_STATE_NONE];
-        for(let img of linksTable.querySelectorAll("img")) {
-            img.src = imgSrc;
+        let prevDownload = ChapterUrlsUI.ImageForState[ChapterUrlsUI.DOWNLOAD_STATE_PREVIOUS];
+        let downloaded = ChapterUrlsUI.ImageForState[ChapterUrlsUI.DOWNLOAD_STATE_LOADED];
+
+        for (let downloadStateDiv of linksTable.querySelectorAll(".downloadStateDiv")) {
+            let state = ChapterUrlsUI.DOWNLOAD_STATE_NONE;
+            let imgSrc = downloadStateDiv.querySelector("img")?.src;
+            if (imgSrc) {
+                const imagesIndex = imgSrc.indexOf("images/");
+                if (imagesIndex !== -1) {
+                    imgSrc = imgSrc.substring(imagesIndex);
+                }
+            }
+            if (imgSrc === prevDownload || imgSrc === downloaded) {
+                state = ChapterUrlsUI.DOWNLOAD_STATE_PREVIOUS;
+            }
+            ChapterUrlsUI.updateDownloadStateImage(downloadStateDiv, state);
         }
     }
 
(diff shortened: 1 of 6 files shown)


gamebeaker:
> I wasn't sure what to do about languages other than English though.

Don't worry the translation is fall through. It tries to use the browser language if it isn't available for this variable it uses english.

I have little time this weekend so i am not going to look at the code.

plugin/js/main.js
```diff
@@ -512,8 +509,12 @@ var main = (function () {
         getManuallySelectParserTag().onchange = populateControls;
         document.getElementById("advancedOptionsButton").onclick = onAdvancedOptionsClick;
         document.getElementById("hiddenBibButton").onclick = onLibraryClick;
-        document.getElementById("ShowMoreMetadataOptionsCheckbox").addEventListener("change", function(){onShowMoreMetadataOptionsClick()});
-        document.getElementById("LibShowAdvancedOptionsCheckbox").addEventListener("change", function(){Library.LibRenderSavedEpubs()});
+        document.getElementById("ShowMoreMetadataOptionsCheckbox").addEventListener("change", function () {
```
dteviot, line 512: If rewriting this, could use lamba notation.

dteviot:
@crybx 

I do NOT like having formatting changes and functionality changes intermixed in a single commit.
It makes evaluating a change harder.
Rule of thumb.  Each commit should be as small as possible, but no smaller. (i.e. Should do only one thing, but everything needed to do that one thing.)
And if you can break a bit commit into multiple, self contained smaller commits, even better.
Doing that makes life easier on the reviewer.

I'm also of mixed opinion about the "already downloaded" icon.  
If I'm understanding it correctly, it makes sense if someone is using the library.  But, it's misleading and confusing if they are not.  (i.e. Why is it NOT showing that I've downloaded chapters that I downloaded last week?)
If I have misunderstood, please correct my understanding.

Anyway, I tried to review the changes, but as I estimate over 50% of the changes being formatting, I refuse to review until you put the formatting and functionality into separate commits.

crybx:
@dteviot 
I reverted everything that wasn't functionality for now. I also separated the text-shadow removal into a separate commit.

If the user is using the Reading List or the Library, it makes sense to show the "already downloaded" icon. The more I test with it, the more I get used to how it works, but I really was baffled by what the library was supposed to do. I care less about there even being an icon or tooltips as I get used to it, but I figured it was good to do something that would have helped someone less "used to it" before I forgot what it was like.

I have a few reasons for including the icon even for those who aren't using the Library or Reading List:
1. It shows some information about your current tab session and it seems fine to me for it to not show what you did in other sessions if you aren't using the currently available session bridging options.
2. Basic use won't make the icon show...a user would need to be redownloading after a successful download, without leaving the page. There are various reasons to do so, like making multiple epubs with different subsets of chapters, but the icon could be informative then.
3. It would have been more in-depth changes to keep the icon around only if they use the Reading List or Library once the user clicks "Pack EPUB" since that action clears the icons in a spot in the code without context for what the chapters are at the moment. Also, I'm pretty sure there would be other behavior to correct for doing multiple downloads in one session without using Library/Reading List, but overall, I generally prefer the icon always there anyway.
4. I was hoping for this to be a step towards caching for the current session that lets the user pause, restart, continue after halt, etc. without redownloading everything. And not keep that cache between sessions unless they choose to in Advance Options or use the Library or something. The icon would have even more meaning then.

crybx:
As for the formatting changes...how extensive is unwelcome?

I currently have my own extension I call Grabby that serves a similar but slightly different purpose than WebToEpub. Grabby only extracts the content from the active tab. I am working towards combining the two to give WebToEpub what I think of as a Live Mode feature. (That's what I'm mentally calling it where it loads the chapter in a tab to let javascript run, then gets the content. I know it's been discussed in issues and website requests.) My combined end result _could_ live separately as a permanent divergence, but I have kind of been testing the waters with how welcome changes to WebToEpub are.

The rest of this might be too much info, but my workflow is currently like this:
First time on a story I'm interested in, I use WebToEpub to generate an epub, if it can handle the site. As new chapters are added or unlocked, I use Grabby to dump the chapters to a folder as an html file. Then I use a Calibre plugin that searches the folder for filenames that match search terms set as metadata on my books and auto adds new chapters to my existing epubs. 

If WebToEpub can't handle the site because of javascript or account logins, etc, I get all the chapters with Grabby. If there are more than a few, I use AutoHotKey scripts to drive Grabby and even scroll for automated Google translate. And then I have a python script I put together that builds all the html files in a folder into an epub. I want to cut steps by combining the browser extensions.

Grabby is small, private, disregards content protection efforts in a way WebToEpub can't because of store policies, and focused only on sites with stories that interest me, but has dom utils and parsers (grabbers) per site like WebToEpub. I barely needed WebToEpub on Tapas to generate a few books from initial free chapters, but as I unlock chapters while reading, I tweak Grabby if something breaks or the formatting bothers me and a some of that can be copied back to WebToEpub.

dteviot:
@crybx 
> As for the formatting changes...how extensive is unwelcome?

Size isn't really a problem for formatting changes. 
By their nature, I can validate them basically on a line by line basis.  I don't have to keep the whole set of changes in my head.
And I realize I'm contradicting what I said earlier here.  But it's all about, how much mental effort does it take for the person inspecting the commit.

And I'd be delighted to see something like Grabby added to WebToEpub.

plugin/js/ChapterUrlsUI.js
```diff
@@ -221,40 +257,41 @@ class ChapterUrlsUI {
     * @private
     */
     static appendCheckBoxToRow(row, chapter) {
-        let col = document.createElement("td");
```
dteviot, line 224: Most of the changes made to  appendCheckBoxToRow() seem to be reformatting. lines 261 and 280 are the only functional changes.

dteviot:
@crybx 
>  testing the waters with how welcome changes to WebToEpub are

I would like to think I'm mostly welcoming of changes.  But I try to evaluate them against the following questions:

1. Will this be difficult for users to understand?
2. Is this likely to cause breaking changes?
3. Is this likely to cause problems in future / how had will it be to maintain?  (e.g. if site changes)
4. Is the added complexity worth the effort?
5. How well does it match the WebToEpub philosophy?
6. Will this get WebToEpub banned from the Chrome store?  

The WebToEpub design philosophy is (I think):
Keep it as simple as possible: Grab pages and pack into an Epub.  In general, Err on grabbing and packing too much, instead of removing everything that's not wanted, in case it removes too much.  So, only remove things we can be very confident are not wanted.)  Any extra processing should be done by a stand alone tool.

Note that the Library function was almost entirely implemented by Gamebeaker, not me.

Of course, you could always ask me of gamebeaker if you think a feature would be accepted before you work on it.

And of course, thank you again for your contributions.

plugin/js/ChapterUrlsUI.js
crybx: Sorry, I missed this method when reverting formatting changes!

PR #1935, opened 2025-05-17
delete it Im learning git sorry

tiklii:


dteviot:
@tiklii 

No problem.  Always good for people to learn. Anything I can do to help?

dteviot:
@tiklii 
>It was just me trying to revert the speed logic change for private use

Ah. I thought it might be something like that.
I wish you success.

PR #1952, opened 2025-05-24
Miscellanous diffs

crybx:
Why move FootnoteExtractor to its own file: it makes it easier to work on the util file when the unrelated class isn't at the bottom.

Why move Information page header out of div: Semantically, it makes sense for the h1 to be the top item alone. But also, because I work so heavily with Calibre, this is another thing I end up having to manually edit for every epub. When Calibre generates Table of Contents from files, it uses the first element in the page. If that's a div, it includes everything in the div. So I have to move the h1 above the div or have the table of contents url smooshed in the ToC entry.

image

image


crybx:
Since Chrome doesn't support webRequestBlocking for manifest v3, there's an error that's always present when loading the extension from unpacked source code:
image

As far as I can tell, Chrome can run the extension from unpacked source code and Firefox can't. And there's already custom build logic for separate manifests, so it makes more sense to add the permission there for Firefox than remove it for Chrome.

eslint/pack.js
```diff
@@ -225,8 +233,6 @@ var makeManifestForChrome = function(data) {
     let manifest = JSON.parse(data.toString());
     delete(manifest.browser_specific_settings);
     delete(manifest.action.browser_style);
-    manifest.permissions = manifest.permissions
-        .filter(p => !p.startsWith("webRequest"));
```
gamebeaker, line 229: Why aren't you removing the webRequest permission for Chrome?

crybx: I totally didn't catch that Chrome was removing "webRequest" **and** "webRequestBlocking" with this, I guess because it doesn't throw any errors to leave webRequest. I undid the delete, so now it should remove webRequest for Chrome, add webRequestBlocking for Firefox and not show warning/error when Loading from unpacked source in Chrome.

PR #1959, opened 2025-05-26
Util.js refactor

crybx:
This PR is purely a "refactor / formatting changes" one that doesn't change any external functionality. It's only focused on the Util.js file, but two function renames are why any of the rest of the files were updated.

----

Changes to Util.js
- change from using function expressions assigned to variables to using function declarations and direct internal calls
- fix semicolons and spacing to be consistent (used lint to do this)
- rename 2 functions: 
	- extactSubstring to extractSubstring (because typo)
	- removeChildElementsMatchingCss to removeChildElementsMatchingSelector (for clarity)
- replace use of now deprecated substr()
- other small fixes that don't affect functionality like replacing `!=` with `!==` where it can be determined to be fully safe (to avoid IDE warnings about type coercion) or replacing var with let or const.

Refactoring Util.js from referencing functions via `util.functionName()` to using direct function calls within the module improves several things:

- Code outside of Util.js that uses the `util` object will continue to work with no changes.
- Improved static analysis: Code analysis tools can better trace function dependencies, improving IDE features like "Find References", "Go to Definition" or detection of unused code.
- Less noise/code: Removes `util.` prefixes within the file and 
	- `function sleep(ms) {` is slightly shorter and easier to read than 
	- `var sleep = function(ms) {`
- Function call optimization: Each time a function calls another internal function via `util.functionName()`, JavaScript must perform an object property lookup, which adds execution time.
- Direct reference efficiency: Direct function calls (`functionName()`) avoid this property lookup entirely, providing a small performance improvement.
- Reduced call stack: Simpler call paths with fewer indirections improve performance for deeply nested function calls.
- JIT optimization: Modern JavaScript engines can better optimize direct function calls compared to property lookups.

Performance improvements would be **very** small, but utility functions are meant to be called a lot.

### DIFF
--- a/plugin/js/Util.js
+++ b/plugin/js/Util.js
@@ -1,70 +1,70 @@
 /*
     General dumping ground for misc functions that I can't find a better place for.
-    Warning: Don't look at this too closely, or you may loose your sanity.
-    Side Note: Putting these all in one place may not have been a good idea. 
+    Warning: Don't look at this too closely, or you may lose your sanity.
+    Side Note: Putting these all in one place may not have been a good idea.
     I think they're breeding. There seem to be more functions in here that I didn't create.
 */
 
 "use strict";
 
-var util = (function () {
+const util = (function() {
 
-    var sleep = function(ms) {
+    function sleep(ms) {
         return new Promise(resolve => setTimeout(resolve, ms));
     }
 
-    var randomInteger = function(min, max) {
+    function randomInteger(min, max) {
         return Math.floor(Math.random() * (max - min + 1)) + min;
     }
 
-    var isFirefox = function() {
-        return (typeof(browser) !== "undefined");
+    function isFirefox() {
+        return (typeof (browser) !== "undefined");
     }
 
-    var extensionVersion = function() {
-        let runtime = util.isFirefox() ? browser.runtime : chrome.runtime;
+    function extensionVersion() {
+        let runtime = isFirefox() ? browser.runtime : chrome.runtime;
         // when running unit tests, runtime is not available
-        return (typeof(runtime) === "undefined") ? "unknown" : runtime.getManifest().version;
+        return (typeof (runtime) === "undefined") ? "unknown" : runtime.getManifest().version;
     }
 
-    var createEmptyXhtmlDoc = function() {
-        let doc = document.implementation.createDocument(util.XMLNS, "", null);
-        util.addXhtmlDocTypeToStart(doc);
-        let htmlNode = doc.createElementNS(util.XMLNS, "html");
+    function createEmptyXhtmlDoc() {
+        let doc = document.implementation.createDocument(XMLNS, "", null);
+        addXhtmlDocTypeToStart(doc);
+        let htmlNode = doc.createElementNS(XMLNS, "html");
         doc.appendChild(htmlNode);
-        let head = doc.createElementNS(util.XMLNS, "head");
+        let head = doc.createElementNS(XMLNS, "head");
         htmlNode.appendChild(head);
-        head.appendChild(doc.createElementNS(util.XMLNS, "title"));
+        head.appendChild(doc.createElementNS(XMLNS, "title"));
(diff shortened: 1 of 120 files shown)


gamebeaker:
> change from using function expressions assigned to variables to using function declarations and direct internal calls

don't merge there is a problem with this i am searching

plugin/js/Util.js
```diff
@@ -596,69 +598,69 @@ var util = (function () {
...
+        let noTrailingSlash = removeTrailingSlash(removeAnchor(url));
+
         const protocolSeperator = "://";
         let protocolIndex = noTrailingSlash.indexOf(protocolSeperator);
-        return (protocolIndex < 0) ?  noTrailingSlash
+        return (protocolIndex < 0) ? noTrailingSlash
             : noTrailingSlash.substring(protocolIndex + protocolSeperator.length);
     }
 
-    var hyperLinkToChapter = function(link, newArc) {
```
gamebeaker, line 626: I found it:

https://github.com/gamebeaker/WebToEpub/blob/44a9a32cccb98e292b0feff8b4cfc2f292c25394/plugin/js/parsers/MadnovelParser.js#L12-L22

There is the definition changed. (I found it as i tested other things and it broke my logic)
I don't know if this happens somewhere else with another variable.
@dteviot was there a reason that it wasn't just functions?

crybx: Both before and after, the full util object is an Immediately Invoked Function Expression with a public api that's still the same. What broke here?

crybx: The extension won't load for me on novelbuddy.com, but that's true with and without the PR changes. So the thing that util.hyperlinksToChapter is overriding here is the property returned by the public api (the stuff in the return statement). Overriding it should still work, even if what is being assigned isn't the same format as what it was when it was set by 
```
return {
.
.
hyperlinksToChapterList: hyperlinksToChapterList,
.
.
}
```
The thing on the right is the function in the util class, the thing on the left is the property returned for everywhere else to use. The MadnovelPaerser snippet would reassign the property on the left to the new content, no longer referencing the function in Util.js. It should be fine.

crybx: Oh, it's overriding util.**hyperLinkToChapter** and then calling util.**hyperlinksToChapterList**, I missed those were two different calls. Before the refactor, hyperlinksToChapterList would call the overridden version and after it would internally call the original version. Hmm.

crybx: I think the correct answer is to find everywhere that was overriding util.functionName and maybe write new functions for them to use. I can do that for this PR.

crybx: I found only one instance of a util function override in the entire codebase. So I ported the logic MadnovelParser was trying to use from the util function directly into the getChapterUrls method and simplified it since it wasn't using all the possible parameters of hyperlinksToChapterList anyway.

Given that this would have broken other parsers before and MadnovelParser doesn't appear to work at all for me at the moment, I don't think this affects whether this refactor is safe to merge. Besides this scenario (which was bad to do anyway) I can't think of another reason the new syntax would change anything.

crybx: I don't know WHAT was going on, but now it works. I've confirmed I can get the chapter urls and names after my changes:
image

crybx:
> If it is already getting refactored why not make it a class? I think everything else is a class.

Currently, Util.js acts as a singleton, a single, global util object that groups a collection of helper functions and constants. There's no apparent need to create multiple util instances, each with its own distinct state. If we made it a class, it would most naturally be a static class with all static methods and the convention for classes is capitalized names (Util). That would mean updating everything that calls `util` to `Util` or sticking with `util` and having various tools or possibly complain about not matching conventions. And internal calls between functions would then need `this.` or `Util.` which brings back some of extra code that the refactor currently removes.

Also, this is a totally selfish reason, but part of the motivation for this refactor for me was to match what I do in my own extension, Grabby, with its dom utils. This was basically prep work for combining them and using the same utilities.

Issue #2016, opened 2025-06-29
Image Compression


Kiradien:
**Is your feature request related to a problem? Please describe.**
Image sizes are getting larger and larger - our e-readers are only so big. Looking at a 2000x2500 image on a 1920x1080 screen will never exceed 1920 pixels in any direction, and most e-readers can't zoom in. Give us smaller images!

**Describe the solution you'd like**
Honestly - it's mostly done aside from a small bug related to async action. Would like other devs to offer suggestions, but it does mostly work.

**Additional context**

![Image](https://github.com/user-attachments/assets/e085146b-7635-4622-914a-685e2dd16f3f)

PR #2017, opened 2025-06-29
Add image compression logic

Kiradien:
- Added checkbox to choose to compress images
- Added field to choose maximum resolution (height/width)
- Passed image object generated to get image dimensions back for secondary usage
- Replaced image from arraybuffer to converted jpeg

Code has one flaw - the last chapters encoded may not be fully compressed. Images still remain, however at full size. Any suggestions to resolve the async issue would be appreciated.

### DIFF
--- a/plugin/_locales/en/messages.json
+++ b/plugin/_locales/en/messages.json
@@ -251,6 +251,14 @@
 		"message": "Add Page with Chapters to Chapters List",
 		"description": "Label for checkbox to toggle adding the page with chapters to list of pages to fetch"
 	},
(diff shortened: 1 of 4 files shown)


Kiradien:
This is connected to issue #2016

Kiradien:
I actually ran into a solution after a nights rest, dreamt of the damn issue; That said, we may need to revert since it has the potential to run into breaking issues if js default handling of toBlob failures isn't elegant... Hopefully the try-catch I dropped in there will suffice.
After a bit more dev testing, unless something critical comes up, I'll squash&merge it this afternoon

Issue #2045, opened 2025-07-18
Handle Mozilla rejection of WebToEpub


dteviot:
Notification from Mozilla:
> Your Extension WebToEpub was manually reviewed by the Mozilla Add-ons team in an assessment performed on our own initiative of content that was submitted to Mozilla Add-ons.
Our review found that your content violates the following Mozilla policy or policies:
- Security, specifically Unsanitized DOM injection: This add-on is creating DOM nodes from HTML strings containing potentially unsanitized data, by assigning to innerHTML, jQuery.html, or through similar means. Aside from being inefficient, this is a major security risk. For more information, see https://developer.mozilla.org/en-US/Add-ons/WebExtensions/Safely_inserting_external_content_into_a_page . Here are some examples that were discovered:
js\parsers\AsianfanficsParser.js - line 55
js\parsers\AppYoruWorldParser.js - line 78
 js\parsers\StellarRealmParser.js - line 91
js\parsers\TapreadParser.js - line 88
js\parsers\YushuboParser.js - lines 37,38
js\parsers\ZirusMusingsParser.js - line 78
js\HttpClient.js - line 311
js\Parser.js - line 799.
Affected versions: 1.0.4.0, 1.0.5.0, 1.0.6.0
Based on that finding, those versions of your Extension have been disabled on https://addons.mozilla.org/addon/webtoepub-for-baka-tsuki/ and are no longer available for download from Mozilla Add-ons, anywhere in the world. Users who have previously installed those versions will be able to continue using them.
You may upload a new version which addresses the policy violation(s).
More information about Mozilla's add-on policies can be found at https://extensionworkshop.com/documentation/publish/add-on-policies/.
If you believe that you did not violate Mozilla's policies, or that this decision was otherwise made in error, you have the right to appeal this decision within 6 months from the date of this email. See https://addons.mozilla.org/abuse/appeal/eb49bdc9-4e1a-4497-a974-06346d27e53a/ for details on the appeal process, including how to file an appeal for this specific decision. You may also choose to have this decision reviewed by a third party neutral arbiter, or to seek judicial redress in a court of law.
Thank you for your attention.
[ref:eb49bdc9-4e1a-4497-a974-06346d27e53a]

I have responded pointing out that:
1. Data IS sanitized before being packaged into the epub (Parser.js line 170)
2. Sanitizing at every point where HTML is received is not viable, as a number of sites put the content in &lt;script&gt; elements.
[...]

PR #2056, opened 2025-07-24
Add site WeTriedTls

basitcodeenv:
Reaperscans.com has been shut down due to a DMCA notice, and now they are updating chapters on the new site wetriedtls.com.

### DIFF
--- a/plugin/js/parsers/WetriedTlsParser.js
+++ b/plugin/js/parsers/WetriedTlsParser.js
@@ -1,0 +1,123 @@
+"use strict";
+
+parserFactory.register("wetriedtls.com", () => new WetriedTlsParser());
+
+class WetriedTlsParser extends Parser {
+    constructor() {
+        super();
+        this.minimumThrottle = 2000;
+    }
+
+    async getChapterUrls() {
+        const chapterJson = (
+            await HttpClient.fetchJson(
+                "https://api.wetriedtls.com/chapters/" +
+                    this.id +
+                    "?page=1&perPage=9999&order=asc"
+            )
+        ).json;
+
+        const chapters = chapterJson.data.map((chapter) => {
+            const seriesUrl =
+                "https://wetriedtls.com/series/" + chapter.series.series_slug;
+
+            return {
+                sourceUrl: `${seriesUrl}/${chapter.chapter_slug}`,
+                title: chapter.chapter_title,
+            };
+        });
+
+        return chapters;
+    }
+
+    async loadEpubMetaInfo(dom) {
+        const bookTitle = dom
+            .querySelector("section .text-xl")
+            .textContent.trim();
+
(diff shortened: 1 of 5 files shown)


plugin/js/parsers/WetriedTlsParser.js
```diff
@@ -0,0 +1,125 @@
...
+		if (!htmlChunk) throw new Error("No HTML chapter content found.");
+
+		const rawHtml = htmlChunk[1]
+			.replace(/\\u003c/g, "<")
+			.replace(/\\u003e/g, ">")
+			.replace(/\\"/g, '"')
+			.replace(/\\\\/g, "\\");
+
+		const container = document.createElement("div");
+		container.innerHTML = rawHtml;
```
gamebeaker, line 95: @dteviot is that ok?
Isn't that the problem as here: #2045

dteviot:
@basitcodeenv 

Also, after fixing the lint issues, please squash the commits.

plugin/js/parsers/WetriedTlsParser.js
```diff
@@ -0,0 +1,123 @@
...
+class WetriedTlsParser extends Parser {
+    constructor() {
+        super();
+        this.minimumThrottle = 2000;
+    }
+
+    async getChapterUrls() {
+        const chapterJson = (
+            await HttpClient.fetchJson(
+                "https://api.wetriedtls.com/chapters/" +
```
dteviot, line 14: Perhaps
```javascript
`https://api.wetriedtls.com/chapters/${this.id}?page=1&perPage=9999&order=asc`
```

plugin/js/parsers/WetriedTlsParser.js
```diff
@@ -0,0 +1,123 @@
...
+        );
+
+        if (!htmlChunk) throw new Error("No HTML chapter content found.");
+
+        return this.buildChapter(htmlChunk[1], url);
+    }
+
+    buildChapter(rawHtml, url) {
+        let newDoc = Parser.makeEmptyDocForContent(url);
+        let content = new DOMParser().parseFromString(
```
dteviot, line 94: @basitcodeenv
DOMParser().parseFromString is no longer accepted by Mozilla unless we can show that it's safe.  See https://github.com/dteviot/WebToEpub/issues/2045
Use
```javascript
 let content = util.sanitize(rawHtml); 
 ```

basitcodeenv: I didn't use it because I'm encountering a TypeError: util.sanitize is not a function.

basitcodeenv: I looked for the `sanitize` function in `plugin/js/Util.js`, but I couldn't find any instances of that function.

Issue #2057, opened 2025-07-26
Use libwebpjs to automatically convert webp files


marcussacana:
Here a working demo of this library: http://libwebpjs.hohenlimburg.org/v0.6.0/

The next WebToEpub version may add support to the new host "global.novelpia.com", this host uses webp not only for the illustration but for the cover as well. Currently the extension can detect the webp file but does not convert it.
With this library I think is possible to add support to the format.

Also, in this site the images has a bad extension, so, after convert the extension should save the image as .png.

May be useful as well: https://gist.github.com/deepak1556/7845796

dteviot:
@marcussacana 

IIRC, version 1.0.6 can already convert images, you just need to go to the advanced settings and turn on image compression.

> Also, in this site the images has a bad extension, so, after convert the extension should save the image as .png.

I have no idea what you're trying to say.

marcussacana:
I see.

Regarding the bad extension, here is the link to a random novel cover from that site: https://gn.novelpia.com/upload/novel/1695/af2766288d50bd3a54380dd4ca4b357b_234039_ori.file
This isn't directly related to the WebP conversion, but some EPUB readers fail to render the book cover because the file extension is not a valid extension (.png, .jpg, etc)

As for the compression, it is indeed there, but I had no idea it would convert WebP files into image formats.
In any case, it would be good to at least have an option to perform a lossless conversion instead of compression (I assume this exports images to JPG instead of PNG).

To summarize what I propose:
- Lossless image conversion
- Ensure downloaded images are saved with a valid image file extension

Kiradien:
Just to confirm for the previous message: Yes, the compression logic exports to jpg at 90% quality and sets max resolution (width or height) to the compressed Resolution value - so by default, a 4000x4000 image would compress down to 1080x1080.

Random note relating to this: A lot of sites actually store their images in ways that do not translate well to having an extension in WebToEpub - It's actually more common that my image files lack extensions than have them, but I never considered it much of an issue until now since the readers I use ignore the issue. The compressed images actually function similarly.

As part of this, we may want to grab the mediaType when saving and force an additional extension; for this particular issue it'd at least translate them to the .webp extension.

*stares at code* I'm just not entirely sure where the 'filename' is actually set before saving. I wanted to do something about this before but considered it a non-issue since it 'just worked'.

marcussacana:
[...]

Issue #1887, opened 2025-05-05
Bug regression from 1.0.4.0 never fixed.


DefectiveByDesign:
Version 1.0.4.0
Released Mar 14, 2025 - 442.65 KB
Works with firefox 115.0 and later, android 120.0 and later
Previous version created malformed epub file that did not work with some epub readers.
Fixed in this release. 

I would argue, this was not fixed, and still causes problems. Quick novel, an epub reader app cannot load anything from the current version.

The last epub I have that works is from the 1.0.2.0 version. The current version will load in librera, but not quick novel, and quick novel is a better epub reader.

You should be able to easily reproduce this issue. Quick novel has the option to create epub files too, but there are some sites it doesn't work on. Just load an epub into quick novel from the current version, and it will fail to load. 

I have tried switching the epub3 option, but it doesn't seem like there's any working solution for compatibility. Unless I downgrade to an older version, which is possible on Firefox, but idk about Chrome, and this is not viable for long term use. 

**Describe the bug**

**To Reproduce**

**Expected behavior**

**Screenshots**

**Desktop (please complete the following information):**

**Smartphone (please complete the following information):**

**Additional context**

dteviot:
@DefectiveByDesign 

It may be worth looking into this a little more.
Can you:
1. Create two Epubs with a single chapter  (both from the same source) one using 1.0.2.0 and 1.0.5.0, 
2. Confirm that both epubs work with librera
3. Confirm that the 1.0.2.0 works with quicknovel, and the 1.0.5.0 does not
4. Email me the epubs, to examine in detail.

dteviotdale:
@Aberto7 added

> The One that worked:
```html
<manifest>
<item href="Images/0000_000992869.jpg" id="cover-image" media-type="image/jpeg"/>
<item href="Text/0000_Chapter_1.xhtml" id="xhtml0000" media-type="application/xhtml+xml"/>
```

> The One that didn't work:
```html
<manifest>
<item href="Images/0000_000992869.jpeg" id="cover-image" media-type="application/octet-stream"/>
<item href="Text/0000_Chapter_1.xhtml" id="xhtml0000" media-type="application/xhtml+xml"/>
```

Well, that's your problem.  :)  
Seriously.  That probably is the problem.  Should be an easy fix.
Thanks for looking into it.

dteviotdale:
Additional notes.

[...]

PR #2059, opened 2025-07-27
Pull image extension from Mime type

Kiradien:
Site links fail to provide appropriate image extension - pull from mime type.

Created simple dictionary lookup for extension based off mime type - reorganized to use mime extension as priority, link extension as secondary.

PR is adjacent to #2057 but does not fulfil original request.

### DIFF
--- a/plugin/js/EpubItem.js
+++ b/plugin/js/EpubItem.js
@@ -183,7 +183,7 @@ class ImageInfo extends EpubItem {
 
     getZipHref() {
         let that = this;
-        let suffix = that.findImageSuffix(that.wrappingUrl);
+        let suffix = util.getDefaultExtensionByMime(that.mediaType) || that.findImageSuffix(that.wrappingUrl);
         return util.makeStorageFileName("OEBPS/Images/", that.index, that.getImageName(that.wrappingUrl), suffix);
     }
 
--- a/plugin/js/Util.js
+++ b/plugin/js/Util.js
@@ -1088,7 +1088,7 @@ const util = (function() {
         "image/ktx": ["ktx"],
         "image/apng": ["apng"],
         "image/avif": ["avif"]
-      };
+    };
 
     return {
         XMLNS: XMLNS,
--- a/plugin/js/Util.js
+++ b/plugin/js/Util.js
@@ -1041,6 +1041,13 @@ const util = (function() {
         element.appendChild(wrapper);
     }
 
+    function getDefaultExtensionByMime(mimeType)
+    {
+        let retval = MIME_TYPE_EXTENSIONS[mimeType];
+        if (retval) retval = retval[0];
+        return retval;
+    }
+
     // Define constants
     const XMLNS = "http://www.w3.org/1999/xhtml";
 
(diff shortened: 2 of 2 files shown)


Kiradien:
Code could be improved by implementing full mime-types library, however this minimal dictionary implementation should suffice for most existing image types.

Kiradien:
For reference with compression active.
image
In the attached, 0000, 0004 and 0019 originally failed to get appropriate extensions (More were wrong due to compression, however).

dteviot:
@Kiradien 

Just to ruin your day, you might like to look at https://github.com/dteviot/WebToEpub/issues/1887#issuecomment-3106302665
And it's matching commit https://github.com/dteviot/WebToEpub/commit/4d6a8aac781a18cdd581b42257d0b11d0d503dae

Kiradien:
Just double checked and reverted to 1.0.6.5 to confirm output - though that changeset does 'fix' mime types, it doesn't do anything for the file extension. The output of the file matched the right side of the above screenshot.

Just to note, using url extensions does sometimes result in odd issues such as a mimetype of "image/php", "image/dib" or even "image/jpe" - those ones can stay a little wonky when the compression is turned off.

dteviot:
@Kiradien 

I guess I wasn't clear.  Your change was "don't trust file extensions, trust the mime type".  
My point was, "You can't trust the mime type 100% either."

I'm not sure there's anything we can do about it, short of examining the raw image bytes to figure out what the image type is.  (Which is what I suspect browsers actually do.)

And if we were going to fix it, I'd probably look at a stand alone tool, like EpubEditor to try fixing up the epubs as a post collection step.  Rather than continue to complicate WebToEpub.

Kiradien:
@dteviot Ah, yeah, I agree that it's not a perfect fix - It's just a simple additonal step at the end of the process to ensure it at least has a usable extension. 
I agree that overcomplicating the epub generation process isn't the best idea, however this is a relatively small change that - along with your previous change - fixes a good 90% of cases I know about, not to mention the 100% number of cases fixed when using compression.

If you want I can certainly revert the fix, but I don't think that's what you're saying.

As for what we could do about it - the easiest solution is using a package to convert everything to png, or just using the canvas logic in compression to do the same conversion (But losing everything after the first frame for anything animated and a much higher overhead on large images)
There's also the overcomplicated solution of pulling the mimetype from the base64 string https://stackoverflow.com/questions/57976898/how-to-get-mime-type-from-base-64-string - honestly not a fan of those magic numbers, but it would work as long as we built up a library of them.

Kiradien:
Thinking about magic numbers, I might actually play with that a bit later - but for reference this should be a complete list in JSON format

`{
  "image/jpeg": ["/9j/"],
  "image/png": ["iVBORw0KGgo="],
  "image/gif": ["R0lGODdh", "R0lGODlh"],
  "image/webp": ["UklGR_________VEJC"],  // WebP has a RIFF header with variable bytes
  "image/bmp": ["Qk0="],
  "image/tiff": ["SUkqAA==", "TU0AKg=="],
  "image/svg+xml": ["PD94bWw="],  // "<?xml"
  "image/x-icon": ["AAABAA=="],
  "image/vnd.microsoft.icon": ["AAABAA=="],
  "image/heif": ["ZnR5cGhlaWNj", "ZnR5cG1pZjE="],
  "image/heic": ["ZnR5cGhlaWNj"],
  "image/jxr": ["SUm8"],
  "image/ktx": ["q0tUWCAxMb0NCgo="],
  "image/x-tga": ["AAACAA=="],
  "image/apng": ["iVBORw0KGgo="],  // Same as PNG; difference is in deeper chunks
  "image/avif": ["ZnR5cGF2aWY="],
  "image/x-xbitmap": [],
  "image/x-portable-bitmap": ["UDAx"],
  "image/x-portable-graymap": ["UDAy"],
  "image/x-portable-pixmap": ["UDAz"],
  "image/x-portable-anymap": ["UDA0"],
  "image/x-cmu-raster": ["WaZqlQ=="]
}`

PR #2061, opened 2025-07-27
Generate MIME from signature

Kiradien:
Added Secondary datasource in fixupInvalidMediaType - generate the mime based off the signature (magic number) at the start of the object. 
Comparison done in Base64 for human readability.

Potential fix for what is discussed in #2059

@dteviot - I'll leave it up to you whether you want to go ahead with this or avoid complicating WebToEpub further.
Note it could theoretically run smoother as a binary comparison rather than Base64, but readability and complexity suffers.

### DIFF
--- a/plugin/js/EpubItem.js
+++ b/plugin/js/EpubItem.js
@@ -187,7 +187,7 @@ class ImageInfo extends EpubItem {
     }
 
     getBase64(maxLength) {
-        var binary = '';
+        var binary = "";
         var bytes = new Uint8Array(this.arraybuffer);
         var len = bytes.byteLength;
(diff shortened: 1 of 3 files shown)


Kiradien:
I'll be honest, I have no idea what triggered `npm run web-ext` to fail here; It's complaining about dompurify, which this code should not touch. 
Don't have time ATM to dig further into it; we can just reject this PR if needed. Hopefully it doesn't end up being endemic

dteviot:
@Kiradien 

I added DOMPurify last night.  It was required by Mozilla due to their security concerns.  Ironic that it's triggering warnings with their security checks.  Anyway, I've disabled running web-ext in the build.  So that should solve that problem for the moment.

Issue #2064, opened 2025-07-28
Kemono.party moved domain to kemono.cr


Davidx321:
kemono site moved domain from kemono.party to kemono.cr and now parser doesn't work.

Kiradien:
Test versions for Firefox and Chrome have been uploaded to https://github.com/dteviot/WebToEpub/releases/tag/developer-build. Pick the one suitable for you, follow the "How to install from Source (for people who are not developers)" instructions at https://github.com/dteviot/WebToEpub/tree/ExperimentalTabMode#user-content-installation and let me know how it goes.

gamebeaker:
@Kiradien your fix broke the update functionality of the library.
To test is open WebToEpub and try to load a kemono.cr link with "Load and Analyse" it redirects you to the standard parser.

dteviot:
@Davidx321 

Updated version (1.0.9.0) has been submitted to Firefox and Chrome stores.
Firefox version should be available within two weeks (but might take 4 to 6 weeks)
Chrome might be available in a few hours to 21 days.  (Typically 3 to 7 days).

PR #2065, opened 2025-07-28
#2064 update kemono rule registration

Kiradien:
Fix for #2064 - Since this apparently isn't the first time the domain has changed TLD's (Was originally .party), changed recognition to use rule-based instead of domain based logic.


dteviot:
@Kiradien Thanks

PR #2081, opened 2025-08-03
#2080 temp fix

### LINKED ISSUE #2080: Help github actions
`npm run lint` fails with this error message:

Image

https://github.com/dteviot/WebToEpub/actions/runs/16703968008/job/47279327333

I don't have this problem locally can someone reproduce this?

Kiradien:
Force zip.js to last working version.

Temporarily fix #2080 - Will probably need code change for permanent fix.

### DIFF
--- a/package.json
+++ b/package.json
@@ -32,167 +32,63 @@
     "type": "git",
     "url": "git+https://github.com/dteviot/WebToEpub.git"
   },
-  "keywords": [
-    "epub"
-  ],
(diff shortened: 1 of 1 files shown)


Kiradien:
Just to be clear, this isn't meant as a final fix; we're forcibly setting to an older (though not yet old) version and preventing any future updates. Eventually either Chrome or Firefox will block us because of this, but hopefully zip.js fixes it on their side.

PR #2085, opened 2025-08-04
Update KemonopartyParser.js

Kiradien:
Fix Cover Image detection on Load&Analyze referenced in #2064

Not sure if we actually want this; Load & Analyze was throwing errors, though.


gamebeaker:
@Kiradien i catch such errors in the update all function from the library as they happen often. To ignore them as the cover, author etc. aren't needed if you are just updating an existing story.

Kiradien:
I'll close the PR and leave it here then; honestly don't see it likely people will 'load and analyze' that site.

gamebeaker:
@Kiradien agree.
As the Library uses the same function to search for new chapters i try to make it so that the chapters get found even if the metadata search doesn't work. (That is so that the stories can be updated automatically.)

PR #2089, opened 2025-08-05
create lint:fix command and add more eslint rules

crybx:
- Additional eslint rules for consistent spacing, semicolon usage, and some code formatting.
- Added command `npm run lint:fix` to make fixing what doesn't match the rules easy.
- File changes from running the new command with the new rules.

This continues to help me reduce diffs with the versions of the plugin I run all the time: (https://github.com/crybx/WebToEpub/tree/dev-v2 if interested. Despite the diverging ahead/behind commit numbers, I keep cherry picking or rebasing updates from the original repo and merging the diffs is a pain, and this branch is from before I started merging with another extension 😭)

### DIFF
--- a/eslint/.eslintrc.js
+++ b/eslint/.eslintrc.js
@@ -11,22 +11,36 @@ module.exports = {
     "rules": {
         "indent": [
             "error",
-            4
+            4,
+            { "SwitchCase": 1 }
         ],
         "linebreak-style": "off",
         "no-extra-semi": [
-            "off"
+            "error"
         ],
         "quotes": [
             "error",
             "double"
         ],
         "semi": [
-            "off",
-            "never"
+            "error",
+            "always"
         ],
+        "no-undef": "off",
+        "no-unused-vars": "off",
+        "space-before-blocks": ["error", "always"],
+        "space-before-function-paren": ["error", {
+            "anonymous": "never",    // function() {}
+            "named": "never",        // function foo() {}
+            "asyncArrow": "always"   // async () => {}
+        }],
+        // Space around keywords (if, for, etc.)
+        "keyword-spacing": ["error", {
+            "before": true,
+            "after": true
+        }],
     },
-	"globals": {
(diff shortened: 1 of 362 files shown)


eslint/.eslintrc.js
```diff
@@ -11,22 +11,36 @@ module.exports = {
...
             "double"
         ],
         "semi": [
-            "off",
-            "never"
+            "error",
+            "always"
         ],
+        "no-undef": "off",
+        "no-unused-vars": "off",
```
dteviot, line 30: Why do you think no-undef and no-unused-vars should be turned off?

dteviot:
Will accept as is, can decide on the lint rules later.
As it's touched almost every file, I don't want to have to fixup a merge conflict if someone commits before we've decided.
And it does make the code a bit more consistent.  And we need to run lint anyway.

Note, My reason for running lint was to look for potential bugs, rather than code style.  The fact that we need to fix the style, I'm not entirely sure is a benefit.  But is necessary to keep the warnings clean.  (Yes, I could probably turn off the indentation warnings. But, well, it works well enough, can't be bothered trying to mess with it.)

I'll note, we probably missed an opportunity to replace all the 
```javascript
parserFactory.register("gravitytales.com", function() { return new GravityTalesParser() });
```
with
```javascript
parserFactory.register("gravitytales.com", () => new GravityTalesParser()});
```
As lint was reformatting them to add the missing colon.

eslint/.eslintrc.js
crybx: @dteviot This was for the sake of making fixing lint errors automatically possible. Without turning those off, the `npm run lint:fix` does this for me: 
no-undef
no-unused-vars

These things ARE all defined when the files are concatenated together into one big packed.js (or running and added with script tags to the html), but when running lint to auto fix the individual files without stitching them together, they don't exist. I believe it's possible to tell eslint which things exist.

PR #2108, opened 2025-08-10
Fictioneer website fixes

crybx:
- Removes non-chapter chapter links from several sites.
- Removes various extra stuff from Lily on the Valley
- Removes separate smeraldogarden.com parser - it's a Fictioneer theme site and the Fictioneer Parser does more, like only select unlocked chapters to start


plugin/js/parsers/FictioneerParser.js
```diff
@@ -30,15 +30,15 @@ class FictioneerParser extends Parser {
     async getChapterUrls(dom) {
         let chapters = [];
         // Put free chapters first
-        [...dom.querySelectorAll("._publish a")].map(a => chapters.push(({
+        [...dom.querySelectorAll(".chapter-group__list ._publish a")].map(a => chapters.push(({
             sourceUrl: a.href,
-            title: a.textContent,
+            title: a.textContent?.trim(),
```
dteviot: This should really be in onLoadFirstPage() in Parser.js.  So that ALL parsers can take advantage of this.

plugin/js/parsers/FictioneerParser.js
```diff
@@ -30,15 +30,15 @@ class FictioneerParser extends Parser {
...
             isIncludeable: true
         })));
         // Put scheduled chapters after free and don't select them
         [...dom.querySelectorAll("._future a")].map(a => chapters.push(({
             sourceUrl: a.href,
-            title: a.textContent,
+            title: a.textContent?.trim(),
```
dteviot: Ditto

plugin/js/parsers/FictioneerParser.js
```diff
@@ -112,11 +112,35 @@ class FictioneerParser extends Parser {
...
     }
 }
+
+class LilyOnTheValleyParser extends FictioneerParser {
+    constructor() {
+        super();
+    }
+
+    customRawDomToContentStep(chapter, content) {
+        content.querySelectorAll("*").forEach(element => {
```
dteviot: Should not be doing this in WebToEpub.  This is "anti-copying" code.  Removing it would be a violation of Googles ToC.
Instead, we have provided a script for EpubEditor to remove this.

dteviot (changes requested):

plugin/js/parsers/FictioneerParser.js
crybx: Removed

plugin/js/parsers/FictioneerParser.js
crybx: Removed this and added line `chapters?.forEach(chapter => chapter.title = chapter.title?.trim());` in onLoadFirstPage in what seemed like the best spot.

plugin/js/parsers/FictioneerParser.js
crybx: Removed this.

PR #2122, opened 2025-08-17
Empty Title

pavan3999:
Title empty fix for chapters and cover, maybe?

### DIFF
--- a/plugin/js/EpubItemSupplier.js
+++ b/plugin/js/EpubItemSupplier.js
@@ -41,15 +41,9 @@ class EpubItemSupplier { // eslint-disable-line no-unused-vars
 
     makeCoverImageXhtmlFile(emptyDocFactory) {
         let doc = emptyDocFactory();
+        doc.title = "Cover";
         let body = doc.getElementsByTagName("body")[0];
         let userPreferences = this.imageCollector.userPreferences;
-    
-        // Set cover page title
-        let title = doc.querySelector("title");
-        if (title && util.isNullOrEmpty(title.textContent)) {
-            title.textContent = "Cover";
-        }
-    
         body.appendChild(this.coverImageInfo.createImageElement(userPreferences));
         return util.xmlToString(doc);
     }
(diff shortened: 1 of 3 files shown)


plugin/js/EpubItemSupplier.js
```diff
@@ -43,6 +43,13 @@ class EpubItemSupplier { // eslint-disable-line no-unused-vars
         let doc = emptyDocFactory();
         let body = doc.getElementsByTagName("body")[0];
         let userPreferences = this.imageCollector.userPreferences;
+    
+        // Set cover page title
+        let title = doc.querySelector("title");
+        if (title && util.isNullOrEmpty(title.textContent)) {
+            title.textContent = "Cover";
+        }
+
```
crybx: doc was just created as an empty page, so title will always be empty. this could just be
```
    makeCoverImageXhtmlFile(emptyDocFactory) {
        let doc = emptyDocFactory();
        doc.title = "Cover";
      
```

crybx: UtestEpubPacker.js needs to be updated for this change as well or unit tests will fail

pavan3999: > doc was just created as an empty page, so title will always be empty. this could just be
> 
> ```
>     makeCoverImageXhtmlFile(emptyDocFactory) {
>         let doc = emptyDocFactory();
>         doc.title = "Cover";
>       
> ```

thanks

plugin/js/Util.js
```diff
@@ -1048,13 +1052,8 @@ const util = (function() {
     }
 
     function sanitize(dirty) {
-        let savedBaseURI = dirty.baseURI;
```
dteviotdale, line 1051: @pavan3999 
Why have you reverted this code?

dteviot:
@pavan3999 

Why are you doing this?  What problem are you trying to solve?
This seems to just be populating the &lt;title&gt; element in the &lt;head&gt; of each XHTML document.  
So far as I am aware, this value is not used for anything in an epub.  So why are you doing this?

dteviot:
@pavan3999 
There has been no response for a month.  I'm closing.

PR #2131, opened 2025-08-21
Consolidate UI text

crybx:
Why do this?

- All the chrome.i18n.getMessage calls in a one place make it easier to find and update messages (IDE renaming can auto find all uses of a property, in js files at least)
- Less likely to typo any given string/key (more benefit if the same text is used multiple times)
- Auto-complete works elsewhere in the code with the text as properties
- Code is slightly more readable
- My personal versions of WebToEpub have a lot more UI features and the increased amount of UI text made me make this change in the first place. So doing it here too reduces diffs for me when I rebase on latest changes or break things up into PRs

### DIFF
--- a/plugin/js/UIText.js
+++ b/plugin/js/UIText.js
@@ -1,0 +1,119 @@
+"use strict";
+
+/**
+ * Centralized UI text constants
+ * User-facing text that needs to be localized should be defined here for easier management
+ */
+class UIText { // eslint-disable-line no-unused-vars
+    // Chapter-related text
+    static Chapter = {
+        tooltipChapterDownloading: chrome.i18n.getMessage("__MSG_Tooltip_chapter_downloading__"),
+        tooltipChapterDownloaded: chrome.i18n.getMessage("__MSG_Tooltip_chapter_downloaded__"),
+        tooltipChapterSleeping: chrome.i18n.getMessage("__MSG_Tooltip_chapter_sleeping__"),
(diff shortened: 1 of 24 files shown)


plugin/js/UIText.js
```diff
@@ -0,0 +1,119 @@
...
+
+    // HTTP Client specific messages
+    static HttpClient = {
+        makeFailCanRetryMessage: chrome.i18n.getMessage("httpFetchCanRetry")
+    };
+
+    // Utility method for localizing UI elements
+    static localizeElement(element) {
+        let localized = chrome.i18n.getMessage(element.textContent.trim());
+        if (localized && localized !== element.textContent.trim()) {
```
dteviot: This isn't the same logic.  was 
```javascript
!util.isNullOrEmpty(localized)
```
Why did you change it?

crybx: I was trying to eliminate UI flickering by preventing unnecessary re-rendering, which is why `&& localized !== element.textContent.trim()` was added (only update if text differs) and `element.textContent` was used over `element.innerText` (which triggers style recalculation).

For the util call...
- `if (localized` - rejects: null, undefined, "" (empty string)
- `if (!util.isNullOrEmpty(localized)` - rejects: null, undefined, "", and whitespace-only strings like "   "

I just checked to figure out the difference and am realizing my change does not reject whitespace-only strings. I don't think that change was intentional, so I added back the util call.

PR #2133, opened 2025-08-24
Update KemonopartyParser.js

Kiradien:
Fixed Kemonoparty paging to get the last page on screen; former pop logic now only gets 'next' button.

This code was tested/pushed from my cellphone; holding back on the merge until the code can be directly tested on PC.

### DIFF
--- a/plugin/js/parsers/KemonopartyParser.js
+++ b/plugin/js/parsers/KemonopartyParser.js
@@ -82,7 +82,7 @@ class KemonopartyParser extends Parser {
 
     async getLastPageOffset(dom, urlbuilder) {
         try {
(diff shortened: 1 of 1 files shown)


Kiradien:
My home Internet connection is effectively FUBAR, it may be a few days before I can properly test this on PC myself. 

If anyone else wants to test it on PC, feel free, else this will probably be here until Wednesday

dteviot:
@Kiradien 

It "works" but probably not the way you intended it to work.
You line means **link** is set to an array of integers.
Next line (which you didn't change) throws an exception because it thinks **link** is a hyperlink.
The exception is caught, and the backup logic runs.

dteviot:
I

PR #2136, opened 2025-08-25
Allow user to disable some warnings

crybx:
These have all bugged me at some point.

The non-english text was AI generated, so I can only hope it's okay.


dteviot:
I feel a **little** uncomfortable with ChapterUrlsUI and EpubPacker reaching into main to fetch the preferences.
In most cases, the preferences would be explicitly passed in.  Either in the constructor, (see EpubPacker getting Epub2 or Epub3) Or as part of a call shortly after.
However, I'll let it go.

PR #2141, opened 2025-08-26
Small fixes for some parsers

crybx:
Small fixes for some parsers

### DIFF
--- a/plugin/js/parsers/FictioneerParser.js
+++ b/plugin/js/parsers/FictioneerParser.js
@@ -96,7 +96,13 @@ class FictioneerParser extends Parser {
         let img =
             dom.querySelector(".wp-post-image") ||
             dom.querySelector("figure.story__thumbnail img");
-        return (img === null) ? img : img.src;
+
+        if (!img?.src) return null;
+
+        // Strip off the arguments for smaller sizes
+        let url = img.src;
+        const pos = url.indexOf("?");
+        return pos !== -1 ? url.substring(0, pos) : url;
     }
 
     customRawDomToContentStep(chapter, content) {
@@ -119,5 +125,10 @@ class FictioneerParser extends Parser {
 
     getInformationEpubItemChildNodes(dom) {
         return [...dom.querySelectorAll(".story__header, .story__summary")];
+    }
+
+    extractSubject(dom) {
+        let tags = ([...dom.querySelectorAll(".story__taxonomies .tag-pill")]);
+        return tags.map(t => t.textContent?.trim()).join(", ");
     }
 }
--- a/plugin/js/parsers/NovelinguaParser.js
+++ b/plugin/js/parsers/NovelinguaParser.js
@@ -19,6 +19,14 @@ class NovelinguaParser extends Parser {
     }
 
     customRawDomToContentStep(chapter, content) {
+        this.cleanup(content);
+    }
+
+    cleanInformationNode(node) {
+        this.cleanup(node);
+    }
+
+    cleanup(content) {
         content.querySelectorAll("*").forEach(element => {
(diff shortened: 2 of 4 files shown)


plugin/js/parsers/FictioneerParser.js
```diff
@@ -96,7 +96,13 @@ class FictioneerParser extends Parser {
         let img =
             dom.querySelector(".wp-post-image") ||
             dom.querySelector("figure.story__thumbnail img");
-        return (img === null) ? img : img.src;
+
+        if (!img || !img.src) return null;
```
dteviot: if (!img?.src) return null;

plugin/js/parsers/FictioneerParser.js
```diff
@@ -120,4 +126,9 @@ class FictioneerParser extends Parser {
     getInformationEpubItemChildNodes(dom) {
         return [...dom.querySelectorAll(".story__header, .story__summary")];
     }
+
+    extractSubject(dom) {
+        let tags = ([...dom.querySelectorAll(".story__taxonomies .tag-pill")]);
+        return tags.map(t => t.textContent?.trim().toLowerCase()).join(", ");
```
dteviot: This doesn't seem right.
If you're using an optional Chaining Operator for textContent, toLowerCase() needs one as well.
And I'm not sure what join will do if one of the elements in the array is null.

That said, I don't think any of the tags with have a textContent of null.  Might be empty string, but that should not be a problem.

crybx: @dteviot With `t.textContent?.trim().toLowerCase()` trim won't be called unless textContent has a string value, even if it's empty, and then toLowerCase would be called. It wouldn't hurt to call toLowerCase on an empty string.

However...it's only my personal preference that tags be lower case. It's probably better to leave them as they originally are on the site, so I removed toLowerCase(). For me, I'll just make all tags lower case in Parser.js in my personal repo.

plugin/js/parsers/FictioneerParser.js
crybx: updated

plugin/js/parsers/FictioneerParser.js
dteviot: I think you misunderstood.

1. Assume t.textContent is null.
2. then t.textContent?.trim() won't execute trim(), but it will evaluate to null.
3. So, we would then evaluate null.toLowerCase().  Which will fail, because null doesn't have a toLowerCase().
4. So, line would need to be "t.textContent?.trim()?.toLowerCase()".  Note the 2 Optional Chaining Operators.
5. However, this would result in a null entry in the array returned by map.  I'm not sure what join does when it encounters a null (Update. I just checked the docs. Nulls are converted into empty strings. So, that should be OK.)

However, as you point out (and I tried to say) I don't think you'll encounter any cases of t.textContent returning a null, so the Optional Chaining Operator is probably unnecessary.
Ultimately, my point was, you need either no Optional Chaining Operators, or 2 of them.
I wasn't worried if the text was lower case or not.

crybx: The optional chaining operator short-circuits the full remaining chain if the value before it is null or undefined. If textContent is null, the expression evaluates to undefined immediately and both trim() and toLowerCase() would never be called. It wouldn't continue executing the next thing in the chain unless you did something like this `(t.textContent?.trim()).toLowerCase()`

But I still think taking toLowerCase() out entirely is better.

dteviot: @crybx 
> The optional chaining operator short-circuits the full remaining chain if the value before it is null or undefined.

I did not know it short-circuits.
Thank you for the learning opportunity

PR #2172, opened 2025-09-12
Added options for image compression

Kiradien:
Turns out webp averages around 30% better compression than jpeg - who knew?
- Added webp and png options for compression (Note, png may result in larger files despite compressed resolution)
- Added 'radio button'/group logic for user preferences

Not entirely sure if this is the path we'd want to go with for the 'grouped' logic for radiobuttons; basing it off the name could trigger issues, but it works in a fundamentally different way. Best alternative is to create a second "addPreference" for grouped items - perhaps "addPreferenceGroup" or fall back on a select box.

### DIFF
--- a/plugin/_locales/en/messages.json
+++ b/plugin/_locales/en/messages.json
@@ -255,8 +255,12 @@
 		"message": "Compress Images",
 		"description": "Label for checkbox and value to compress images to a certain size"
 	},
+	"__MSG_label_Compress_Images_Format__": {
+		"message": "Compressed Image Format",
+		"description": "Label for mime type of compressed image"
+	},
 	"__MSG_label_Compress_Images_Resolution__": {
-		"message": "Resolution",
+		"message": "Compressed Resolution",
 		"description": "Label for number input for maximum dimension size of pixel height or width in image"
 	},
 	"__MSG_label_Cover_from_URL__": {
@@ -554,6 +558,10 @@
 	"__MSG_Searching_For_URLs_Please_Wait__": {
 		"message": "Searching for URLs.  Please wait.",
 		"description": "Message shown while looking for list of Chapter URLs."
+	},
+	"__MSG_option_value_auto__": {
+		"message": "auto",
+		"description": "Generic option text for 'automatic'"
 	},
 	"__MSG_Shift_Click__": {
 		"message": "You can select or unselect a range of chapters by clicking on the checkbox for the first chapter, then  hold down the shift key and click the last chapter in the range.",
--- a/plugin/_locales/en/messages.json
+++ b/plugin/_locales/en/messages.json
@@ -255,6 +255,10 @@
 		"message": "Compress Images",
 		"description": "Label for checkbox and value to compress images to a certain size"
 	},
+	"__MSG_label_Compress_Images_JPG_Cover__": {
+		"message": "Restrict Cover Image to JPEG",
+		"description": "Label for JPEG cover restriction checkbox"
+	},
 	"__MSG_label_Compress_Images_Format__": {
 		"message": "Compressed Image Format",
 		"description": "Label for mime type of compressed image"
--- a/plugin/js/ImageCollector.js
+++ b/plugin/js/ImageCollector.js
@@ -326,6 +326,9 @@ class ImageCollector {
             {
                 let outputType = "image/jpeg";
                 switch (this.userPreferences.compressImagesType.value) {
+                    case "auto":
+                        outputType = util.detectMimeType(imageInfo.getBase64(25));
+                        break;
                     case "webp":
                         outputType = "image/webp";
                         break;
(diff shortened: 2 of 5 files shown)


gamebeaker:
@Kiradien i would remove the choice from the user to choose the filetype.
If someone already compresses i don't think the want to save it lossless with png and if webp is better than jpeg it should be the default.

Kiradien:
My initial thoughts were to do just that, however many epub readers still don't support webp - and might never support webp. 

Due to their age, jpeg and png are entirely safe formats, as they are older and fully adopted before the epub format was created, however webp wasn't released as a format until 3 years later, and wasn't even fully adopted by browsers until 10 years after that - To this day, Operating Systems don't know what to do with this format by default.

The compression does still affect the png - it's just mostly a resolution compression in this case... In my own tests the filesize was still 4MB smaller than uncompressed, but that book is probably an outlier. 

The main reason I am hesitant to just remove png as an option was user preferences noted in #2057 - it would also neatly provide a more complete resolution for that issue.

Maybe I should switch it's naming from image compression to image conversion? Though the codes _main_ purpose is still compression.

gamebeaker:
@Kiradien you are right. I thought with epub3 webp is officially supported. But not all support epub3...

> Maybe I should switch it's naming from image compression to image conversion? Though the codes main purpose is still compression.

I think renaming isn't necessary as pictures only get scaled down not up. (or did i miss the upscalling?)
Is an auto option possible that just keeps the original file type?
My thought is that if an author wants to show an uncompressed png the highest resolution resize is saved with lossless compression. (My personal opinion is that most people don't care what filetype they share online and that they should convert it to jpeg/ webp to reduce it's size for online content.)

I change my statement that webp should be the default to that it should just keep the file type as default.

Kiradien:
Nope, there is no upscaling - made absolutely sure that wasn't included after the earliest versions.

My initial thoughts regarding 'auto' were a big 'no', but then I remembered util.detectMimeType - it actually should be possible to implement, though it will default to png when the mime type fails to save. I'll take a look at implementing 'auto' tonight.

Note: another variant of "auto" would be to match [this](http://dev.kiradien.com/imgConvert.html) page, looping through each and choosing the smallest one. Doing this, unfortunately, completely eats up memory, in an already memory hungry process.

gamebeaker:
@Kiradien 
Maybe just my annoyance with large epubs speaking (I am endless thankful for the compression feature) but why choose png over jpeg?
I personally can't see the difference between them if i look at normal pictures without zooming.

Kiradien:
Personally, unless I revert to an older epub reader that can't handle webp's, I don't see myself swapping back to jpg or png - but the option doesn't hurt.

If I go to an older reader, and I suddenly care about quality, I will either swap to png or turn off compression all together.

Admittedly, there are actually rare cases where a png is smaller than webp or jpg, but it is so rare that it would be overwhelmed by the savings from other images in the novel. There are still cases where jpg beats webp as well, but webp averages to around 30% smaller at 90% quality.

Besides this, png is one of the few formats (gif, webp, png) that fully supports both transparency and animation - though the main advantage is 'lossless' compression for those who actually care; this is what issue 2057 was actually requesting on top of 'webp doesn't work for me'. 

2057 was actually the main reason I remembered to not eliminate the jpg option, my initial concept was just a 'webp' checkbox but radio buttons fit the role better; I didn't change it at the time since, in all honesty, I wasn't sure the best way to handle the UI for it and wasn't entirely sure how toBlob handled pngs at that time.

Like I said, I won't use it myself - but if we're giving a choice of format in the first place, why not let those who want larger have larger?
Personally, I'm happy with my 150MB epub shrunk down to 12MB (It was 18MB with jpegs!)

Edit: I think, while including auto as an option, I will actually keep jpg as the default image type when compression is turned on. It is actually the safest.

Kiradien:
Huh... Okay, so I need to thank you, gamebreaker, for suggesting the auto mode - while testing, it pointed out an edgecase issue with detectMimeType on files overloaded with header details on pages without a distinct extension. While it wouldn't cause issues elsewhere in the code, some epub readers might still struggle loading images - something that code was meant to eliminate. It wasn't a new issue from the code, just something that was meant to be fixed already.

On top of being rare, It's something that we'd never notice so long as compression was turned on before.

dteviot:
@Kiradien 

FYI. Rather than use radio buttons, my preference is a drop down list.  (Although I'm sure any UI guru would strongly disagree with me.

Kiradien:
Considering the current UI, I don't know how I missed that option... 

I'll give that a go, though the radio buttons might be more intuitive.
image
vs
image

Admittedly, the UI took work to make it work to fit everything nicely with the radio buttons, but it feels intuitive like that, I think? I'll save both versions of the source code; not sure which version to stick with atm.

Edit: Eh, I probably have the resolution input & label backwards.

Edit 2 the next morning: Wow, sleep deprived judgements are bad. Yeah, I'll polish up the drop down version & submit that.

Kiradien:
Alright, so now that I made the change while properly awake, I've changed the indentation a bit to kinda group the 'compressed image options' together and it works better.

image

I'll merge it in after some more devtesting.

PR #2189, opened 2025-09-17
Add parser for shitouxs.com

Yomafil:
Added a new parser for the site shitouxs.com

The description is under ".bookintro" but I couldn't find a way to have it packed into the info of the EPUB, so I didn't include it.

### DIFF
--- a/plugin/js/parsers/ShitouxsParser.js
+++ b/plugin/js/parsers/ShitouxsParser.js
@@ -34,6 +34,6 @@ class ShitouxsParser extends Parser {
     }
 
     getInformationEpubItemChildNodes(dom) {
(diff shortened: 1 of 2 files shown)


Kiradien:
The description is pulled in the "extractDescription(dom)" method - the following would probably work - untested though.

```

extractDescription(dom) {
    return dom.querySelector(".bookintro").textContent.trim();
}

```

dteviot:
@Yomafil 
You should not need to provide a extractDescription.  If you provide getInformationEpubItemChildNodes(), the default logic will use that to obtain the description.  If that's not happening, then something is broken and an issue should be raised to fix it.

dteviot:
@Yomafil Thank you for your hard work

PR #2193, opened 2025-09-18
update ChrysanthemumgardenParser

gamebeaker:
#2173

This adds the fonts in the epub but doesn't link them in stylesheet.css for that this PR in EpubEditor is needed:
https://github.com/dteviot/EpubEditor/pull/12

### DIFF
--- a/plugin/js/EpubItem.js
+++ b/plugin/js/EpubItem.js
@@ -320,3 +320,15 @@ class ImageInfo extends EpubItem { // eslint-disable-line no-unused-vars
         // images do not appear in table of contents
     }
 }
+
+class FontInfo extends ImageInfo {
+    constructor(fontName) {
+        super();
+        this.fontName = fontName;
+    }
+
+    packInEpub(zipWriter) {
+        zipWriter.add("OEBPS/Fonts/"+this.fontName,
+            new zip.BlobReader(new Blob([this.arraybuffer])));
+    }
+}
--- a/plugin/js/parsers/ChrysanthemumgardenParser.js
+++ b/plugin/js/parsers/ChrysanthemumgardenParser.js
@@ -157,15 +157,3 @@ class ChrysanthemumgardenParser extends WordpressBaseParser {
         }
     }
 }
-
-class FontInfo extends ImageInfo {
-    constructor(fontName) {
-        super();
-        this.fontName = fontName;
-    }
-
-    packInEpub(zipWriter) {
-        zipWriter.add("OEBPS/Fonts/"+this.fontName,
-            new zip.BlobReader(new Blob([this.arraybuffer])));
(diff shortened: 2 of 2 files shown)


dteviot:
@gamebeaker 

I'm a little uncomfortable with this.
1. Possible security issues with fonts.  I don't want Mozilla to request we sanitize the downloaded fonts.
2. Only Epub v3 supports embedded fonts.  (And not all readers.)
3. It's also getting very close to obviously trying to circumvent the site's "anti-copying".
4. I think perhaps we could offer an "include fonts" option with WebToEpub 
5. I would prefer this being on it's own branch, and we can provide a link to the pre-built version for those that want it.  (Until the "Include fonts" option is a bit less obvious what it's for.)

Technical notes:
1. I don't think you needed to override fetchWebPageContent(),  Overriding fetchChapter() should be sufficient.
2. I think you're also not fully constructing the FontInfo.  At minimum, mediaType, sourceUrl, and dataOrigFileUrl and will be wrong in the epub's manifest.

dteviot:
@gamebeaker 

Also, thank you very much for trying this.  It might be the way to go, long term.

gamebeaker:
> I think perhaps we could offer an "include fonts" option with WebToEpub

I don't think this will be possible. Why?
The standard why to set fonts is in my opinion with classes or id. Only the obfuscation is at the moment inline.
The only reason that i added it in WebToEpub and not in EpubEditor is that it isn't possible to download the fonts because cors blocks it.

> I don't think you needed to override fetchWebPageContent(), Overriding fetchChapter() should be sufficient.

I only wanted to analyze the chapter content and what fonts there got used not the whole page that's why i didn't do it in fetchChapter()

> I think you're also not fully constructing the FontInfo. At minimum, mediaType, sourceUrl, and dataOrigFileUrl and will be wrong in the epub's manifest.
I think you're also not fully constructing the FontInfo. At minimum, mediaType, sourceUrl, and dataOrigFileUrl and will be wrong in the epub's manifest.

I just hacked something together and needed a way to add "random" files to the epub that aren't chapters. That's why i decided to abuse the image system. It is more like a proof of concept that it is possible.


I could imagine a new function in parser.js like async getFonts() that returns an array/map of fonts that would need to be implemented in the specific parsers.

gamebeaker:
> Possible security issues with fonts. I don't want Mozilla to request we sanitize the downloaded fonts.

I don't think this will be an issue as they don't require that we sanitize images. They only had issues with WebToEpub as we used DOMParser and this could be a problem if it is displayed in the browser. After this missunderstanding was cleared (it only gets downloaded not displayed) it was ok. (at least that was my take away.)

dteviot:
OK.  Let's give it a go.

PR #2197, opened 2025-09-19
Tweak MIME signature detection

Kiradien:
webp signature is weird - can fail atob transition. Google uses 'UklGRg' but this also triggers issues. Swapped to try-catch for now. Will do further testing on alt signature for future updates.


Kiradien:
Just to be clear, the above commit works, it solves the issues, but it is **not** a permanent solution. 

I have a potential permanent solution locally - a mix of Google's and the old solution - but it will take several weeks to test if I want to avoid getting IP Banned. The issue is triggered by webp saving every single bit in every byte in its headers.

PR #2225, opened 2025-10-01
Add ciweimao

kuwoyuki:
chapter content goes into `div.encrypted-content`.
```js
const encryptedDiv = newDoc.dom.createElement("div");
encryptedDiv.className = "encrypted-chapter-content";
encryptedDiv.dataset.encryptedContent = json.chapter_content;
encryptedDiv.dataset.accessKey = json.chapter_access_key;
encryptedDiv.dataset.encryptionKeys = JSON.stringify(
    json.encryt_keys
);
```

EpubEditor post-processor once/if https://github.com/dteviot/EpubEditor/pull/14 is merged:

```js
let chChanged = false;
const encryptedDiv = dom.querySelector("div.encrypted-chapter-content");

if (encryptedDiv) {
    const {
        encryptedContent,
        accessKey,
        encryptionKeys: keysJson,
    } = encryptedDiv.dataset;

    if (encryptedContent && accessKey && keysJson) {
        const base64ToArrayBuf = (base64) => {
            const binaryString = atob(base64);
            const bytes = new Uint8Array(binaryString.length);
            return bytes.map((_, i) => binaryString.charCodeAt(i));
        };

        try {
            const keys = JSON.parse(keysJson);
            const keysLength = keys.length;

            const keysToUse = [
                keys[accessKey.charCodeAt(accessKey.length - 1) % keysLength],
                keys[accessKey.charCodeAt(0) % keysLength],
            ];

            let currentContentB64 = encryptedContent;
            let finalDecryptedBuf;

            for (const keyB64 of keysToUse) {
                const keyData = base64ToArrayBuf(keyB64);
                const cryptoKey = await crypto.subtle.importKey(
                    "raw",
                    keyData,
                    { name: "AES-CBC" },
                    false,
                    ["decrypt"]
                );

                const rawContentBuf =
                    base64ToArrayBuf(currentContentB64).buffer;
                const iv = rawContentBuf.slice(0, 16);
                const ciphertext = rawContentBuf.slice(16);

                const decryptedBuf = await crypto.subtle.decrypt(
                    { name: "AES-CBC", iv },
                    cryptoKey,
                    ciphertext
                );

                currentContentB64 = new TextDecoder("latin1").decode(
                    decryptedBuf
                );
                finalDecryptedBuf = decryptedBuf;
            }

            const decryptedHtml = new TextDecoder("utf-8").decode(
                finalDecryptedBuf
            );
            encryptedDiv.innerHTML = decryptedHtml;

            encryptedDiv.classList.remove("encrypted-chapter-content");
            delete encryptedDiv.dataset.encryptedContent;
            delete encryptedDiv.dataset.accessKey;
            delete encryptedDiv.dataset.encryptionKeys;

            chChanged = true;
        } catch (error) {
            console.error("failed to decrypt: ", error);
        }
    } else {
        console.error("missing decryption attrs");
    }
}

const spansToRm = dom.querySelectorAll("span");

if (spansToRm.length > 0) {
    spansToRm.forEach((span) => span.remove());
    chChanged = true;
}

return chChanged;
```


dteviot:
@kuwoyuki 

I'm disappointed you decided to withdraw this.
But thank you very much for your efforts.

kuwoyuki:
> @kuwoyuki
> 
> I'm disappointed you decided to withdraw this. But thank you very much for your efforts.

@dteviot I can reopen a PR with free chapter download and text decryption in post-processing only, but that's a very partial support and complex to use.

As mentioned in the issue, maybe it's worth looking into ways to get full parsers available in the extension? Plugin system, repositories, script import (by URL for example) etc. without breaking web store T&C? Similar to how userscript extensions work for example.

gamebeaker:
> As mentioned in the issue, maybe it's worth looking into ways to get full parsers available in the extension? Plugin system, repositories, script import (by URL for example) etc. without breaking web store T&C? Similar to how userscript extensions work for example.

It is against tos: https://developer.chrome.com/docs/webstore/program-policies/mv3-requirements

At least i interpret point 1 like this.

If someone wants to use disabled parser he needs to use the test version and edit the source files. We could maybe provide a "unlocked" WebToEpub version in GitHub.
I am against it as some website owners have requested to disable their parser and if someone still wants to download from them "they" can at least put the work in to delete 4 lines of code.

kuwoyuki:
> > As mentioned in the issue, maybe it's worth looking into ways to get full parsers available in the extension? Plugin system, repositories, script import (by URL for example) etc. without breaking web store T&C? Similar to how userscript extensions work for example.
> 
> It is against tos: https://developer.chrome.com/docs/webstore/program-policies/mv3-requirements
> 
> At least i interpret point 1 like this.
> 
> If someone wants to use disabled parser he needs to use the test version and edit the source files. We could maybe provide a "unlocked" WebToEpub version in GitHub. I am against it as some website owners have requested to disable their parser and if someone still wants to download from them "they" can at least put the work in to delete 4 lines of code.

That depends on how it's implemented, from point 1 it would be compliant, as long as the remote configuration contains only data and not logic.
If a parser remote configuration file contains query selectors and encryption information, both of those are data and not logic. I.e. `document.querySelector(...)` impl. The remote string, like `".chapter-content"`, is just data being passed as an argument to that pre-existing, locally-packaged logic.
Same with say `crypto.subtle.encrypt`. Fetching a public key, a salt, or an Initialization Vector (IV) from a server is fetching data that this algorithm will use..

The interpreter point is kind of gray area however, since that's essentially a sequence of commands, so yeah probably some parsers could be trickier to implement - to fully contain the logic within the extension's submitted code

PR #2263, opened 2025-10-17
Add MimihuiParser

Yomafil:
Add new parser for https://www.mimihui.com/

Tested with both comics and novels.
Untested with VIP chapters unlocked.


dteviot:
@Yomafil Thank you for your hard work.

PR #2273, opened 2025-10-22
Add YedugeParser

Yomafil:
Add new parser for www.yeduge.com


gamebeaker:
@Yomafil thanks for the help.

PR #2291, opened 2025-11-07
Fix NCX Id and OPF Id Descrepancy for Epub 3 Generation

pjkaufman:
When creating an epub version 3, we prefix the epub's OPF id with `uri:` as per the logic [here](https://github.com/dteviot/WebToEpub/blob/2678b0e8c623e67a8cf119d90ee501570325a5a8/plugin/js/EpubPacker.js#L112-L114). However we don't currently do that for the NCX file. I am not sure if this is intentionally left out, but the EPUBCheck validator throws the following error for me locally when the file is generated:
``` bash
Validating using EPUB version 3.3 rules.
ERROR(NCX-001): ./Gosick.epub/OEBPS/toc.ncx(-1,-1): NCX identifier ("https://lightnovelstranslations.com/novel/gosick/?tab=table_contents&status=complete") does not match OPF identifier ("uri:https://lightnovelstranslations.com/novel/gosick/?tab=table_contents&status=complete").
ERROR(RSC-005): ./Gosick.epub/OEBPS/Text/Cover.xhtml(1,110): Error while parsing file: Element "title" must not be empty.

Check finished with errors
Messages: 0 fatals / 2 errors / 0 warnings / 0 infos

EPUBCheck completed
```

I can include the generated epub if that is helpful as well. Thanks for making this extension!

_Note that the second issue is a separate issue that seems to be specific to epub 3 as well, so I will ignore that for this PR._

### DIFF
--- a/package.json
+++ b/package.json
@@ -92,7 +92,8 @@
     { "name": "basitcodeenv"},
     { "name": "X-Xadro"},
     { "name": "Yomafil"},
-    { "name": "Varun Patkar", "email": "varunpatkar501@gmail.com" }
+    { "name": "Varun Patkar", "email": "varunpatkar501@gmail.com" },
+    { "name": "Peter Kaufman", "email": "PeterJamesKaufman@gmail.com" }
   ],
   "license": "GPL-3.0-only",
   "bugs": {
--- a/plugin/js/EpubPacker.js
+++ b/plugin/js/EpubPacker.js
@@ -274,7 +274,7 @@ class EpubPacker {
     }
 
     populateHead(head, ns, depth) {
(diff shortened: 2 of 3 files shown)


gamebeaker:
@pjkaufman Thanks for the PR.

please add yourself as a contributer here is more info about this: https://github.com/dteviot/WebToEpub/blob/ExperimentalTabMode/CONTRIBUTING.md

pjkaufman:
Thanks for the quick response @gamebeaker . I should now have that information added to the PR.

gamebeaker:
@pjkaufman thank you for the PR.
Tested with chapter 1 from https://www.royalroad.com/fiction/138598/unceasing-slumbering

PR #2293, opened 2025-11-08
Set Cover Page Title

pjkaufman:
There is an EPUBCheck error that gets thrown when generating an epub 3 using Web to Epub. The error states:
```
Validating using EPUB version 3.3 rules.
ERROR(RSC-005): ./Gosick.epub/OEBPS/Text/Cover.xhtml(1,110): Error while parsing file: Element "title" must not be empty.

Check finished with errors
Messages: 0 fatals / 1 errors / 0 warnings / 0 infos

EPUBCheck completed
```

To fix this issue, we just need to set a title. I at first thought about hardcoding the value to `Cover`, but that would possibly need localization. So I instead implemented the logic using the user input title for the book. It is possible to use a hardcoded value if that is what works best for the project.

### DIFF
--- a/plugin/js/EpubItemSupplier.js
+++ b/plugin/js/EpubItemSupplier.js
@@ -39,11 +39,16 @@ class EpubItemSupplier { // eslint-disable-line no-unused-vars
         }
     }
 
-    makeCoverImageXhtmlFile(emptyDocFactory) {
+    makeCoverImageXhtmlFile(emptyDocFactory, title) {
         let doc = emptyDocFactory();
         let body = doc.getElementsByTagName("body")[0];
         let userPreferences = this.imageCollector.userPreferences;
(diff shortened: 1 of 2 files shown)


gamebeaker:
@pjkaufman i would use a hard coded "Cover" instead of the title of the epub as it is more descriptive of the Chapter than the title of the whole epub.
I had a look at the epub2 spec and there the title is optional so we don't have to make a difference between epub3 and epub2.

Are you going to make the change or should i do it?

pjkaufman:
@gamebeaker , I can hardcode the value as `Cover`.

pjkaufman:
The change should be present now.

gamebeaker:
@pjkaufman thanks for your PRs.

pjkaufman:
No problem. I am glad to help with tools I use where I have the ability to do so.

PR #2298, opened 2025-11-09
Add `NovelsOnline.org` to NovelsOnline Parser and Extract Author

pjkaufman:
I am not sure if the site is a clone or just a rebrand of the other one, but `novelsonline.org` seems to have the exact same structure as what was expected for `novelsonline.net`. So I just added it as an option for download. It seems to have identified the data as expected.

I also went ahead and included the extraction of the author. I was not sure if we could guarantee that the author would always be at a specific child index in the html, so I instead opted to iterate over the possible elements and find it that way. It should work relatively quickly, but if we want to hard code the child element instead, that could be done too.


dteviot:
@pjkaufman 

PR #2313, opened 2025-11-16
novelshub.org fetcher

nothing0074:
note that AI was used in creating this fetcher
tested on https://novelshub.org/series/academy-retirement-diary-of-the-student-council-president and worked fine, unknown if premium chapters will work
good chance that chapter titles won't work well on any other novel other than the one I linked
chapter counts will also probably not function correctly on a lot of novels (not really a problem since it just stops fetching after last valid chapter)
haven't tested but this fetcher would probably not be able to fetch more than 1000 chapters at once


dteviot:
@nothing0074 

Yes, that AI generated code is horrible.  I've tried to clean it up a bit. But it's still not good.
See: https://github.com/dteviot/WebToEpub/commit/9fc6e9943560989b477a04c1eb51c856df30b5ac


Tested with:
* https://novelshub.org/series/an-idol-who-died-right-before-his-debut-novel, chapters 1, 2

Notes
Time taken:  71 minutes.  (Far to long, site lacks decent CSS tags.)

PR #2335, opened 2025-11-29
WanderinginnParser fix, colors, and mrsha-italics

### LINKED ISSUE #2330: Error when scraping from wanderinginn.com
**Describe the bug**
Unable to scrape wandering inn chapters

**To Reproduce**
1. Go to table of contents
2. Click on web2epub
3. Select chapters
4. click pack to epub

**Expected behavior**
Expected an epub file containing the chapters

**Screenshots**

Image

**Desktop (please complete the following information):**
 - OS: windows
 - Browser: Edge
 - Version 142.0.3595.94

johnpyp:
Fixes https://github.com/dteviot/WebToEpub/issues/2330

- Fix `wanderinginn.com` parsing:
	-  new excluded toc class
	- `#reader-content` is used as the main content container, rather than `.entry-content`
	- `h2.elementor-heading-title` is now used for chapter title
- Add `mrsha-write` italics, to emulate the custom font used in the web
reader (which used to use italics, before switching to the custom font)
- Add ibooks dark-mode-allowed class to wherever inline color styles are
used
	- This makes ibooks display the colors used, while reading in dark mode. TWI colors are made for dark mode, so this works great!

---

I tested on some chapters of the book and confirmed that both the class additions and mrsha-write italics are working as expected, and work properly in iBooks to show both :)

### DIFF
--- a/package.json
+++ b/package.json
@@ -97,7 +97,8 @@
     { "name": "patiorjunrick"},
     { "name": "MD Shabrez" },
     { "name": "jurassicplayer" },
-    { "name": "X2E4VXpZKv" }
+    { "name": "X2E4VXpZKv" },
+    { "name": "johnpyp", "email": "johnpyp.dev@gmail.com" }
   ],
   "license": "GPL-3.0-only",
   "bugs": {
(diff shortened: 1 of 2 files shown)


gamebeaker:
@johnpyp thank you for the PR.

Krynh:
mrsha-write is just the charm font

johnpyp:
Right, I made it italics because I didn't know of a good way to export the actual font into an epub?

(And it doesn't get exported into my viewer, at least)

Lmk if you think there's a better approach

PR #2363, opened 2025-12-13
Add parser for ficbook.net

AgatZan:
I add new site from title.

Further work:
1. Split tags for categories.


unitTest/UtestFicbookParser.js
gamebeaker, line 1: I would remove the whole test file.
Should the parser no longer work someone will open a new issue. If no one opens a issue an update isn't needed as no one uses the site.

AgatZan: @gamebeaker no problem

plugin/js/parsers/FicbookParser.js
```diff
@@ -33,9 +33,12 @@ class FicbookParser extends Parser {
     findContent(dom) {
         return dom.querySelector("#content");
     }
+    customRawDomToContentStep(chapter, content) {
+        content.innerHTML = content.innerText.split("\n\n").map(x=>"<p>" + x + "</p>").join("");
```
dteviot: Assigning unsanitized html to an innerHTML will get WebToEpub rejected from the Firefox store.

You'd need to do something like
```javascript
let raw = content.innerText.split("\n\n").map(x=>"<p>" + x + "</p>").join("");
let sanitilzed = let content = util.sanitize(raw);
content.innerHTML = sanitized.body.innerHTML.
```

Unfortunately, the sanitizer that Mozilla requires us to use is very aggressive in removing attributes and elements.  So, if you're going to sanitize here, double check function to find the chapter title still works.  (Otherwise you'll need to remove the title before sanitizing, then put it back afterwards.)

AgatZan: As I can think, it is fixed.

dteviot:
@AgatZan 

Thank you for your hard work.

Alla183:
> @AgatZan
> 
> Спасибо за вашу усердную работу.

To ensure correct operation, do I need to reinstall the plugin?

AgatZan:
@Alla183 Maybe it will help. If you are having a plugin locally, you need to download [latest developer build](https://github.com/dteviot/WebToEpub/releases/tag/developer-build), then unpack and install it. If you install it through an extension store, the package will be automatically updated later.

Ringotser:
For some reason, some books do not retain the formatting of the original text, there are no paragraphs, etc.

dteviot:
@Ringotser 

The original site has raw text, formatted with spaces.  WebToEpub needs to turn it into XHTML.  The conversion is, shall we say, somewhat imperfect.  Although you might be able to improve things by tweaking the stylesheet in the epub with something like Calibre.

PR #2382, opened 2025-12-24
Fix Novel543Parser: correctly combine multi-page chapters

s4daharu:
Title: Fix Novel543Parser: correctly combine multi-page chapters

Description:

Novel543.com splits long chapters into multiple pages (e.g., 8096_1.html → 8096_1_2.html). The previous 
moreChapterTextUrl
 logic only checked for exactly 2 underscores, which didn't correctly identify continuation pages.

This fix:

Extracts the chapter base from the original URL (e.g., 8096_1 from 8096_1.html)
Checks if the next link URL contains the base followed by an underscore (e.g., 8096_1_2.html)
Stops fetching when a new chapter is reached (e.g., 8096_2.html)
Tested: Verified that multi-page chapters are now correctly combined into single EPUB chapters.

### DIFF
--- a/plugin/js/parsers/Novel543Parser.js
+++ b/plugin/js/parsers/Novel543Parser.js
@@ -42,12 +42,27 @@ class Novel543Parser extends Parser {
         return this.walkPagesOfChapter(url, this.moreChapterTextUrl);
     }
 
-    moreChapterTextUrl(dom) {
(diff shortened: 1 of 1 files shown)


dteviot:
@s4daharu 

Thank you for your hard work

PR #2410, opened 2026-01-16
[Feature Upgrade] Added handling of arc titles in Syosetu

ImmortalDreamer:
-Updated extractPartialChapterList to associate arc titles with chapters when parsing Syosetu's chapter list.
This improves support for stories with arcs by correctly grouping chapters under their respective arc titles.

-Added fallback logic when chapter list doesn't exist (Happens when process is not initiated on series page).
Instead of throwing an error, parser returns empty chapter list.

Tested and working on Syosetu and R18 Syosetu.

### DIFF
--- a/plugin/js/parsers/SyosetuParser.js
+++ b/plugin/js/parsers/SyosetuParser.js
@@ -32,7 +32,26 @@ class SyosetuParser extends Parser {
 
     extractPartialChapterList(dom) {
         let chapterList = dom.querySelector("div.index_box") || dom.querySelector("div.p-eplist");
-        return [...chapterList.querySelectorAll("a")].map(a => util.hyperLinkToChapter(a));
+        if (!chapterList) {
+            return [];
+        }
+        let chapters = [];
+        let arcTitle = null;
+        for (let element of chapterList.querySelectorAll(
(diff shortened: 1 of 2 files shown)


dteviot:
@ImmortalDreamer 
For future reference, You should probably disable the version bump for commits to your branch. 
Otherwise we're going to have merge conflicts in the future.

ImmortalDreamer:
@dteviot 
No problem.
Will disable the version bump before my next commit.
I will be looking at a few other issues since I have more time now and try to implement fixes. 
Would a single pull request with multiple fixes be better? Or do you prefer that I send them individually?

dteviot:
@ImmortalDreamer 

If they're independent, submit individually.  Easier for me to check.

ImmortalDreamer:
@dteviot 
Sounds good.

gamebeaker:
@ImmortalDreamer if you use different branches like it is shown here: https://github.com/dteviot/WebToEpub/wiki/Developer-guide the auto version bump doesn't happen.

ImmortalDreamer:
@gamebeaker 
Thanks for the link. I hadn't been on here in a long while, so I forgot about it and didn't see it until after the merge.
Will use different branches from now on. 👍

Issue #2406, opened 2026-01-13
Please add site https://lorenovels.com/


LKADity:
Based on What I read from other issue, it seems like the initial web page does not have the text and use JS to call the text later because I was able to get the content via body but not when directly refer the text element

Provide URL for web page that contains Table of Contents (list of chapters) of a typical story on the site:
https://lorenovels.com/the-support-ate-it-all/

The Setting I used : 
* URL of first chapter = https://lorenovels.com/chapter-1-reincarnation-quest-part-1/
* CSS selector for element holding content to put into EPUB = #wp--skip-link--target > div.entry-content.wp-block-post-content.has-global-padding.is-layout-constrained.wp-block-post-content-is-layout-constrained.dracula-style-txt-border.dracula-processed
* CSS selector for element holding Title of Chapter = #wp--skip-link--target > div.wp-block-group.has-global-padding.is-layout-constrained.wp-container-core-group-is-layout-334a2726.wp-block-group-is-layout-constrained.dracula-style-txt-border.dracula-processed > h1
* CSS selector for element(s) to remove = #wp--skip-link--target > div.entry-content.wp-block-post-content.has-global-padding.is-layout-constrained.wp-block-post-content-is-layout-constrained.dracula-style-txt-border.dracula-processed > div

ImmortalDreamer:
@LKADity 
I've created a parser for this.
Please wait until it's merged and available.

dteviot:
@LKADity 

Has been merged.

Test versions for Firefox and Chrome have been uploaded to https://github.com/dteviot/WebToEpub/releases/tag/developer-build. Pick the one suitable for you, follow the "How to install from Source (for people who are not developers)" instructions at https://github.com/dteviot/WebToEpub/tree/ExperimentalTabMode#user-content-installation and let me know how it goes.
Tested with:
* https://lorenovels.com/why-are-all-the-fairies-so-obsessed-with-my-downfall/, chapters 1 and 2

LKADity:
@ImmortalDreamer @dteviot 
It's works perfectly like a charm
Thank you for the update

dteviot:
@LKADity 
Reopen, so I know to notify you when Chrome and Firefox stores are updated.

dteviot:
@LKADity @ImmortalDreamer 

Updated version (1.0.11.0) has been submitted to Firefox and Chrome stores.
Firefox version should be available within two weeks (but might take 4 to 6 weeks)
[...]

PR #2411, opened 2026-01-17
Added LoreNovelsParser.

ImmortalDreamer:
-Added parser for Lorenovels, addressing #2406.

Tested on https://lorenovels.com/the-support-ate-it-all/ and confirmed to be working.

### DIFF
--- a/plugin/js/parsers/LoreNovelsParser.js
+++ b/plugin/js/parsers/LoreNovelsParser.js
@@ -1,0 +1,62 @@
+"use strict";
+
+parserFactory.register("lorenovels.com", () => new LoreNovelsParser());
+
+class LoreNovelsParser extends Parser {
+    constructor() {
+        super();
+    }
+
+    getChapterUrls(dom) {
+        const toc = dom.querySelector("ul.wp-block-latest-posts__list");
+        if (!toc) {
+            return Promise.resolve([]);
+        }
+
+        const items = [...toc.querySelectorAll(
+            "a.wp-block-latest-posts__post-title"
+        )];
+
+        return Promise.resolve(this.buildChapterList(items));
+    }
+
+    buildChapterList(items) {
+        return items
+            .reverse() // newest → oldest → reading order
+            .map(a => ({
+                sourceUrl: a.href,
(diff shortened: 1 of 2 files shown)


plugin/js/parsers/LoreNovelsParser.js
```diff
@@ -0,0 +1,62 @@
+"use strict";
+
+parserFactory.register("lorenovels.com", () => new LoreNovelsParser());
+
+class LoreNovelsParser extends Parser {
+    constructor() {
+        super();
+    }
+
+    getChapterUrls(dom) {
```
dteviot: FYI.

Make this async, and it will automatically return a Promise.

plugin/js/parsers/LoreNovelsParser.js
```diff
@@ -0,0 +1,62 @@
...
+    buildChapterList(items) {
+        return items
+            .reverse() // newest → oldest → reading order
+            .map(a => ({
+                sourceUrl: a.href,
+                title: a.textContent.trim()
+            }));
+    }
+
+    extractContent(dom) {
```
dteviot: I think this extractContent is never called.  It's not a member of the Parser class.

dteviot (commented):
I'll test this when I get home tonight and clean up any issues.

plugin/js/parsers/LoreNovelsParser.js
ImmortalDreamer: Sorry about that. I reused code from an older parser. Will make it async from now on. Already made a commit to fix it.

plugin/js/parsers/LoreNovelsParser.js
ImmortalDreamer: Sorry about that. It was my initial findContent, but got renamed to extractContent at some point and since findContent, findChapterTitle and removeUnwantedElementsFromContentElement were also in the code and got called, I didn't realize it was not getting called at all. Will make sure to double check the code before making a PR from next time. Already made a commit removing it.

dteviot:
@ImmortalDreamer 

Thank you for your hard work

Issue #2407, opened 2026-01-14
Skipping texts (Knoxt)


AishaPB:
Hi! I encountered this issue after creating an epub from knoxt. This is the specific  novel -> https://knoxt.space/qing-gu/

Somehow, every chapter is missing some texts in the beginning. 

This whole text was skipped for chapter 1:

![Image](https://github.com/user-attachments/assets/18da8cc3-90ec-42b5-a56b-c4d8f3c9e5a8)

and my chapter starts with this:

![Image](https://github.com/user-attachments/assets/de8e0fc7-cfc7-478a-bcb7-5027199f1ba2)

I was getting the same issue with Firefox and chrome.

AishaPB:
I think it's skipping the texts in between the "advertisement" at the beginning of each chapters. Is there a way to fix this? Thank you!

ImmortalDreamer:
@AishaPB 

You are right. The parser removes blocks containing the ads, but the first block contains the paragraphs in between the first and second ad, hence removing them.

I've begun working on this, so please wait for an update.

dteviot:
@AishaPB 

Test versions for Firefox and Chrome have been uploaded to https://github.com/dteviot/WebToEpub/releases/tag/developer-build. Pick the one suitable for you, follow the "How to install from Source (for people who are not developers)" instructions at https://github.com/dteviot/WebToEpub/tree/ExperimentalTabMode#user-content-installation and let me know how it goes.
Tested with:
* https://knoxt.space/the-captain-who-specializes-in-treating-disobedience/, chapter 1

Notes
Time taken: 22 minutes

AishaPB:
@dteviot @ImmortalDreamer This fixed the issue. Thank you so much!

dteviot:
@AishaPB 

Updated version (1.0.11.0) has been submitted to Firefox and Chrome stores.
Firefox version should be available within two weeks (but might take 4 to 6 weeks)
Chrome might be available in a few hours to 21 days.  (Typically 3 to 7 days).

PR #2413, opened 2026-01-17
Fixed Knoxtspace parser

ImmortalDreamer:
Addresses #2407, #2414 and other improvements.
-Adds preprocessing to prevent removal of chapter text from first ad block.
-Removes ad blocks from information page.
-Removes empty paragraphs from chapters to enhance readability.

### DIFF
--- a/plugin/js/parsers/NoblemtlParser.js
+++ b/plugin/js/parsers/NoblemtlParser.js
@@ -149,6 +149,8 @@ class KnoxtspaceParser extends NoblemtlParser {
             }
             parent.removeChild(block);
         }
+        util.removeElements([...webPageDom.querySelectorAll("p")]
+            .filter(p => !p.textContent.trim()));
     }
 
     cleanInformationNode(node) {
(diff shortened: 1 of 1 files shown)


plugin/js/parsers/NoblemtlParser.js
```diff
@@ -139,6 +139,24 @@ class KnoxtspaceParser extends NoblemtlParser {
...
+        for (let block of webPageDom.querySelectorAll("div.code-block")) {
+            let parent = block.parentNode;
+            while (block.firstChild) {
+                parent.insertBefore(block.firstChild, block);
+            }
+            parent.removeChild(block);
+        }
+    }
+
+    cleanInformationNode(node) {
```
dteviot: This doesn't do what you think it does.
cleanInformationNode() is used for cleaning up the "information" page only.  It doesn't apply to every chapter.

dteviot:
@ImmortalDreamer 

Apologies, I decided to steal (some of your code) and fix.  (I think)

ImmortalDreamer:
@dteviot No worries. As long as it's fixed.

Issue #2424, opened 2026-01-21
chrome gets recognized as firefox


gamebeaker:
Chrome version 144.0.7559.59 gets recognized as firefox.

<img width="646" height="180" alt="Image" src="https://github.com/user-attachments/assets/64683c00-fce2-4cf1-85c9-d981169b6bc5" />

gamebeaker:
Here is the reason why: https://developer.chrome.com/docs/extensions/reference/api

<img width="854" height="102" alt="Image" src="https://github.com/user-attachments/assets/9368237b-b359-4486-b1ee-312c344decee" />

gamebeaker:
@dteviot maybe push the new version to the chrome store as the current stable branch is 144.0.7559.96 -> the newest updated version from chrome doesn't work anymore with WebToEpub.

dteviot:
Test versions for Firefox and Chrome have been uploaded to https://github.com/dteviot/WebToEpub/releases/tag/developer-build. Pick the one suitable for you, follow the "How to install from Source (for people who are not developers)" instructions at https://github.com/dteviot/WebToEpub/tree/ExperimentalTabMode#user-content-installation and let me know how it goes.
Tested with:

Chrome version 143 and 144
Notes
Time taken: 30 minutes

dteviot:
@gamebeaker 

Updated version (1.0.11.0) has been submitted to Firefox and Chrome stores.
Firefox version should be available within two weeks (but might take 4 to 6 weeks)
Chrome might be available in a few hours to 21 days.  (Typically 3 to 7 days).

PR #2425, opened 2026-01-21
change chrome firefox recognition

gamebeaker:
#2424

### DIFF
--- a/plugin/js/Util.js
+++ b/plugin/js/Util.js
@@ -32,7 +32,9 @@ const util = (function() {
         }
         else
         {
-            return (typeof (browser) !== "undefined");
+            // this only works as long as firefox hasn't implemented this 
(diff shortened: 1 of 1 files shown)


dteviot:
@gamebeaker 

Given older versions of chrome DON'T implement browser, possibly this code should be

```javascript
return (typeof (browser?.runtime?.PlatformNaclArch) == "undefined");
```

dteviot:
@gamebeaker

Optional Chaining doesn't seem to work.  Probably needs to be something like
```javascript
            else if (typeof (browser) === "undefined")
            {
                return false;
            }
            else
            {
                // this only works as long as firefox hasn't implemented this 
                // https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/runtime/PlatformNaclArch
                return (typeof (browser.runtime.PlatformNaclArch) == "undefined");
            }
```

gamebeaker:
@dteviot i created #2427

PR #2433, opened 2026-01-23
Twkan.com parser addition

joelsnl:
its a parser for twkan because 69shu that used to work fine has broken down mostly

### DIFF
--- a/plugin/js/parsers/TwkanParser.js
+++ b/plugin/js/parsers/TwkanParser.js
@@ -1,0 +1,149 @@
+"use strict";
+
+parserFactory.register("twkan.com", () => new TwkanParser());
+
+class TwkanParser extends Parser {
+    constructor() {
+        super();
+        // Add delay between requests to avoid rate limiting
+        this.minimumThrottle = 1000;
+    }
+
+    async getChapterUrls(dom) {
+        // Extract book ID from the current URL
+        let bookId = this.extractBookId(dom.baseURI);
+        if (!bookId) {
+            // Fallback: try to get from meta tag
+            let metaBookId = dom.querySelector("meta[property='og:book_id']");
+            if (metaBookId) {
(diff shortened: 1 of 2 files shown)


joelsnl:
it was made with the help of claude code

joelsnl:
Output epub has been fully tested successfully

dteviot:
@joelsnl 

I think it's more complicated than it needs to be, but it appears to work.  And a quick inspection shows nothing obviously wrong or problematic.
I'll accept it.  Thank you for your contribution.

FYI, You didn't add yourself to package.json or readme.md, but I will add that for you.

joelsnl:
@dteviot Do you mind pointing me to a specific parser so that I can see what makes this more complicated than it needs to be? Would be wonderful to learn more from experienced people like you also don't want to take too much time from you. Thank you

dteviot:
@joelsnl 

>I can see what makes this more complicated than it needs to be?

In general, it's trying multiple ways to get things.  e.g. findCoverImageUrl()

Just needs to be 
```javascript
    findCoverImageUrl(dom) {
        return util.getFirstImgSrc(dom, ".bookbox");
    }
```
Usually, only reason functions try multiple ways is because site has changed over time.  Or is a nasty site with multiple formats.

Have a look at [Template.js](https://github.com/dteviot/WebToEpub/blob/ExperimentalTabMode/plugin/js/parsers/Template.js).  This is the starting point I use for most parsers.

joelsnl:
> Usually, only reason functions try multiple ways is because site has changed over time. Or is a nasty site with multiple formats.
> 
> Have a look at [Template.js](https://github.com/dteviot/WebToEpub/blob/ExperimentalTabMode/plugin/js/parsers/Template.js). This is the starting point I use for most parsers.

Thank you very much @dteviot, appreciate it

Issue #2116, opened 2025-08-14
Some .then() that can be replaced with await


dteviot:
Minor ToDo task for me.

There's a number of places where then() can be replaced with await
e.g. Parser.js
* onLoadFirstPage()
* fetchWebPageContent()
* fetchImagesUsedInDocument()

Also 
* KobatochanParser.js
* WixParser.js
* DefaultParserUI.js, line 149
* Imgur.js expandGalleries()
* Main.js, lines 371, 572

justinmmott:
I noticed this was a relatively easy change and had been partially completed, so I finished it up. I mentioned it in the [PR](https://github.com/dteviot/WebToEpub/pull/2453), but I may as well bring it up here as well. Is there any reason as to why you only listed these specific files instead of wanting to factor out the .thens() throughout the project?

dteviot:
@justinmmott 
> why you only listed these specific files

From memory, those were low risk.  IIRC, there's much more complicated promises in HttpClient.js, especially when the retry/error reporting error handling is factored in.

PR #2453, opened 2026-02-05
Refactor/2116

justinmmott:
Refactor with the original goal of completing [2116](https://github.com/dteviot/WebToEpub/issues/2116).

- Added `pack.js` and `release.js` to eslint
- Fixed Buffer() deprecation
- Changed .then()s to await for those listed in 2116.
  - Some of those changes were already accomplished in [453bea9](https://github.com/justinmmott/WebToEpub/commit/453bea9ae61a179dba8cfc698046c9e21caaa6a3#diff-77f1c72b88f0d58a5f5fbae2778915b5f40b116d6a1ba597373aaf7d4be72cd9)
  - Did a couple of extra so that any file edited had no .then()s
  - The issue listed `KobatochanParser.js`, which would have been an easy change. However, I noticed that this website has been down since 2021, so I didn't make the change. 
    - https://www.reddit.com/r/Trashero/comments/17elj8z/what_happened_to_kobatochan_translations/
    - https://www.patreon.com/kobatochan
      - The most recent post from 2021 says, "We are down"

I wasn't sure if there was a reason that only these files were listed in the issue, so I didn't go any further in the refactor. Would be happy to change the rest of the .then()s in another refactor if that is wanted.

### DIFF
--- a/eslint/.eslintrc.js
+++ b/eslint/.eslintrc.js
@@ -3,6 +3,7 @@ module.exports = {
         "browser": true,
         "es6": true,
         "webextensions": true,
+        "node": true,
     },
     "parserOptions": {
         "ecmaVersion": 2022
--- a/eslint/pack.js
+++ b/eslint/pack.js
@@ -12,26 +12,26 @@ var extractFileListFromHtml = function(htmlAsString) {
             .map(e => e.getAttribute("src"));
     }
     return [];
-}
+};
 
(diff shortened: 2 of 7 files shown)


justinmmott:
Didn't notice it before, but it seems the github workflow checks don't run the tests. Here is the output of the tests
image

dteviot:
@justinmmott 

Thank you for your contribution

PR #2474, opened 2026-02-13
Refactor/2116 2

justinmmott:
Change the rest of the `.then()`s to async-await. 

Ran the tests on the first commit, where there are no changes to the test files, and got all the tests to pass.

It's late, and I haven't given the code a once-over for style mistakes, so leaving this as a draft for now.

### DIFF
--- a/eslint/pack.js
+++ b/eslint/pack.js
@@ -14,10 +14,9 @@ var extractFileListFromHtml = function(htmlAsString) {
     return [];
 };
 
-var getFileList = function(fileName) {
-    return readFilePromise(fileName).then(function(data) {
-        return extractFileListFromHtml(data.toString());
-    });
+var getFileList = async function(fileName) {
+    let data = await readFilePromise(fileName);
+    return extractFileListFromHtml(data.toString());
 };
 
 var adjustedFileListForEslint = function(fileList) {
@@ -55,18 +54,15 @@ var writeFilePromise = function(fileName, buffer) {
 };
 
 // package geting all the files
-var readAllFiles = function(fileList, loadedFiles) {
-    return fileList.reduce(function(sequence, fileName) {
-        return sequence.then(function() {
-            return readFilePromise(fileName);
-        }).then(function(data) {
-            console.log("saving file:  " + fileName);
-            loadedFiles.push({
-                fileName: fileName, 
-                text:     data.toString()
-            });
+var readAllFiles = async function(fileList, loadedFiles) {
+    return await fileList.forEach(async function(fileName) {
+        let data = await readFilePromise(fileName);
+        console.log("saving file:  " + fileName);
+        loadedFiles.push({
+            fileName: fileName, 
+            text:     data.toString()
         });
-    }, Promise.resolve());
+    });
 };
 
 var countLines = function(fileText) {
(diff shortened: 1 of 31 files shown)


eslint/pack.js
```diff
@@ -77,56 +73,25 @@ var makeIndexLine = function(fileName, startIndex, count) {
...
-    let temp = "";
-    let lineCount = 0;
-    let index = "";
-    for (let f of loadedFiles) {
-        temp += f.text;
-        let count = countLines(f.text);
-        index +=  makeIndexLine(f.fileName, lineCount, count);
-        lineCount += count;
-    }
-    fs.writeFileSync("packed.js", temp);
```
dteviot: Where are you creating files packed.js and index.csv?

eslint/pack.js
```diff
@@ -169,44 +130,29 @@ var addCssFileToZip = function(zip, fileName) {
...
+        await addImageFileToZip(zip, "ChapterStateNone.svg");
+        await addImageFileToZip(zip, "ChapterStateSleeping.svg");
+        await addImageFileToZip(zip, "FileEarmarkCheck.svg");
+        await addImageFileToZip(zip, "FileEarmarkCheckFill.svg");
+        await addCssFileToZip(zip, "default.css");
+        await addCssFileToZip(zip, "alwaysDark.css");
+        await addCssFileToZip(zip, "autoDark.css");
+        let fileList = await getFileList("../plugin/popup.html");
+        let localeNames = await getLocaleFilesNames();
+        ["js/ContentScript.js"].concat(localeNames).concat(fileList.filter(n => !n.includes("/experimental/")));
```
dteviot, line 147: I think this needs to be 

```javascript
fileList = ["js/ContentScript.js"].concat(localeNames).concat(fileList.filter(n => !n.includes("/experimental/")));
```

So that next line processes the correct list of files

plugin/js/parsers/ComrademaoParser.js
```diff
@@ -16,12 +16,12 @@ class ComrademaoParser extends Parser {
         document.getElementById("removeOriginalRow").hidden = false; 
     }
 
-    getChapterUrls(dom, chapterUrlsUI) {
-        return this.getChapterUrlsFromMultipleTocPages(dom,
+    async getChapterUrls(dom, chapterUrlsUI) {
```
dteviot, line 19: Pretty sure this needs to be 

```javascript
        return (await this.getChapterUrlsFromMultipleTocPages(dom,
            ComrademaoParser.extractPartialChapterList,
            ComrademaoParser.getUrlsOfTocPages,
            chapterUrlsUI
        )).reverse();
```

plugin/js/parsers/QinxiaoshuoParser.js
```diff
@@ -35,27 +35,25 @@ class QinxiaoshuoParser extends Parser {
...
         if (url === null) {
             return Promise.resolve(finalDom);
         } else {
-            return HttpClient.wrapFetch(url).then(function(xhr) {
-                fetchedUrls.add(url);
-                QinxiaoshuoParser.copyContentNodes(finalDom, xhr.responseXML);
-                let nextUrl = QinxiaoshuoParser.urlOfNextPageOfChapter(xhr.responseXML, fetchedUrls);
-                return QinxiaoshuoParser.fetchPagesOfChapter(finalDom, fetchedUrls, nextUrl);
-            });
+            let xhr = await HttpClient.wrapFetch(url);
```
dteviot, line 52: This is using recursion due to use of then.  Can be re-written as simple iteration.

plugin/js/HttpClient.js
```diff
@@ -297,42 +297,35 @@ class FetchResponseHandler {
...
     }
 
-    responseToJson(response) {
-        return response.text().then(function(data) {
-            this.json =  JSON.parse(data);
-            return this;
-        }.bind(this));
+    async responseToJson(response) {
+        let data = await response.text();
+        this.json =  JSON.parse(data);
```
dteviot, line 328: You're not returning this

eslint/pack.js
```diff
@@ -55,18 +54,15 @@ var writeFilePromise = function(fileName, buffer) {
...
-        return sequence.then(function() {
-            return readFilePromise(fileName);
-        }).then(function(data) {
-            console.log("saving file:  " + fileName);
-            loadedFiles.push({
-                fileName: fileName, 
-                text:     data.toString()
-            });
+var readAllFiles = async function(fileList, loadedFiles) {
+    return await fileList.forEach(async function(fileName) {
```
zolantris, line 58: reduce should be used here not a forEach. Return of forEach will be undefined. (if this is an array)

See this Doc on Array.prototype.forEach. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/forEach#return_value 

What could be optimized is flattening the functions. Use an await a Promise.All for each of these items,

And each promise item in an array.

After promise.all you can return the array and it will be for all fulfilled promises. But it would throw a error still if anything failed in promise.all

zolantris: Regardless.

Would be good to have unit tests of this functionality to prove it's not thread blocking at the await syntax similar to original implementation and that the return is as expected

eslint/pack.js
```diff
@@ -239,18 +185,42 @@ var makeManifestForChrome = function(data) {
...
-readFilePromise("../plugin/manifest.json")
-    .then(function(data) {
-        packExtension(makeManifestForFirefox(data), ".xpi");
-        packExtension(makeManifestForChrome(data), ".zip");
-    }).catch(function(err) {
+    return await packNonManifestExtensionFiles(zipWriter, "WebToEpub" + manifest.version + fileExtension);
+};
+
+
+(async () => {
```
zolantris, line 196: Make this a function like init() that can be called. Call it afterwards but wrap it in a process.env variable so init is not called on import/require

This would make the code much more testable by not forcing execution on evaluation.

eslint/pack.js
```diff
@@ -77,56 +73,25 @@ var makeIndexLine = function(fileName, startIndex, count) {
...
 };
 
-var addFilesToZip = function(zip, fileList) {
-    return fileList.reduce(function(sequence, fileName) {
-        return sequence.then(function() {
-            return addToZipFile(zip, fileName, "../plugin/" + fileName);
-        });
-    }, Promise.resolve());
+var addFilesToZip = async function(zip, fileList) {
+    return await fileList.forEach(async function(fileName) {
```
zolantris, line 92: Same undefined bug as above.

zolantris: Add tests for inputs/outputs please!

dteviot: @zolantris 

eslint/pack.js is a build script.  Providing tests is somewhat redundant as it's purpose is to assemble the files into the WebToEpub extension.  So, it should either work or not.  Easiest way to check is run script before changes and save WebToEpub, then make changes, build again and compare built WebToEpub.  Should be same.

PR #2491, opened 2026-02-23
Update MimihuiParser

Yomafil:
Updated the MimihuiParser to correctly get the description and added support to get the story tags.

### DIFF
--- a/plugin/js/parsers/MimihuiParser.js
+++ b/plugin/js/parsers/MimihuiParser.js
@@ -14,7 +14,9 @@ class MimihuiParser extends Parser {
 
         let menu = tocPage.querySelector(".chapter-list");
 
-        return util.hyperlinksToChapterList(menu);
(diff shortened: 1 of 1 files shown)


Yomafil:
@Kiradien I still want to see if I can add the unlisting of VIP chapters and the cleanup of chapter titles in the TOC.
But I can make it in a separate pull request, if needed.

Kiradien:
It's better to add the changes to the existing request then; technically the best practice is to only create the pull request when everything seems perfect, but I've been known to do the same as you with this.

Yomafil:
I was a little hasty in making this pull request, wasn't originally planning on working on more than just adding tags to the parser. I'll be more careful next time.

Yomafil:
Tested, everything seems to work, I added everything I needed. This can be merged.

gamebeaker:
@Kiradien i use "Squash and merge" to prevent cluttering the git history.
image

PR #2493, opened 2026-02-23
Update YedugeParser

Yomafil:
Updated the YedugeParser, it can now get tags, it won't include VIP chapters, and it will now clean the chapters title in the ToC by removing "免费" (free) from the end of the titles.


gamebeaker:
@Yomafil again thanks for your work.

PR #2494, opened 2026-02-23
Update UaaParser

Yomafil:
Fixed cover image and added tags.


gamebeaker:
@Yomafil again thank you.

PR #2497, opened 2026-02-24
Update 69shuParser

Yomafil:
Added tags and description metadata to 69shuParser.
Registered 69yuedu.net as a Dead Site, automatically redirect to 69shuba.com.


gamebeaker:
@Yomafil i am happy that i implemented https://github.com/dteviot/WebToEpub/commit/67938f060b622992112271965fdd7b6a39055b9d so in case a website from type 69shu is a little different it no longer crashes WebToEpub.
Thanks for the help.

Issue #2500, opened 2026-02-25
Patreonparser failed to fetch url post


BettyBanana:
**Describe the bug**
It was working yesterday. But today I tried it for a couple of times already and on different browser and really can't fetch posts url. And it was the latest version.

**To Reproduce**
chrome-extension://ellngllbglabjcdegajejoleaopibfad/popup.html?id=1923207532

**Screenshots**

**Desktop (please complete the following information):**

**Smartphone (please complete the following information):**
 Redmi (qeutta, kiwi, lemur and firefox)

**Additional context**
(I'm no dev I really tried my best to describe it guys.)

BettyBanana:
<img width="1071" height="526" alt="Image" src="https://github.com/user-attachments/assets/d4f2d456-de75-4dc6-8c44-9fda208a2b6d" />

gamebeaker:
@BettyBanana 
please provide a patreon url .
Which version of WebToEpub do you use? (There are fixes for patreon live in the developer version.)
https://github.com/dteviot/WebToEpub/releases/tag/developer-build

BettyBanana:
https://www.patreon.com/collection/1282526?view=expanded
Version 1.0.11.36

BettyBanana:
I tried it on different versions actually but none of them work. But works on different sites 

<img width="1048" height="378" alt="Image" src="https://github.com/user-attachments/assets/5596418c-61d6-439c-95e1-d0e58736ea39" />

BettyBanana:
Also tried it on the very latest version 1.0.11.36 still didn't work.



<img width="919" height="618" alt="Image" src="https://github.com/user-attachments/assets/c7481425-e10a-40d8-92ea-dd83eb91c600" />

nitramkh:
Alright, I'm not @gamebeaker but out of curiosity I looked into it. I was able to reproduce this and seemingly make a fix. It seems that at some point Patreon changed the layout of the collection view(s) which broke the CSS selectors used to extract the chapters.

I'm currently figuring out how to make a PR for this and will link it here at that time, but as I'm just some guy, unless you want to try to run my forked branch, we'll have to wait for a bit :)

BettyBanana:
Thank you @nitramkh! Even though I don't understand a thing I'm glad it can be fix. I'll just have to wait.

nitramkh:
No problem :)

[...]

PR #2502, opened 2026-02-25
Fix Patreon collections view(s) CSS selectors (#2500)

nitramkh:
This should be a fix for #2500.

I squashed the contributors/credits and fix commits since they were tiny, but I can undo this if that is preferred. Also if I've done something else wrong feel free to point it out and I will fix it ASAP :)

## Pull Request Check List

- **Do all existing unit tests pass?**
Yes.

- **Have all warnings and errors reported by eslint been fixed?**
Technically no, but no eslint warning for the files I've touched.

- **If you've added new behavior ...**
No.

- **Have you updated the "contributors" section of packages.json with your name? (and ideally email, so I can contact you easily.)**
Yes.

- **Have you updated the "Credits" section of readme.md with your name?**
Yes.

- **Are you committing to the ExperimentalTab branch?**
Yes.

- **Have you rebased your commit?**
Yes.

### DIFF
--- a/plugin/js/parsers/PatreonParser.js
+++ b/plugin/js/parsers/PatreonParser.js
@@ -19,7 +19,7 @@ class PatreonParser extends Parser {
 
     getCollectionLinks(dom) {
         let getTitle = (e) => {
-            return [...e.querySelectorAll("span.cm-ugDCiy")]
+            return [...e.querySelectorAll("span.LineClamp-module__N_eOMG__lineClamp1")]
                 .map(s => s.textContent.trim())
                 .join(" ");
         };
@@ -29,7 +29,7 @@ class PatreonParser extends Parser {
             let getLink = (e) => {
                 return e.querySelector("a");
             };
-            let linksContainer = [...dom.querySelectorAll("div.cm-hhCVrV.cm-WzHHbB div.cm-bkNQIo:not(:has(svg[data-tag='IconLock'])) div:not([class])")];
+            let linksContainer = [...dom.querySelectorAll("div.ListPost-module__d2AM5a__listPost div:not([class])")];
             return linksContainer.map(linkContainer => {
                 return {
                     sourceUrl: getLink(linkContainer).href,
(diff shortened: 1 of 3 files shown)


plugin/js/parsers/PatreonParser.js
```diff
@@ -29,7 +29,7 @@ class PatreonParser extends Parser {
             let getLink = (e) => {
                 return e.querySelector("a");
             };
-            let linksContainer = [...dom.querySelectorAll("div.cm-hhCVrV.cm-WzHHbB div.cm-bkNQIo:not(:has(svg[data-tag='IconLock'])) div:not([class])")];
+            let linksContainer = [...dom.querySelectorAll("div.ListPost-module__d2AM5a__listPost div:not([class])")];
```
Kiradien, line 32: Removal of :not(:has(svg[data-tag='IconLock'])) is a bit of an annoyance for people without full access subscriptions, but it works.

Kiradien: Also, div:not([class]) appears redundant in the new UI; that was a workaround for sloppy code on their part.

Kiradien:
Alright, I did point out a few minor issues - but the important part is the code works, and we want people to actually be able to use it, so that stuff can be fixed in the future by yourself or me. 
Thanks for getting this done before me, hard to keep on top of all patreon's changes.

plugin/js/parsers/PatreonParser.js
nitramkh: Agreed on both points. As for the redundancy I was going for more of fast and functional over perfect before going to sleep. As for the `:not(:has(svg[data-tag='IconLock']))` that's my mistake. I don't make use of the collections nor partial subs basically ever, and didn't realize locked posts could show up; and therefore figured that part was legacy/redundant...

I will fix this, test more thoroughly, and go through the new HTML/CSS again to see if the selectors I found can be improved now that theres not as much time pressure as it's only partially broken (and I don't have a _Sleep Deprivation™_ debuff).

Issue #2470, opened 2026-02-11
Please add site https://soafp.com


trininly8831-hub:
Please note, there are only a few developers working on WebToEpub, and we're not paid for doing this.  (WebToEpub is completely free, and generates no money.) By asking to add a site, you're asking us to give you some of our limited free time. So, we think it's not unreasonable for me to ask you to do as much as you can to help us.

### Check the site has not already been requested.
Not requested

### Did you try using the Default Parser for the site?  If not, why not?
Instructions for using the default parser can be found at https://github.com/dteviot/WebToEpub/wiki/FAQ#how-to-convert-a-new-site-using-the-default-parser

### What settings did you use?  What didn't work?
* URL of first chapter
* CSS selector for element holding content to put into EPUB
* CSS selector for element holding Title of Chapter
* CSS selector for element(s) to remove

### If the Default Parser did not work, if you have developer skills, did you try writing a new parser?
Instructions https://github.com/dteviot/WebToEpub/wiki/FAQ#how-to-write-a-new-parser

SpiralBaka:
the default parser (Advanced Options --> Manually Select Parser --> Default) still works. here, I'll write it:
-URL of first chapter: https://soafp.com/novel/anti-cuckold/v1ch1-part-1/
-CSS selector for element holding content to put into EPUB: **div.entry-content**
-CSS selector for element holding Title of Chapter: **h1.entry-title**
-CSS selector for element(s) to remove: **button**
only the toc need to be updated, but even without updating, you can still select the chapters after using the default parser.
I'm not a developer btw, so this is just a little help

trininly8831-hub:
Works fine thanks

dteviot:
@trininly8831-hub 
Test versions for Firefox and Chrome have been uploaded to https://github.com/dteviot/WebToEpub/releases/tag/developer-build. Pick the one suitable for you, follow the "How to install from Source (for people who are not developers)" instructions at https://github.com/dteviot/WebToEpub/tree/ExperimentalTabMode#user-content-installation and let me know how it goes.

dteviot:
Reopen, so I know to notify you when Chrome and Firefox stores are updated.

dteviot:
@trininly8831-hub @SpiralBaka 

Updated version (1.0.12.0) has been submitted to Firefox and Chrome stores.
Firefox version should be available within two weeks (but might take 4 to 6 weeks)
Chrome might be available in a few hours to 21 days.  (Typically 3 to 7 days).

PR #2503, opened 2026-02-25
Add SoafpParser

Yomafil:
Add SoafpParser for site https://soafp.com/.
Addition for #2470

The site doesn't provide any authors for books, so we're using the TL.

### DIFF
--- a/plugin/js/parsers/SoafpParser.js
+++ b/plugin/js/parsers/SoafpParser.js
@@ -6,11 +6,10 @@ class SoafpParser extends Parser {
     constructor() {
         super();
     }
-    
-    async getChapterUrls(dom) {
 
-        return [...dom.querySelectorAll(".volume-container ul li a")]
-            .map(a => util.hyperLinkToChapter(a));
+    async getChapterUrls(dom) {
+        let menu = dom.querySelector(".novel-chapter-list");
+        return util.hyperlinksToChapterList(menu);
     }
 
     findContent(dom) {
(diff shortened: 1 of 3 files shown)


plugin/js/parsers/SoafpParser.js
```diff
@@ -0,0 +1,54 @@
...
+        super();
+    }
+
+    async getChapterUrls(dom) {
+        let menu = dom.querySelector(".novel-chapter-list");
+        return util.hyperlinksToChapterList(menu);
+    }
+
+    findContent(dom) {
+        return dom.querySelector("div.entry-content");
```
dteviot, line 16: FYI.

.entry-content and .entry-title mean this is a WordPress based site.
If you SoafpParser from WordpressBaseParser, these functions are not needed here.

plugin/js/parsers/SoafpParser.js
```diff
@@ -0,0 +1,54 @@
...
+        return dom.querySelector(".entry-title");
+    }
+
+    extractAuthor() {
+        //The site doesn't provide any author for books, so we're just using the TL
+        return "Soafp";
+    }
+
+    extractLanguage(dom) {
+        return dom.querySelector("html").getAttribute("lang");
```
dteviot, line 29: extractLanguage isn't needed in this case.  WebToEpub will look here for the language.

dteviot (commented):
Some minor points for future reference.

dteviot:
@Yomafil 

And thank you again.

PR #2504, opened 2026-02-26
Add AliceswParser

Yomafil:
Added AliceswParser for site https://www.alicesw.com/.


Sen01fd:
https://www.alicesw.com/  Every chapter show this error  Loading...
Chapter loading...

Yomafil:
@Sen01fd Looks like they now encrypt their content. Nothing I can do in the short term.
You should probably create an issue about it.

PR #2509, opened 2026-02-27
Update UaaParser

Yomafil:
Updated findContent method of UaaParser
It should now, no longer bring back the content in a single paragraph.

### DIFF
--- a/plugin/js/parsers/UaaParser.js
+++ b/plugin/js/parsers/UaaParser.js
@@ -15,7 +15,19 @@ class UaaParser extends Parser {
     }
 
     findContent(dom) {
-        return dom.querySelector(".article");
+        let contentLines = dom.querySelectorAll(".article div");
+
+        let content = document.createElement("div");
+
+        contentLines.forEach(line => {
+            let p = document.createElement("p");
+
+            p.textContent = line.textContent;
+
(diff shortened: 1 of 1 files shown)


Yomafil:
Sadly, I can't seem to figure out a way to squash commits already in the GitHub branch repo. 
Can only do that on local branch. 
I can create a new branch each time I want to do a second update to a Parser, or I can keep using the same one.

plugin/js/parsers/UaaParser.js
```diff
@@ -15,7 +15,19 @@ class UaaParser extends Parser {
     }
 
     findContent(dom) {
-        return dom.querySelector(".article");
+        let contentLines = dom.querySelectorAll(".article div");
```
dteviot, line 18: findContent() is called multiple times.
Modifying content in findContent() can result in other changes being lost.

Correct place to do this is either customRawDomToContentStep() or preprocessRawDom()

dteviot (changes requested):
Going to reject this.  Will post fixed example shortly

dteviot:
@Yomafil 

See https://github.com/dteviot/WebToEpub/commit/23a69ca4088eb9c3632fe843abeae248a97cfce5

PR #2512, opened 2026-02-27
Update YedugeParser

Yomafil:
Updated the YedugeParser to now list the VIP chapters in the UI, but leave them deselected by default.


dteviot:
Thank you for this.
I note this parser and the Minihui one have a lot of code in common.
You could derive one from the other and eliminate the duplicated code.

Issue #2508, opened 2026-02-26
Please add site https://ranovel.com


XMAD222:
The extension gets everything except Chapter content. Instead of chapter content it's try to get data from ko-fi site.

nitramkh:
Okay so I took a look at this and was at least partially able to replicate the problem, but I/we are probably going to need more information. 

First, I was able to reproduce the problem if I quickly reload the site and then trigger WebToEpub before the chapters list has finished loading. But if you simply wait a bit after this has occurred, for me at least, it works as expected; with the singular exception that the chapter titles are incomplete (sending a PR _soon™_).

Please try again, but wait say 10 seconds after the page _**seems to have finished loading**_ before triggering WebToEpub and see what happens. If you are attempting to use "Load and Analyse" I'm going to have to pass this onto someone else as I don't know if it's possible to use on this site.

Finally, could you provide some additional information, such as:

1. Specific URLs of at least one, but preferably more novels which are not working
2. Screenshots of the issue
3. What browser/platform you are using
4. What version of WebToEpub you are using

XMAD222:
Link : https://ranovel.com/novel/infinite-regression-with-an-infinite-inventory/

I used the 1.0.11.47 version of the chrome extension on win 10. The extension tries top open a ko-fi page then gives error and after I click on open url for captcha, it opened a ko-fi page with cloudflare check. After the cloudflare captcha is complete I click on retry getting content and it worked.

While you at it, Can you please stop these texts to be scraped in every chapter. I checked few pages and all of these 3 texts have the same css selector in every page. I couldn't find the ranovel parser in the extension or I could've added the code myself.

1. At the Top of the page

Only Ran(o)vel dot com
Read Novel 𝒓𝒂𝒏𝒐𝒗𝒆𝒍 com

2.in the middle of the page

* * * * * * * * * * * * * *
* * Read Korean Novel * *
* * * Ranovel dot com * * *
* * * * * * * * * * * * * *

3. At the end of the page

Read Novel Ra/n(o)ve/l com
Read Novel 𝖗𝖆𝖓𝖔𝖛𝖊𝖑 com


Thanks for replying.

nitramkh:
Alright good, I have been trying with that link specifically, but I haven't been able to reproduce as you are describing it. I have a few more things to try, including setting up a Windows 10 VM to see if that makes any difference.

[...]

PR #2513, opened 2026-02-27
Fix https://ranovel.com/

nitramkh:
In response to #2508.

The site already naturally falls under the [MadaraParser](https://github.com/dteviot/WebToEpub/blob/ExperimentalTabMode/plugin/js/parsers/MadaraParser.js) and mostly works. I was able to replicate #2508 (on both webstore and dev build), but only if I trigger WebToEpub before the page has completed loading. If you wait until the chapter list has finished loading everything seemingly works; except the full chapter titles which this fixes.

Finally, a question: I created a sub-class for this parser following the example of the [KdtnovelsParser](https://github.com/dteviot/WebToEpub/blob/824aaed97e7babc2273279d6bc22a2c6c140e04f/plugin/js/parsers/MadaraParser.js#L143). But this seems deeply overkill, and since [MadaraParser](https://github.com/dteviot/WebToEpub/blob/ExperimentalTabMode/plugin/js/parsers/MadaraParser.js) already seems to be a multi-parser; should this change instead be a logic branch inside the [MadaraParser](https://github.com/dteviot/WebToEpub/blob/ExperimentalTabMode/plugin/js/parsers/MadaraParser.js)?

Anyways, if there are any issues I'm happy to fix ;)

EDIT: The tests failed before my force push because I'm dumb and did a: change - git add - change - git commit; so my local lint succeeded but not the remote... my bad...

## Pull Request Check List

- **Do all existing unit tests pass?**
Yes.

- **Have all warnings and errors reported by eslint been fixed?**
Yes.

- **If you've added new behavior ...**
No.

- **Have you updated the "contributors" section of packages.json with your name and email?**
Already there.

- **Have you updated the "Credits" section of readme.md with your name?**
Already there.

- **Are you committing to the ExperimentalTab branch?**
Yes.

- **Have you rebased your commit?**
Yes.

### DIFF
--- a/plugin/js/parsers/MadaraParser.js
+++ b/plugin/js/parsers/MadaraParser.js
@@ -69,9 +69,8 @@ class MadaraParser extends WordpressBaseParser {
         return descriptionElement === null ? "" : descriptionElement.textContent.trim();
     }
     
-
     removeUnwantedElementsFromContentElement(element) {
-        util.removeChildElementsMatchingSelector(element, "div.addtoany_share_save_container");
+        util.removeChildElementsMatchingSelector(element, "div.addtoany_share_save_container, div.code-block");
         super.removeUnwantedElementsFromContentElement(element);
     }
 


dteviot:
@nitramkh 

> Finally, a question: I created a sub-class for this parser following the example of the [KdtnovelsParser](https://github.com/dteviot/WebToEpub/blob/824aaed97e7babc2273279d6bc22a2c6c140e04f/plugin/js/parsers/MadaraParser.js#L143). But this seems deeply overkill, and since [MadaraParser](https://github.com/dteviot/WebToEpub/blob/ExperimentalTabMode/plugin/js/parsers/MadaraParser.js) already seems to be a multi-parser; should this change instead be a logic branch inside the [MadaraParser](https://github.com/dteviot/WebToEpub/blob/ExperimentalTabMode/plugin/js/parsers/MadaraParser.js)?

An excellent question.  And you're correct.  A better solution would be to have findChapterTitle(dom) search in each of the 3 locations in turn until it finds something.  Then the MadaraParser can work with a bigger range of sites.

nitramkh:
@dteviot Perfect. For now I have updated my commit to remove the subclass and merge it into the base parser. I have retested the target site and two random other [MadaraParser](https://github.com/dteviot/WebToEpub/blob/ExperimentalTabMode/plugin/js/parsers/MadaraParser.js) sites and nothing seems broken.

I'm not sure if you implied this, but from my point of view the same could be said for both the [KdtnovelsParser](https://github.com/dteviot/WebToEpub/blob/824aaed97e7babc2273279d6bc22a2c6c140e04f/plugin/js/parsers/MadaraParser.js#L143) and the [MadaraVariantParser](https://github.com/dteviot/WebToEpub/blob/824aaed97e7babc2273279d6bc22a2c6c140e04f/plugin/js/parsers/MadaraParser.js#L102). The former as mentioned above, but the latter as I can't see any real reason for its existence, other than perhaps to not bloat the main parser? But if thats the case, that to me seems like a bad tradeoff?

Anyway, I did also convert the [KdtnovelsParser](https://github.com/dteviot/WebToEpub/blob/824aaed97e7babc2273279d6bc22a2c6c140e04f/plugin/js/parsers/MadaraParser.js#L143) into the main [MadaraParser](https://github.com/dteviot/WebToEpub/blob/ExperimentalTabMode/plugin/js/parsers/MadaraParser.js), but I was completely unable to test it as even with the current upstream and webstore version as my IP is seemingly completely blocked; not even a capcha or nothing. So I'll maybe put that in another PR as to not block this one?

nitramkh:
Okay so, I think this PR is done at this point. Me and the author of #2508 have come to the conclusion that it was more likely than not a local issue causing any remaining problems.

The second commit fixes the remaining concerns in #2508 regarding annoying/unwanted elements not being properly stripped. I've retested everything on the subject page, and one other MadaraParser site (I genuinely could only get one of the entire list to work...) and the fix works, and I don't think its likely I've broken any other MadaraParser sites.

plugin/js/parsers/MadaraParser.js
```diff
@@ -69,14 +69,13 @@ class MadaraParser extends WordpressBaseParser {
...
-
     removeUnwantedElementsFromContentElement(element) {
-        util.removeChildElementsMatchingSelector(element, "div.addtoany_share_save_container");
+        util.removeChildElementsMatchingSelector(element, "div.addtoany_share_save_container, div.code-block");
         super.removeUnwantedElementsFromContentElement(element);
     }
 
     findChapterTitle(dom) {
-        return dom.querySelector("ol.breadcrumb li.active, .wp-manga-chapter.reading a").textContent;
+        return dom.querySelector("ol.breadcrumb li.active, .wp-manga-chapter.reading a, #chapter-heading").textContent;
```
Kiradien: I'm curious as to when the #chapter-heading part is actually necessary; the text it grabs is a fair bit different from the normal chapter titles, as it includes the title of the novel as well.

I'm assuming the breadcrumb fails at some point, but was the select box text not available? e.g. 

`
return dom.querySelector("ol.breadcrumb li.active, .wp-manga-chapter.reading a").textContent || dom.querySelector("div.select-view select")?.selectedOptions[0]?.innerText
`

... It's not really a huge issue, I'm personally just not a fan of the entire story title being appended to the chapter name - some stories have ridiculolusly long titles.

nitramkh: > I'm curious as to when the #chapter-heading part is actually necessary; the text it grabs is a fair bit different from the normal chapter titles, as it includes the title of the novel as well.
> ...
> ... It's not really a huge issue, I'm personally just not a fan of the entire story title being appended to the chapter name - some stories have ridiculolusly long titles.

As to whether it is necessary, it depends entirely on whether you want the "full title" by which I mean the story title. I tend to prefer leaving them as is, but I get why it can be annoying.

> I'm assuming the breadcrumb fails at some point, but was the select box text not available? e.g.
> 
> `return dom.querySelector("ol.breadcrumb li.active, .wp-manga-chapter.reading a").textContent || dom.querySelector("div.select-view select")?.selectedOptions[0]?.innerText`
> 

If you revert this change, you get only the `Chapter X` for each. So no need to go even that far. I was just already there and looking for problems and went for it, I frankly didn't consider that this might be intentional... So I can nuke this commit easily enough, just say the word.

Like I said, I just noticed it and changed it without thinking too deeply; I don't really have an opinion either way :)

Kiradien: The biggest problem with including the Story Title is that it can actually get long enough to cut of the actual chapter number and actual chapter title (when available).

While it can be handy for debugging in rare cases, people typically know what story they are reading - having it reinforced at the start of every single chapter is anywhere from redundant to downright irritating for readers using accessibility options (such as TTS).

I'm not necessarily going to argue on this, since I'll just revert to the old code on my personal branch, but  I'm personally not a fan of having the story title in there - it's already in the meta file properties and file name, after all. I think we've had requests to remove this from certain parsers in the past.

nitramkh: I'll trust your judgement on this one. Regardless, this was very much a "go looking for problems, and you will doubtlessly find one" situation, now transitioning into a "if it ain't broke don't fix it".

I'll nuke the commit, and if someone ever requests it, it would be easy enough to implement.

Kiradien (approved):

dteviot:
@Kiradien

Is there a reason you did not merge this?

Kiradien:
Ah, tbh I thought I did, my bad. Clicking the merge now.

PR #2523, opened 2026-03-05
Parser updates

crybx:
Two new parsers and updates to FictioneerParser.

### DIFF
--- a/plugin/js/parsers/MistmintHavenParser.js
+++ b/plugin/js/parsers/MistmintHavenParser.js
@@ -1,0 +1,209 @@
+"use strict";
+
+parserFactory.register("mistminthaven.com", () => new MistmintHavenParser());
+
+class MistmintHavenParser extends Parser {
+    constructor() {
+        super();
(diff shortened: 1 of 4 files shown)


plugin/js/parsers/MistmintHavenParser.js
```diff
@@ -0,0 +1,209 @@
...
+        this.chapterTitles = new Map();
+    }
+
+    async getChapterUrls(dom) {
+        // Free chapters are in the first tab panel, inside a grid of <a> elements
+        let chapters = [...dom.querySelectorAll("[role='tabpanel'] a[href*='/novels/'][title^='Chapter']")]
+            .map(a => ({
+                sourceUrl: a.href,
+                title: a.title || a.querySelector(".font-semibold")?.textContent?.trim(),
+                isIncludeable: true
```
dteviot, line 17: Don't need to set isIncludeable if it's always true.

dteviot:
@crybx 

Thank you for your hard work

Issue #2531, opened 2026-03-08
https://wattpad.com.vn/


anhptt569-cmd:
Check the site has not already been requested.
Site has Not Been Requested.
https://wattpad.com.vn/

Provide URL for web page that contains Table of Contents (list of chapters) of a typical story on the site
https://wattpad.com.vn/boss-la-nu-phu

Did you try using the Default Parser for the site? If not, why not?
I dont know how, and dont have anyone that can help me.

Thank you for your amazing work !

dteviot:
@anhptt569-cmd 

Test versions for Firefox and Chrome have been uploaded to https://github.com/dteviot/WebToEpub/releases/tag/developer-build. Pick the one suitable for you, follow the "How to install from Source (for people who are not developers)" instructions at https://github.com/dteviot/WebToEpub/tree/ExperimentalTabMode#user-content-installation and let me know how it goes.
Tested with:
* Caution: I have not tested to verify.

dteviot:
Updated version (1.0.13.0) has been submitted to Firefox and Chrome stores.
Firefox version should be available within two weeks (but might take 4 to 6 weeks)
Chrome might be available in a few hours to 21 days.  (Typically 3 to 7 days).

PR #2533, opened 2026-03-09
Added wattpad.com.vn parser

nothing0074:
Addresses #2531 
warning this uses (not very good) ai code


dteviot:
@nothing0074 

> warning this uses (not very good) ai code

It could be cleaned up a little, but it's not too horrible.

Thank you for your hard work.

PR #2545, opened 2026-03-17
Update FalooParser

Yomafil:
Updated the FalooParser to now list all VIP chapters, but leave them unchecked by default.
Also, use the table of content to get the chapters, and get the non-truncated title of a chapter from it.

### DIFF
--- a/plugin/js/parsers/FalooParser.js
+++ b/plugin/js/parsers/FalooParser.js
@@ -12,15 +12,36 @@ class FalooParser extends Parser {
     }
 
     async getChapterUrls(dom) {
-        // If there are no VIP chapters, get all chapters
-        if (!dom.querySelector(".DivVip")) {
-            return [...dom.querySelectorAll("div.DivTable div div a")]
-                .map(a => util.hyperLinkToChapter(a));
+        let tocUrl = dom.querySelector(".T-L-T-C-Box2 a:nth-child(4)");
+
+        let tocDom = (await HttpClient.wrapFetch(tocUrl)).responseXML;
+
+        let nodes = [...tocDom.querySelectorAll("div.c_con_list, .c_con_viptitle")];
+
+        let isVipReached = false;
+        let chapters = [];
+
+        for (let node of nodes) {
+            if (node.classList.contains("c_con_viptitle")) {
+                isVipReached = true;
+                continue;
+            }
+
+            let links = node.querySelectorAll("div.c_con_li_detail_p a, a.c_con_li_detail");
+
+            for (let link of links) {
+                let span = link.querySelector("span");
+                let title = span.textContent.trim();
+
+                chapters.push({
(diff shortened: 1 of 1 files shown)


plugin/js/parsers/FalooParser.js
```diff
@@ -12,15 +12,36 @@ class FalooParser extends Parser {
...
+
+        let nodes = [...tocDom.querySelectorAll("div.c_con_list, .c_con_viptitle")];
+
+        let isVipReached = false;
+        let chapters = [];
+
+        for (let node of nodes) {
+            if (node.classList.contains("c_con_viptitle")) {
+                isVipReached = true;
+                continue;
```
dteviot: I think this is going to skip VIP chapters, rather than add them to the list.
I think you need to remove the continue.

dteviot (changes requested):

plugin/js/parsers/FalooParser.js
Yomafil: The reason we skip `c_con_viptitle` is that it's an empty header that is a sibling of `div.c_con_list`. It marks the end of free volumes and the beginning of VIP ones.
All volumes after it are VIP. I didn't find any other indicators. 
Maybe if a chapter is stored `a.c_con_li_detail` and not `div.c_con_li_detail_p a` we could count them as VIP more reliably?
I already tested the parser on multiple books and it should work as intended. 
https://b.faloo.com/1510132.html:
image
`.c_con_viptitle`:
image

plugin/js/parsers/FalooParser.js
```diff
@@ -12,15 +12,36 @@ class FalooParser extends Parser {
...
+            let links = node.querySelectorAll("div.c_con_li_detail_p a, a.c_con_li_detail");
+
+            for (let link of links) {
+                let span = link.querySelector("span");
+                let title = span.textContent.trim();
+
+                chapters.push({
+                    sourceUrl: link.href, 
+                    title: title,
+                    isIncludeable: !isVipReached
```
dteviot: This line
```javascript
isIncludeable: !isVipReached
```
Suggests you want to include all chapters, but mark the VIP chapters as disabled.
The "continue" line I commented on previously means the VIP chapters will be excluded from the chapter list.
Which of the two are you trying to do?

You should remove one of the continue or the isIncludeable lines.

Yomafil:
Changed how we check for VIP chapters, we now look at how a chapter element is stored inside `div.c_con_list`, and not just assuming that everything is VIP after hitting a single header.
If an element is `c_con_li_detail`, it's a VIP chapter.

It should be better now.
Haven't yet found a book that has a free volume, then a VIP one, and then another free one, so I couldn't test this case.
But it should be handled by this new way of checking for VIP.

dteviot:
@Yomafil 
Thank you for your hard work.

Issue #2471, opened 2026-02-12
Please add site https://noveldex.io 


trininly8831-hub:
Just updating

https://noveldex.io/

 

The site novelshub.org was now transferred 

### Provide URL for web page that contains Table of Contents (list of chapters) of a typical story on the site: 
https://noveldex.io/series/novel/a-heart-devoted-to-the-sword

### Did you try using the Default Parser for the site?  If not, why not?
Instructions for using the default parser can be found at https://github.com/dteviot/WebToEpub/wiki/FAQ#how-to-convert-a-new-site-using-the-default-parser : I try but I truly don't understand how to get it properly sorry

mmikail07:
+1

Laplace9:
+1

gamebeaker:
looks like they changed to next.js and have implemented protection measures. will have a look at it tomorrow.

trimmit66:
+1

AnubisRyu:
+1

hackrabbits1:
+1

nitramkh:
Okay so, I have a working but very unrefined parser for this; but I just need a bit of clarification from someone who knows more about this stuff like @gamebeaker. I wouldn't call this obfuscation, but the site uses some JS framework, which places the chapter content inside a script tag like this:

<img width="1506" height="45" alt="Image" src="https://github.com/user-attachments/assets/a9bac1ab-4f4a-4187-804e-7830598ba1d9" />

The content, in my opinion, does not look intentionally obfuscated; and instead it looks like a side effect of their JS framework. It is straight HTML which has been unicode-escaped, stored inside the script tag. This can easily be extracted, but does this count as obfuscation? And by extension would making a parser for this make the corporate overlords angry?

dteviot:
@nitramkh 

We support sites like that.  There are already a number of parsers which handle that sort of thing.

nova1-0:
I have been trying but I don't think that it supports this site, please do look into it . I don't know how to handle next.js  and creating parser from that much, would appreciate you looking into it.

nitramkh:
> I have been trying but I don't think that it supports this site, please do look into it . I don't know how to handle next.js and creating parser from that much, would appreciate you looking into it.

Apologies for my lateness. I am going to push a parser for this, but I work very sporadic hours and have been swamped since I made this comment; but I finally have time to get going again; so you can expect a initial version _hopefully soon™_.

dteviot:
[...]

PR #2546, opened 2026-03-17
Add http://noveldex.io/ parser (#2471)

nitramkh:
Fix for #2471.

This should be a mostly good parser for http://noveldex.io/, there are some improvements which can be made but I think it's sufficient for people who want to try using it rather than continuing to wait.

A list of things which may be of note in no particular order:

1. The CSS selectors used look a bit odd, since I've been busy I started on this 2 weeks ago and in that time I've seen three separate occurrences of change wherein either ids and classes were changed. But I observed no structural HTML changes, so I went for that instead.

2. The location of the chapter names are inconsistently placed, there is an "official" location for the chapter titles; but many stories seem to just chuck them in the content in a non-standardized way. So it just ignores these for now, but I may fix that later if someone cares.

3. This seems to primarily be a translation site and I've included the translators as an author. However, again many individual chapters seem to have the translator listed in a non-standardized way inside the chapter content. I ignored this as I couldn't see an obvious way to modify the author field after fetching the chapter contents; and didn't think a hacky implementation would be appreciated ;)

4. I didn't test this too deeply as I don't use the site myself, but I did download a couple stories using Chrome and Firefox, including the dev-tools mobile view thing; and skimmed the contents and it looked alright. But frankly, I figure if it doesn't work the requesters will let me know and I actually have time to fix it in a timely manner now.

5. I added automatic skipping of locked chapters which works fine. But I can't test if unlocking and downloading these chapters actually works. But I saw no harm in including it, with the hope that one of the requesters can test it; and if broken I can fix it or remove it.

As always, if there are any issues or even thing anything you just don't like, I'm happy to fix it ;)

<details>

    1. ⬜ Can you turn it off?
    2. ⬜ Is it off by default?

</details>

### DIFF
--- a/plugin/js/parsers/NoveldexParser.js
+++ b/plugin/js/parsers/NoveldexParser.js
@@ -1,0 +1,200 @@
+"use strict";
+
+parserFactory.register("noveldex.io", () => new NoveldexParser());
+
+/**
+ * Parser for the http://noveldex.io/ site.
+ */
+class NoveldexParser extends Parser { // eslint-disable-line no-unused-vars
+    constructor() {
+        super();
+    }
+
+    /*
+     * NOTE: As of 2026-03, it seems that classes and ids on this site can't be
+     * trusted to remain stable. In a short timespan, over three separate
+     * occurences, multiple classes and at least one id has been changed.
+     */
+
+    /**
+     * Function for extracting a specific value from any of the schema/context
+     * "scripts" in the main page dom.
+     * 
+     * @param { Document } dom Source to extract from.
+     * @param { string } key The property to find (first found).
+     * @returns { unknown | null } The found property, or null.
+     * 
+     * @private
+     */
+    extractFromContext(dom, key) {
+        /** @type { NodeListOf<HTMLScriptElement> } */
+        const contextContainers = dom.querySelectorAll("script[type='application/ld+json']");
+
+        const property = Array.from(contextContainers, contextContainer => {
+            let context = null;
+            
+            // Theoretically overkill but I have trust issues.
+            try {
+                context = JSON.parse(contextContainer.innerHTML);
+            } catch (e) {} // eslint-disable-line no-empty
+
+            return context?.[key] ?? null;
+        }).find(property => property != null);
+
+        return property;
+    }
+
+    /**
+     * @param { Document } dom
+     */
+    async getChapterUrls(dom) {
+        /**
+         * SVG check excludes all locked chapters.
+         * 
+         * @type { NodeListOf<HTMLAnchorElement> }
+         */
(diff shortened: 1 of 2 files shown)


plugin/js/parsers/NoveldexParser.js
```diff
@@ -0,0 +1,202 @@
...
+     * 
+     * @param { Document } dom Source to extract from.
+     * @param { string } key The property to find (first found).
+     * @returns { unknown | null } The found property, or null.
+     * 
+     * @private
+     */
+    extractFromContext(dom, key) {
+        /** @type { NodeListOf<HTMLScriptElement> } */
+        const contextContainers = dom.querySelectorAll(`script[type="application/ld+json"]`); // eslint-disable-line quotes
```
dteviot: This is how you avoid the eslint warning.
```javascript
const contextContainers = dom.querySelectorAll("script[type='application/ld+json']");
```

plugin/js/parsers/NoveldexParser.js
```diff
@@ -0,0 +1,202 @@
...
+    /**
+     * @param { Document } dom
+     */
+    async getChapterUrls(dom) {
+        /**
+         * SVG check excludes all locked chapters.
+         * 
+         * @type { NodeListOf<HTMLAnchorElement> }
+         */
+        let chapterLinksElements = dom.querySelectorAll(`div[id] div[data-state] a[href*="chapter"]:not(:has(svg.lucide-lock))`); // eslint-disable-line quotes
```
dteviot: See previous comment

plugin/js/parsers/NoveldexParser.js
```diff
@@ -0,0 +1,202 @@
...
+             * doesn't work in an all selector when the spans are are nested.
+             * 
+             * @type { NodeListOf<HTMLSpanElement> }
+             */
+            let titleSpans = a.querySelector("div:first-of-type").querySelectorAll("span");
+
+            return {
+                sourceUrl: a.href,
+                title: Array.from(titleSpans, span => span.innerText.trim()).join(" "),
+                newArc: null
```
dteviot: Don't need to add newArc if null

dteviot (changes requested):

plugin/js/parsers/NoveldexParser.js
nitramkh: Done

plugin/js/parsers/NoveldexParser.js
nitramkh: Done

plugin/js/parsers/NoveldexParser.js
nitramkh: Done

dteviot:
@nitramkh 

Thank you for your work

Issue #2395, opened 2026-01-07
Lily on the Valley Blocked Content


JaneDoe-star:
**Describe the bug**
When running WebToEpub application on Lily on the valley, instead of words or even gibberish I get this instead
<img width="1180" height="561" alt="Image" src="https://github.com/user-attachments/assets/5877f416-91cf-4086-ab2a-cfd3dd91e1d7" />

**To Reproduce**
1. Go to ['https://lilyonthevalley.com/story/true-ending-of-a-certain-horror-game/'](https://lilyonthevalley.com/story/true-ending-of-a-certain-horror-game/)

**Desktop (please complete the following information):**
 - OS: Windows [e.g. iOS]
 - Browser: Chrome [e.g. chrome, safari]
 - Version: 1.0.10.11[e.g. 22]

dteviot:
@JaneDoe-star 
Interesting.

Looks like, in order to discourage scraping, the content is no longer in clear text.
Instead, it's encoded into a script element.  (For first chapter of this story, script element has id="tiv-payload-1e000d", next chapter uses "tiv-payload-1a0fb2") and then it replaces the content via the following javascript.
```javascript
  var payloadEl = document.getElementById('tiv-payload-1e000d');
  if (!payloadEl) return;

var data = JSON.parse(payloadEl.textContent || payloadEl.innerText || '{}');
var decodedHTML = decodeURIComponent(atob(data.data || ''));

var container = payloadEl.parentNode;
if (!container) return;

container.innerHTML = decodedHTML;
```

So, adding the following to the FictioneerParser (which is used for LilyOnTheValley) seems to work

```javascript
    preprocessRawDom(chapterDom) {
        let antiScrape = chapterDom.querySelector(".tiv-anti-scrape")?.parentNode;
        if (!antiScrape) return;

        let payloadEl = antiScrape.querySelector("script");
        if (!payloadEl) return;

        let data = JSON.parse(payloadEl.textContent || payloadEl.innerText || "{}");
        let decodedHTML = decodeURIComponent(atob(data.data || ""));
        antiScrape.innerHTML = decodedHTML;
    }
```

Tested with
* https://lilyonthevalley.com/story/true-ending-of-a-certain-horror-game/, chapters 1 and 2
* https://lilyonthevalley.com/story/the-kings-alpha/, chapter 45

However, as this is being done to prevent scraping, I can't add this code to WebToEpub, as I think that would violate Google's ToS and get WebToEpub booted.

[...]

PR #2560, opened 2026-03-22
Add XtruyenParser for xtruyen.vn

meowmereo:
Known Limitations

Chapters not currently visible in any list on the page or not yet visited will fallback to a generic "Chapter X" title format. This is an unavoidable side effect of the site's lazy-loading architecture.

Tip for chapter titles: To capture all specific chapter titles, simply click on the "Danh sách chương" (Chapter List) section on the story's main page and scroll/load the list before launching WebToEpub.

Check List
Do all existing unit tests pass? - Yes, all existing unit tests pass.

Have all warnings and errors reported by eslint been fixed? - Yes.

If you've added new behavior, can a user turn it off? - No the new behavior is encapsulated within this parser, it follows the standard extension architecture. It does not change any global extension behavior without warning.

Have you updated the "contributors" section of package.json with your name? - Yes.

Have you updated the "Credits" section of readme.md with your name? - Yes.

Are you committing to the ExperimentalTab branch? - Yes.

Have you rebased your commit? - Yes, I have performed a rebase against the latest upstream ExperimentalTab branch.


dteviot:
@meowmereo 

There are merge conflicts.  I suggest rebase and squash.

dteviot:
@meowmereo 

I just had a look at your code.  You're decrypting obfuscated text.  WebToEpub isn't allowed to do that.  That's a violation of Google's Terms of Service.

At best, WebToEpub can extract the obfuscated text and put it in the epub, and then users use a different tool to do the decryption.  For example see: https://github.com/dteviot/WebToEpub/issues/2395

meowmereo:
> @meowmereo
> 
> I just had a look at your code. You're decrypting obfuscated text. WebToEpub isn't allowed to do that. That's a violation of Google's Terms of Service.
> 
> At best, WebToEpub can extract the obfuscated text and put it in the epub, and then users use a different tool to do the decryption. For example see: #2395

Thank you for the clarification regarding ToS. To keep this PR compliant, should I modify the xtruyen.vn parser to only extract the raw obfuscated text, or would it be better to close it entirely and only keep the other parser?

PR #2565, opened 2026-03-22
Add LightNovelWorldOrg Parser

### LINKED ISSUE #2540: Please add site https://lightnovelworld.org/
Hello, 

The website mentioned does not fetch the individual chapters, it cannot find the chapters, I believe it could be due to the chapters not being links on the website? So then WebToEpub can't list the chapters when running it's search.  

If you go to the site https://lightnovelworld.org/novel/shadow-slave/chapters/ and right click a chapter (Chapter 1 - 1: Nightmare Begins), there is no "open in new tab" etc. so the chapter is not set as a link but is buried in the page.  

I imagine other websites will be laid out similar to this in the future, if there is a workaround or added feature it would be a huge help.  

If you need more info I can try and provide. 

Thanks

nothing0074:
resolves #2540

### DIFF
--- a/plugin/js/parsers/LightNovelWorldOrgParser.js
+++ b/plugin/js/parsers/LightNovelWorldOrgParser.js
@@ -1,0 +1,80 @@
+"use strict";
+
+// Use one or more of these to specify when the parser is to be used
+
+parserFactory.register("lightnovelworld.org", () => new LightNovelWorldOrgParser());
+
+class LightNovelWorldOrgParser extends Parser {
+    constructor() {
+        super();
+    }
+
+    // eslint-disable-next-line no-unused-vars
+    async getChapterUrls(dom, chapterUrlsUI) {
+        return await this.getChapterUrlsFromApi();
+    }
+
+    async getChapterUrlsFromApi() {
+        const baseUrl = "https://lightnovelworld.org/api/novel/shadow-slave/chapters/";
+        const limit = 500; // max limit (more than 500 just defaults to 500)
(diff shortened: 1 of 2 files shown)


nothing0074:
nvm dont merge this I messed up badly

nothing0074:
ok should work now

plugin/js/parsers/LightNovelWorldOrgParser.js
```diff
@@ -0,0 +1,86 @@
...
+
+    extractAuthor(dom) {
+        let authorLabel = dom.querySelector(".novel-author");
+        let authorText = authorLabel?.textContent?.trim() ?? super.extractAuthor(dom);
+        // Remove "Author: " prefix if present
+        return authorText.replace(/^Author:\s*/i, "");
+    }
+
+    extractLanguage(dom) {
+        return dom.querySelector("html").getAttribute("lang");
```
dteviot, line 67: FYI.  Base class already looks for the language there.  So, this function is not needed in this case.

dteviot:
@nothing0074 

Thank you for your hard work

plugin/js/parsers/LightNovelWorldOrgParser.js
```diff
@@ -0,0 +1,86 @@
+"use strict";
+
+// Use one or more of these to specify when the parser is to be used
```
dteviot, line 3: FYI. This comment is no longer relevant, you've specified one.

PR #2567, opened 2026-03-22
Add TangthuvienParser

meowmereo:
Add TangthuvienParser for tangthuvien.net

### DIFF
--- a/plugin/js/parsers/TangthuvienParser.js
+++ b/plugin/js/parsers/TangthuvienParser.js
@@ -1,0 +1,140 @@
+"use strict";
+
+parserFactory.register("tangthuvien.net", () => new TangThuVienParser());
+
+class TangThuVienParser extends Parser {
+    constructor() {
+        super();
+    }
+
+    async getChapterUrls(dom, chapterUrlsUI) {
+        let chapters = await this.getChapterUrlsFromMultipleTocPages(
+            dom,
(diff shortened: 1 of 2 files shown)


dteviot:
@meowmereo 

Thank you (again)

plugin/js/parsers/TangthuvienParser.js
```diff
@@ -0,0 +1,140 @@
...
+        
+        if (!img) {
+            img = dom.querySelector(".book-img img");
+        }
+
+        if (img && img.src) {
+            return img.src;
+        }
+
+        return util.getFirstImgSrc(dom, "meta[property='og:image']");
```
dteviot, line 124: This isn't right.  getFirstImageSrc will return the url of first image matching the selector.  In this case, the URL is the content attribute of the element.

plugin/js/parsers/TangthuvienParser.js
```diff
@@ -0,0 +1,140 @@
...
+        return urls;
+    }
+
+    static extractPartialChapterList(dom) {
+        let items = [];
+        let listItems = dom.querySelectorAll("li");
+
+        listItems.forEach(li => {
+            if (li.classList.contains("divider-chap")) {
+                items.push({
```
dteviot, line 68: I'm not sure this works.  I think chapters with no sourceUrl are ignored.

PR #2571, opened 2026-03-23
Add TiemtruyenchuParser

meowmereo:
Add TiemtruyenchuParser for tiemtruyenchu.com

### DIFF
--- a/plugin/js/parsers/TiemtruyenchuParser.js
+++ b/plugin/js/parsers/TiemtruyenchuParser.js
@@ -1,0 +1,77 @@
+"use strict";
+
+parserFactory.register("tiemtruyenchu.com", () => new TiemTruyenChuParser());
+
+class TiemTruyenChuParser extends Parser {
+    constructor() {
+        super();
+    }
+
+    async getChapterUrls(dom) {
+        let chapters = [];
+        let addedUrls = new Set();
+        let baseUrl = "https://tiemtruyenchu.com";
+        
+        let links = Array.from(dom.querySelectorAll("a.chapter-item-link"));
+        
+        links.forEach(link => {
+            let href = link.getAttribute("href");
+            if (href) {
+                let fullUrl = new URL(href, baseUrl).href;
+                if (!addedUrls.has(fullUrl)) {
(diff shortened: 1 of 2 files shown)


plugin/js/parsers/TiemtruyenchuParser.js
```diff
@@ -0,0 +1,77 @@
...
+
+    async getChapterUrls(dom) {
+        let chapters = [];
+        let addedUrls = new Set();
+        let baseUrl = "https://tiemtruyenchu.com";
+        
+        let links = Array.from(dom.querySelectorAll("a.chapter-item-link"));
+        
+        links.forEach(link => {
+            let href = link.getAttribute("href");
```
dteviot, line 18: Just use

let href = link.href;

This will give you the full URL, if the link has an href

plugin/js/parsers/TiemtruyenchuParser.js
```diff
@@ -0,0 +1,77 @@
...
+        return authorNode ? authorNode.textContent.trim() : super.extractAuthor(dom);
+    }
+
+    extractDescription(dom) {
+        let descNode = dom.querySelector(".content-text");
+        return descNode ? descNode.innerText.trim() : super.extractDescription(dom);
+    }
+
+    findCoverImageUrl(dom) {
+        let img = dom.querySelector(".story-poster");
```
dteviot, line 54: This seems complicated.  I'll check it tonight

dteviot:
Thank you.

dteviot:
@meowmereo 

Fixes:  https://github.com/dteviot/WebToEpub/commit/500c95c44c1a3b310f8e45016891f31586aa341f

Notes, 
1. As chapters are in order on page, don't need to sort them.
2. WebToEpub will remove any duplicate chapter URLs automatically.
3. Removing unwanted elements should be done in removeUnwantedElementsFromContentElement()
4. util.getFirstImgSrc() will return the src of the first image node that is a child of the supplied CSS.

meowmereo:
Thanks for the fixes! Looks very good.

PR #2575, opened 2026-03-24
chore: correct LumosStoriesParer class name to LumosStoriesParser

Crn0:


dteviot:
@Crn0 

D'oh!  I missed that.
Although for such a minor thing, I'd probably not even bother trying to fix.  But then I'm a lazy S.O.B.

PR #2577, opened 2026-03-25
Add MetruyenchuParser

meowmereo:
Add MetruyenchuParser for metruyenchu.com.vn

### DIFF
--- a/plugin/js/parsers/MetruyenchuParser.js
+++ b/plugin/js/parsers/MetruyenchuParser.js
@@ -1,0 +1,83 @@
+"use strict";
+
+parserFactory.register("metruyenchu.com.vn", () => new MeTruyenChuParser());
+
+class MeTruyenChuParser extends Parser {
+    constructor() {
+        super();
+    }
+
+    async getChapterUrls(dom) {
+        const chapters = [];
+        const bid = dom.querySelector("input[name='bid']")?.value;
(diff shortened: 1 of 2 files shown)


plugin/js/parsers/MetruyenchuParser.js
```diff
@@ -0,0 +1,83 @@
...
+                sourceUrl: new URL(link.getAttribute("href"), baseUrl).href
+            })));
+        }
+        
+        return chapters;
+    }
+
+    extractTitleImpl(dom) {
+        return dom.querySelector("h1[itemprop='name']")?.textContent.trim() || 
+               dom.querySelector(".chapter-title h1 a")?.textContent.trim() ||
```
dteviot, line 57: Is trying both places necessary?  How often is title in one place, but not other?
Also, you don't need trim() here, or textContent.
If you return an element, WebToEpub will use the textContent.

From Template.js
```javascript
    // title of the story  (not to be confused with title of each chapter)
    /*
    extractTitleImpl(dom) {
        // typical implementation is find node with the Title and return name from title
        // NOTE. Can return Title as a string, or an  HTML element
        return dom.querySelector("h1");
    }
    */
```

meowmereo: Ah, thank you, I understood.

PR #2579, opened 2026-03-26
Update TiemtruyenchuParser

meowmereo:
Update TiemtruyenchuParser for tiemtruyenchu.com
- Added detection for locked/VIP chapters and automatically uncheck locked chapters.

### DIFF
--- a/plugin/js/parsers/TiemtruyenchuParser.js
+++ b/plugin/js/parsers/TiemtruyenchuParser.js
@@ -8,11 +8,17 @@ class TiemTruyenChuParser extends Parser {
     }
 
     async getChapterUrls(dom) {
-        return Array.from(dom.querySelectorAll(".chapter-page a.chapter-item-link"))
-            .map(link => ({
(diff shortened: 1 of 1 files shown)


plugin/js/parsers/TiemtruyenchuParser.js
```diff
@@ -8,11 +8,17 @@ class TiemTruyenChuParser extends Parser {
     }
 
     async getChapterUrls(dom) {
-        return Array.from(dom.querySelectorAll(".chapter-page a.chapter-item-link"))
-            .map(link => ({
+        return Array.from(dom.querySelectorAll(".chapter-page a.chapter-item-link")).map(link => {
+            const parent = link.parentElement || link;
```
dteviot, line 12: I can't think of any case where the link won't have a parentElement.

meowmereo: That was just an overly cautious fallback on my part. Feel free to remove the link part directly. Thanks for pointing it out!

PR #2582, opened 2026-03-26
fix: exclude paywalled chapters

Crn0:
Some chapters are getting included even though they're premium. The parser was only checking release dates to determine chapter availability, but ignored LumosStories' paywall system

tested on https://lumostories.com/en/story/512/, https://lumostories.com/en/story/554/ and https://lumostories.com/en/story/50/


Crn0:
~~ah my commit remove the new parser~~

fixed now

dteviot:
Thank you

Issue #2518, opened 2026-03-02
Please add site https://gratuit-4049025.webador.com/finished-novels/my-white-moonlight-took-off-his-women-s-clothes


NescientBystander:
Hi :)

I was wondering if there was a way to add this one? They just link every second chapter so the webup can only pick up half the chapters. Is there any way for epub to pick up the missing ones automatically?

Thank you!

<img width="1306" height="368" alt="Image" src="https://github.com/user-attachments/assets/4fd0c208-dc3c-4384-9974-01346c305b6d" />

PR #2597, opened 2026-04-02
Add https://gratuit-4049025.webador.com/ parser

nitramkh:
Implementation for #2518.

This should be a mostly functional parser for https://gratuit-4049025.webador.com/. The parser implementation is cursed, but that is due to the site being cursed; so I tried to compensate this by adding comments explaining the weirdness in case some poor soul has to try to make changes to it in the future.

There is not a single consistency on the entire site which means 17000 cases have to be handled which resulted in extremely ugly code. Of note is the fact that I added `sketchyHackContentCache` due to how the site works.

Only half the chapters are available on the ToC page, and the rest you can only access by fetching the chapter page, and scanning the next button. This means that it has to fetch 50% of all chapters in `getChapterUrls`, and since the implementation was already so cursed I decided to just cache them. At first I tried to set the `rawDom` property, but that is just completely ignored if it is set there.

So I chose to just hook into `fetchChapter` as I saw little risk in accidentally breaking something else.

I feel like there is a high likelyhood I've done this implementation wrong; so if there are any issues I'll fix them, but not happily as usual because it is cursed ;)

<details>

    1. ⬜ Can you turn it off?
    2. ⬜ Is it off by default?

</details>

### DIFF
--- a/plugin/js/parsers/Gratuit4049025WebadorParser.js
+++ b/plugin/js/parsers/Gratuit4049025WebadorParser.js
@@ -1,0 +1,300 @@
+"use strict";
+
+parserFactory.register("gratuit-4049025.webador.com", () => new Gratuit4049025WebadorParser());
+
+class Gratuit4049025WebadorParser extends Parser {
+    constructor() {
+        super();
+    }
+
+    /**
+     * Content cache: Caches chapter content fetched when finding chapter URLs.
+     *                Reduces fetches by literally 50%.
+     * 
+     * Hack:          The right way to do this would probably be to make a
+     *                modification allowing `getChapterUrls` to pass `rawDom`.
+     * 
+     * Sketchy:       Because I don't know the codebase well enough to be
+     *                certain this won't have side-effects.
+     * 
+     * @type { Map<string, Document> }
+     * @private
+     */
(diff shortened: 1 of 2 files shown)


plugin/js/parsers/Gratuit4049025WebadorParser.js
```diff
@@ -0,0 +1,300 @@
...
+        return Array.from(candidates, p => /.*?(?:G|g)enre\s*:(.*)/.exec(p.textContent.trim())?.[1])
+            .find(genres => !util.isNullOrEmpty(genres))?.trim();
+    }
+
+    /**
+     * @override
+     * @param { Document } dom 
+     */
+    extractLanguage(dom) {
+        return dom.querySelector("html")?.getAttribute("lang");
```
dteviot, line 272: If this is where language is, you don't need this code.  Base implementation does this.

dteviot:
@nitramkh 
I only skimmed the code.

dteviot:
@nitramkh 
D'oh!  https://github.com/dteviot/WebToEpub/commit/6874ce575da9b9c3adf71308c95dbca3ab36c0a0

Issue #2501, opened 2026-02-25
Please add site https://fictionread.xyz


pavan3999:
Please note, there are only a few developers working on WebToEpub, and we're not paid for doing this.  (WebToEpub is completely free, and generates no money.) By asking to add a site, you're asking us to give you some of our limited free time. So, we think it's not unreasonable for me to ask you to do as much as you can to help us.

### Check the site has not already been requested.
You can search the issues to see if any have the site name by putting the site name in the search box near the top of the screen.  See https://docs.github.com/en/issues/tracking-your-work-with-issues/using-issues/filtering-and-searching-issues-and-pull-requests.
If there is an existing issue, you can add a comment to the issue to indicate you're also interested in the site.

### Provide URL for web page that contains Table of Contents (list of chapters) of a typical story on the site
https://fictionread.xyz/novel/679cb098710a480a206b7a4e

### Did you try using the Default Parser for the site?  If not, why not?
Instructions for using the default parser can be found at https://github.com/dteviot/WebToEpub/wiki/FAQ#how-to-convert-a-new-site-using-the-default-parser

### What settings did you use?  What didn't work?
* URL of first chapter: https://fictionread.xyz/chapter/679e623f0d96b7f82a65a411?n=679cb098710a480a206b7a4e 
* CSS selector for element holding content to put into EPUB
* CSS selector for element holding Title of Chapter
* CSS selector for element(s) to remove

### If the Default Parser did not work, if you have developer skills, did you try writing a new parser?
Instructions https://github.com/dteviot/WebToEpub/wiki/FAQ#how-to-write-a-new-parser

nitramkh:
Okay, so I've looked into this one and based on my understanding of how the extension works this site cannot be supported. To fetch the chapter content you need to make a `POST` request with a specific NextJS header. This is no issue; except for the fact that **if you provide an Origin**, it must be `fictionread.xyz` or you get a 500. If you pass no origin, it works. The problem is that from my understanding it is not possible via fetch/XMLHttpRequest pass both a "custom" request header, and no origin. So in this case that means you either pass both, or you can't pass the custom NextJS header.

I've tested this by making a dead simple post request with only a single header, the NextJS one, `NextAction: d1519557e507ca98f7d048353337ed88be599264` (the value is magic), and it works flawlessly. But if you pass an Origin which is anything other than `fictionread.xyz` its 500 time.

[...]

PR #2612, opened 2026-04-08
Add https://fictionread.xyz/ parser

nitramkh:
Implementation for #2501.

# Parser
Simple parser for https://fictionread.xyz/, with only two things of note.

First the cover images the site uses come directly from a novelupdates.com CDN. I think I remember reading somewhere that they don't want stuff taken from them? Is this a problem?

Second, premium chapters are completely unsupported by this implementation. Their premium paywall implementation is extremely sloppy and does the **authentication client-side**. So if you try to fetch the chapter content, it just gives it to you; and the only way to check if the user actually has unlocked it, is by knowing their username and email and shoving it into the request body. But that is not available in the DOM.

So instead I just blocked it out entirely. I don't really care, IMO it's their own fault; but I figure the corporate overlords wouldn't like "enabling" this type of thing?

# Decoder
The decoded is for the React Flight Protocol, which is used for their RSC stuff. I kept seeing the format seemingly everywhere, so I decided to do some learning by doing and make a parser (which I called decoder because it's kinda a busy keyword here ;)).

The protocol is their custom way of transferring information between client and server, mainly designed to allow for React Server Components, I.E. partial server side "rendering". But it's also used to as the standard information transfer protocol for data; such as chapter information in this case. Or as another example, noveldex, keeps most of it's metadata and chapter content inside script tags with blobs in this format.

Frankly, its probably a majority comments; because JSDOC is quite possibly the most verbose thing in existence. It parses the basic elements of the format into simple objects which can then be accessed without the need for regex. It's not fully faithful to the "real" parser/spec as it's intended for quite a different use-case, but it is built for purpose, to extract information as easily as possible; but naturally doesn't have the full capabilities of the protocol as that is irrelevant.

As I said I made it mostly to figure out what the weird format was, so I don't know if it's actually needed but I figure I would offer it anyway; it does have the effect of making the call-sites look quite a bit nicer in my opinion (especially if it got a custom `ResponseHandler`). There are a quite a few improvements I would want to make, but I don't wanna spend that time if it won't get merged ;)

But as always, if there are any issues or wanted changes, I'm happy to fix; and yes that includes ripping out the entire Decoder and replacing it with a bunch of regex ;)

<details>

    1. ⬜ Can you turn it off?
    2. ⬜ Is it off by default?

</details>

### DIFF
--- a/plugin/js/parsers/FictionreadParser.js
+++ b/plugin/js/parsers/FictionreadParser.js
@@ -1,0 +1,1278 @@
+"use strict";
+
+/**
+ * NOTE: To add additional hostname/site support hard-coded hostnames used in
+ *       implementation must also be fixed.
+ */
+parserFactory.register("fictionread.xyz", () => new FictionreadParser());
+
+/**
+ * Parser for the https://fictionread.xyz/ site.
+ */
+class FictionreadParser extends Parser { // eslint-disable-line no-unused-vars
+    constructor() {
+        super();
+    }
+
+    /**
+     * @private
+     */
+    removeOriginFromWebRequests = (() => {
+        let done = false;
+
+        return async () => {
+            if (done)
+                return;
+
+            done = true;
+
+            /**
+             * Don't pass an origin header, because API fetches will fail if you
+             * pass anything other than "fictionread.xyz" or nothing.
+             * 
+             * @type { chrome.declarativeNetRequest.Rule[] }
+             */
+            const fetchRules = [
+                {
+                    "id": 1,
+                    "priority": 1,
+                    "condition": {
+                        "urlFilter": "fictionread.xyz",
+                    },
+                    "action": {
+                        "type": "modifyHeaders",
+                        "requestHeaders": [
+                            {
+                                "header": "origin",
+                                "operation": "remove",
+                            },
+                        ]
+                    }
+                }
+            ];
+
+            await HttpClient.setDeclarativeNetRequestRules(fetchRules);
+        };
+    })();
+
+    /**
+     * @param { Document } dom
+     * @param { string } url 
+     * @param { string } body
+     * @param { HeadersInit } headers
+     * @private
+     */
+    async fetchFlightEndpoint(dom, url, body, headers) {
+        this.removeOriginFromWebRequests();
+
+        // Set header defaults, provided have priority.
+        headers = {
+            ...{
+                //"x-from": "https://fictionread.xyz/"
+            },
+            ...headers
+        };
+
+        /** @type { RequestInit } */
+        let fetchOptions = {
+            method: "POST",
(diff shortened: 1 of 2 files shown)


dteviot:
Something this big is going to take me a while to inspect.  Not going to get it done tonight.

Also, thanks for your efforts.

nitramkh:
> Something this big is going to take me a while to inspect. Not going to get it done tonight.

That is only to be expected ;)

EDIT: I accidentally hit the update branch button which I'd never seen before; and thus discovered that I'd forgotten to fetch upstream before rebasing, sigh... I'll see if I can figure out how to rebase that away.

plugin/js/parsers/FictionreadParser.js
```diff
@@ -0,0 +1,1278 @@
...
+
+        return payload;
+    }
+
+    /**
+     * @override
+     * @param { Document } dom 
+     */
+    async getChapterUrls(dom) {
+        let novelID = /(?<=novel\/)(?<novelID>[^/?\s]+)(?=\/|\?|$)/.exec(dom.baseURI)?.groups?.["novelID"];
```
Kiradien, line 98: Just adding a comment here since this is a bit more advanced utilization of RegEx than we normally see. 

The advanced form used here is better because it both names a group for better accessibility should the regex change, and it restricts the actual full result of the regex to just the content of the group (e.g. the 0 result would match the novelID result.) This is because of Positive Lookaheads and Positive Lookbehinds - they're pretty useful. 
While not actually necessary when you have access to more complex features such as groups, advanced regex users tend to use them habitually to stay in practice for Negative Lookahead/Lookbehinds since those are substantially harder to replicate.

Fundamentally, the above isn't different from:
`let novelID = /(?:novel\/)([^/?\s]+)(?:\/|\?|$)/.exec(dom.baseURI)?.groups?.[1];`

Breakdown of the simplified version:
(?:novel\/) - Unsaved group (?:), holds identifiers for the part of url to scrape for ID
([^/?\s]+) - Saved group, holds the ID information we actually want.
(?:\/|\?|$)/ - Unsaved group, holds acceptable characters following the ID above.

plugin/js/parsers/FictionreadParser.js
```diff
@@ -0,0 +1,1278 @@
...
+     * @param { Document } dom 
+     */
+    findCoverImageUrl(dom) {
+        /** @type { HTMLImageElement | null } */
+        const cover = dom.querySelector("img[alt*='main']");
+
+        if (!cover)
+            return null;
+
+        const url = new URL(cover?.src);
```
dteviot, line 211: cover can't be null

dteviot:
@nitramkh 

Thanks, that's a big commit.

Issue #2525, opened 2026-03-06
https://101kks.com/


steve30990:
https://101kks.com/

Site has not been Requested a fairly clean site.
Just the top and bottom of each chapter need remove.
= it be like 69shu in term of clean site, and Format similar to 69shu.

### Provide URL for web page that contains Table of Contents (list of chapters) of a typical story on the site

https://101kks.com/book/4801/index.html

### Did you try using the Default Parser for the site?  If not, why not?

I dont know how, and dont have anyone to help.

Please help Thanks

sakthi719:
@steve30990 

<img width="1070" height="768" alt="Image" src="https://github.com/user-attachments/assets/0cb75c68-9a43-42c0-b09e-525182fdbc7e" />

dteviot:
@steve30990 

Thanks to @nitramkh Test versions for Firefox and Chrome have been uploaded to https://github.com/dteviot/WebToEpub/releases/tag/developer-build. Pick the one suitable for you, follow the "How to install from Source (for people who are not developers)" instructions at https://github.com/dteviot/WebToEpub/tree/ExperimentalTabMode#user-content-installation and let us know how it goes.

dteviot:
Updated version (1.0.13.0) has been submitted to Firefox and Chrome stores.
Firefox version should be available within two weeks (but might take 4 to 6 weeks)
Chrome might be available in a few hours to 21 days.  (Typically 3 to 7 days).

PR #2614, opened 2026-04-08
Add https://101kks.com/ parser

nitramkh:
Implementation for #2525.

I think everything seems fine, but I don't read Chinese so I can't be 100% sure. Only thing is some chapters will have a duplicated title, as they have the standardized title found by `findChapterTitle`; and the same title in the content body. But the one in the body is not standardized and only some chapters have it, and I have no idea how to write a regex to match chinese characters.

As always, if there are issues or wanted changes, I'm happy to fix ;)

<details>

    1. ⬜ Can you turn it off?
    2. ⬜ Is it off by default?

</details>

### DIFF
--- a/plugin/js/parsers/101kksParser.js
+++ b/plugin/js/parsers/101kksParser.js
@@ -1,0 +1,126 @@
+"use strict";
+
+parserFactory.register("101kks.com", () => new _101kksParser());
+
+class _101kksParser extends Parser {
+    constructor() {
+        super();
+    }
+
+    setRefererOnWebRequests = (() => {
+        let done = false;
+
+        return async () => {
+            if (done)
+                return;
+
+            done = true;
+
+            /**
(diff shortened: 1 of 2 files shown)


plugin/js/parsers/101kksParser.js
```diff
@@ -0,0 +1,126 @@
...
+    findChapterTitle(dom) {
+        return dom.querySelector(".txtnav > h1");
+    }
+
+    /**
+     * @override
+     * @param { Document } dom 
+     */
+    extractAuthor(dom) {
+        return dom.querySelector("div.booknav2 > p a[href*='author']")?.textContent.trim() ?? super.extractAuthor();
```
dteviot, line 99: You're probably over specifying.  I suspect the CSS just needs to be "div.booknav2 a[href*='author']"

Issue #2526, opened 2026-03-06
https://www.twbook.cc/


steve30990:
https://www.twbook.cc/

Site has not been Requested.
Similar to Novel543
Each chapter can be divided into 2 Links, but mostly is One chapter per link.

Ex: seen like chapter 1 divided into chapter 1 part 1/2, chapter 1 part 2/2.

Provide URL for web page that contains Table of Contents (list of chapters) of a typical story on the site

https://www.twbook.cc/1411584374/dir

Did you try using the Default Parser for the site? If not, why not?

I dont know how, and dont have anyone to help.

Please help Thanks

dteviot:
@steve30990 

Thanks to @nitramkh, test versions for Firefox and Chrome have been uploaded to https://github.com/dteviot/WebToEpub/releases/tag/developer-build. Pick the one suitable for you, follow the "How to install from Source (for people who are not developers)" instructions at https://github.com/dteviot/WebToEpub/tree/ExperimentalTabMode#user-content-installation and let us know how it goes.

dteviot:
Updated version (1.0.13.0) has been submitted to Firefox and Chrome stores.
Firefox version should be available within two weeks (but might take 4 to 6 weeks)
Chrome might be available in a few hours to 21 days.  (Typically 3 to 7 days).

PR #2615, opened 2026-04-08
Add https://twbook.cc/ to Novel543Parser

nitramkh:
Implementation for #2526.

* https://twbook.cc/ is a clone of https://novel543.com/ and thus just hooks into [Novel543Parser](https://github.com/dteviot/WebToEpub/blob/ExperimentalTabMode/plugin/js/parsers/Novel543Parser.js).

* I added `extractSubject` to get the genre. It is the same across both sites so I saw no reason not to.

* I JSDOC-ified it while I was there because why not.

As always, if there are any issue or wanted changed, I'm happy to fix ;)

<details>

    1. ⬜ Can you turn it off?
    2. ⬜ Is it off by default?
8. ✅ Have you ensured you committed all relevant changes?
9. ✅ Have you rebased your commits?
10. ✅ Did you fetch from upstream before rebasing?

</details>

### DIFF
--- a/plugin/js/parsers/Novel543Parser.js
+++ b/plugin/js/parsers/Novel543Parser.js
@@ -1,6 +1,7 @@
 "use strict";
 
 parserFactory.register("novel543.com", () => new Novel543Parser());
+parserFactory.register("twbook.cc", () => new Novel543Parser());
 
 class Novel543Parser extends Parser {
     constructor() {
@@ -30,6 +31,10 @@ class Novel543Parser extends Parser {
         return authorLabel?.textContent ?? super.extractAuthor(dom);
     }
 
+    extractSubject(dom) {
+        return dom.querySelector("p.meta a[href*='bookstack']")?.textContent?.trim();
+    }
+
     extractLanguage() {
         return "zh";
     }
@@ -42,8 +47,17 @@ class Novel543Parser extends Parser {
         return this.walkPagesOfChapter(url, this.moreChapterTextUrl);
     }
 
+    /**
+     * @param { Document } dom 
+     * @param { string } baseUrl 
+     * @private
+     */
     moreChapterTextUrl(dom, baseUrl) {
-        // Extract chapter base from original URL (e.g., "8096_1" from "8096_1.html" or "8096_1_2.html")
+        /**
+         * Extract chapter base from original URL (e.g., "8096_1" from "8096_1.html" or "8096_1_2.html")
+         * 
+         * @param { string  } url 
+         */
         let getChapterBase = (url) => {
             let match = url.match(/\/(\d+_\d+)(?:_\d+)?\.html/);
             return match ? match[1] : null;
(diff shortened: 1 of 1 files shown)


dteviot:
@nitramkh 
> I JSDOC-ified it while I was there because why not.

Please don't do that for functions that are just overrides of functions in the base parser.
Only add comments for new functions, where what it does might not be obvious from the name.
e.g. **Why** does moreChapterTextUrl() exist?  (This one could use a comment.)
FYI. https://stackoverflow.blog/2021/12/23/best-practices-for-writing-code-comments/

nitramkh:
@dteviot 

> Please don't do that for functions that are just overrides of functions in the base parser.

Fair enough, I'll remove them, and later I can go through and remove them from the other parsers where I've put them; as I assume the same applies there? I just like adding them to get typescript hinting working right as I'm a bit of a dumbass and make little mistakes constantly. Also does this apply to JSDOC @ type annotation for example in the code too?

Regardless, shouldn't be too onerous ;)

dteviot:
@nitramkh 

> I'll remove them, and later I can go through and remove them from the other parsers where I've put them; as I assume the same applies there?

I'm OK if you choose not to fix up the older stuff.  It's a minor thing.

>  I just like adding them to get typescript hinting working right as

I'd also like to use typescript.  Unfortunately,  using it complicates getting WebToEpub approved, due to code being transpiled.  Which make the inspection/audit longer and more complcated.

nitramkh:
@dteviot 

> I'd also like to use typescript. Unfortunately, using it complicates getting WebToEpub approved, due to code being transpiled. Which make the inspection/audit longer and more complcated.

Sure, but in this case I meant specifically using the JSDOC type annotations to get Typescript type-checking in JavaScript. The support is pretty good at this point, I've been using this the entire time. I have a private branch where I've been exploring the codebase and trying to learn it where I've just been adding type notations as I went; and its quite helpful.

Like for example #2591, which I only noticed because Typescript was screaming due to the missing return.

EDIT: For convenience I figure I can drop the spec link here [Typescript JSDOC reference](https://www.typescriptlang.org/docs/handbook/jsdoc-supported-types.html). If it wasn't for this support, I wouldn't have been adding the type annotations either as they are verbose and ugly af, but it's quite helpful.

Also, I wasn't planning on ever making a PR for this as I know I am a comment-aholic and thus that it's unlikely anyone else would want my comment stew. But if you want a demo of what I mean here is the branch I've been playing around in when trying to learn about the codebase, [nitramkh:personal/jsdocifying](https://github.com/nitramkh/WebToEpub/tree/personal/jsdocifying). As a beginner it was much less "scary" using ex. `HttpClient` having the types, and not having to blindly imitate existing usages and pray or keep the entire thing in memory (which I am completely incabable of anyway ;)).

dteviot:
@nitramkh 

Thanks for your efforts.

Issue #2527, opened 2026-03-06
https://uukanshu.cc/


steve30990:
### Check the site has not already been requested.
Site has Not Been Requested.
https://uukanshu.cc/

### Provide URL for web page that contains Table of Contents (list of chapters) of a typical story on the site

https://uukanshu.cc/book/17318/

### Did you try using the Default Parser for the site?  If not, why not?

I dont know how, and dont have anyone that can help.

Please help Thanks

sakthi719:
@steve30990 

<img width="1069" height="766" alt="Image" src="https://github.com/user-attachments/assets/11c8bd11-8bef-43ca-96f3-70725b2af293" />

dteviot:
@steve30990 @sakthi719 

Thanks to @nitramkh Test versions for Firefox and Chrome have been uploaded to https://github.com/dteviot/WebToEpub/releases/tag/developer-build. Pick the one suitable for you, follow the "How to install from Source (for people who are not developers)" instructions at https://github.com/dteviot/WebToEpub/tree/ExperimentalTabMode#user-content-installation and let us know how it goes.

dteviot:
Updated version (1.0.13.0) has been submitted to Firefox and Chrome stores.
Firefox version should be available within two weeks (but might take 4 to 6 weeks)
Chrome might be available in a few hours to 21 days.  (Typically 3 to 7 days).

PR #2616, opened 2026-04-08
Add https://uukanshu.cc/ parser

nitramkh:
Implementation for #2527.

Again, I don't read Chinese but I think it works. It is very similar to [MtnovelParser]() but not exactly, and it seems to be down so I can't check to see if I broke it if I were to merge them. So instead I marked it as dead with a timestamp to make it a future problem.

As always, if there are any issues or wanted changes, I'm happy to fix ;)

<details>

    1. ⬜ Can you turn it off?
    2. ⬜ Is it off by default?
8. ✅ Have you ensured you committed all relevant changes?
9. ✅ Have you rebased your commits?
10. ✅ Did you fetch from upstream before rebasing?

</details>

### DIFF
--- a/plugin/js/parsers/MtnovelParser.js
+++ b/plugin/js/parsers/MtnovelParser.js
@@ -1,5 +1,6 @@
 "use strict";
 
+// Dead site (2026-04-08)
 parserFactory.register("mtnovel.net", () => new MtnovelParser());
 
 class MtnovelParser extends Parser {
--- a/plugin/js/parsers/UukanshuParser.js
+++ b/plugin/js/parsers/UukanshuParser.js
@@ -1,0 +1,98 @@
+"use strict";
+
+parserFactory.register("uukanshu.cc", () => new UukanshuParser());
+
+class UukanshuParser extends Parser {
+    constructor() {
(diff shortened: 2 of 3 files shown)


plugin/js/parsers/MtnovelParser.js
```diff
@@ -1,5 +1,6 @@
 "use strict";
 
+// Dead site (2026-04-08)
```
dteviot, line 3: This is the old way of doing this.  Recommended way now is registerDeadSite()

plugin/js/parsers/UukanshuParser.js
```diff
@@ -0,0 +1,98 @@
...
+        return dom.querySelector("h1.pt10");
+    }
+
+    /**
+     * @override
+     * @param { Document } dom 
+     */
+    extractTitleImpl(dom) {
+        /** @type { HTMLMetaElement | null } */
+        const meta = dom.querySelector("meta[property='og:title");
```
dteviot, line 43: try

return dom.querySelector("meta[property='og:title")?.content;

You can do similar elsewhere

PR #2631, opened 2026-04-19
Add site https://beastnovels.com

crybx:
Parser for https://beastnovels.com

### DIFF
--- a/plugin/js/parsers/BeastnovelsParser.js
+++ b/plugin/js/parsers/BeastnovelsParser.js
@@ -1,0 +1,85 @@
+"use strict";
+
+parserFactory.register("beastnovels.com", () => new BeastnovelsParser());
+
+class BeastnovelsParser extends Parser {
+    constructor() {
+        super();
+    }
+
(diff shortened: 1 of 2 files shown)


plugin/js/parsers/BeastnovelsParser.js
```diff
@@ -0,0 +1,85 @@
...
+    constructor() {
+        super();
+    }
+
+    async getChapterUrls(dom) {
+        const removeNum = document.getElementById("removeChapterNumberCheckbox").checked;
+        const cards = [...dom.querySelectorAll("a.chapter-card")]
+            .filter(a => a.dataset.premium !== "1");
+        return cards.map(a => {
+            const num = a.querySelector(".chapter-card__num")?.textContent.replace(/\s+/g, " ").trim();
```
dteviot, line 15: If you use "?." you should continue to use it.  Otherwise it will just break at the next operation

dteviot:
@crybx 

Thank you.

PR #2649, opened 2026-05-01
Fix Ranobes chapter title fallback

ltmerletti:
Sometimes the Ranobes reader pages put the chapter title inside `div#arrticle` instead of `h1.title`. This fix checks the existing `h1.title` selector first, then falls back to article headings, so normal Ranobes pages keep their current behavior. Without this, the download would never happen, due to a crash (from the failure parsing).


dteviot:
@ltmerletti 

Thank you.  You should have added yourself to the credits, but I'll take care of that.

PR #2656, opened 2026-05-05
Add site https://oceanstale.com

crybx:
New parser for https://oceanstale.com

Also, regarding TruyenFullVisionParser.js, I update line endings from CRLF to LF. See https://github.com/dteviot/WebToEpub/pull/1937 for a more thorough explanation of why line endings in the repo ought to be (and usually are) LF. The person who created that parser must have had 
 
`git config core.autocrlf false` 

as their git line ending settings for this to have happened. It's not that big of a deal, but does show up as a diff for cross OS dev, so I updated.


dteviot:
@crybx 

Thanks.

Issue #155, opened 2018-01-29
Metadata dc:date and dc:publisher


GallusMax:
I like to suggest scraping for two more fields describing a publication besides author and title.
Some sources clearly can deliver this, so why should we let it slip away?

Enhancing the EpubMetaInfo() class with the appropriate fields plus default methods in the Parser class should do the job.

The metadata should then show up in the generated epub file.

dteviot:
I'm already using dc:date to hold the date the epub is created.
You're welcome to add publisher if you wished

PR #2662, opened 2026-05-09
Add Publisher Metadata to Epub

Yomafil:
Added `dc:publisher` metadata when building an EPUB.
#155

Default implementation is getting the publisher based on the website name from metadata: `og:site_name`.
Else we can use `extractPublisher()` in a Parser to get the name of the publisher as a string or from some elements.
If not implemented, default to ``.

Added input field in `Advanced Options` > `Show more Metadata options`.

Tested and working with Calibre.
Tested and working with Extension Library.

Example of site with `og:site_name`: 
- https://www.webnovel.com/book/35044944208489005
- https://www.scribblehub.com/series/1325911/bofuri-the-strongest-shield-of-tensura/

Example Parser implementation:
- https://novelbin.com/b/i-arrived-at-wizard-world-while-cultivating-immortality

### DIFF
--- a/plugin/js/Parser.js
+++ b/plugin/js/Parser.js
@@ -308,6 +308,20 @@ class Parser {
     }
 
     /**
+    * default implementation, 
+    * if not available, return ''
+    */
+    extractPublisher(dom) {   // eslint-disable-line no-unused-vars
+        // try metadata extraction
+        let publisher = dom.querySelector("meta[property='og:site_name']");
+        if (publisher !== null) {
+            return publisher.content;
+        }
+
+        return "";
+    }
+
+    /**
     * default implementation, Derived classes will override
     */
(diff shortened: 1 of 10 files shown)


plugin/js/Parser.js
```diff
@@ -307,6 +307,20 @@ class Parser {
...
     }
 
+    /**
+    * default implementation, 
+    * if not available, return ''
+    */
+    extractPublisher(dom) {   // eslint-disable-line no-unused-vars
+        // try metadata extraction
+        let publisher = dom.querySelector("meta[property='og:site_name']");
+        if (publisher !== null) {
```
dteviot, line 317: try
```javascript
return publisher?.content ?? "";
```

dteviot:
FYI, did small cleanup.

PR #2687, opened 2026-05-23
Added sbxh parser

Harimz:
Adds support for sbxh novel pages


dteviot:
@Harimz 

That's an impressive piece of work.
I especially like the bit where you switch the tab and scrape the text from it.
Unfortunately, you're also decrypting encrypted content.  And that violates Google's rules for an extension.
Basically point 6 here https://developer.chrome.com/docs/webstore/program-policies/malicious-and-prohibited.
I'm assuming the content is encrypted to prevent copying.
Note, if content was encrypted to prevent it being scraped by AI, that's a different thing, but you'd need to show me the evidence.

I will move your parser into the "experimental" folder under parsers.  Then it will be available for any users who build it themselves, but it won't be included in the version of WebToEpub that is published to the stores.  Sorry, that's about the best I can do.

PR #2703, opened 2026-06-02
Add support for Althistory.com

l3gacyb3ta:


dteviot:
@l3gacyb3ta 

You've basically copy/pasted the SpaceBattles parser.
Recommended way would be to derive from SpaceBattles.  e.g.

Add this to top of the Spacebattles.js
```javascript
parserFactory.register("althistory.com", () => new AlthistoryParser());
```

And add this to bottom
```javascript
class AlthistoryParser extends SpacebattlesParser {
    constructor() {
        super();
    }

    async getChapterUrls(dom) {
        let chapters = [...dom.querySelectorAll(".threadmark_depth0 > ul > li > a")];
        return chapters.map(a => util.hyperLinkToChapter(a));
    }

    extractTitleImpl(dom) {
        return dom.querySelector(".threadmarkLabel");
    }

    async fetchArticle(url) {
        let fetchedDom = await this.cache.fetch(url);
        let newUrl = new URL(url);
        let id = newUrl.hash.substring(1);
        let parent = fetchedDom.querySelector(`article.message[data-content='${id}']`);
        if (parent === null)
        {
            parent = fetchedDom.querySelector("#" + id)?.parentElement;
        }
        return parent;
    }
}
```

dteviot:
@l3gacyb3ta 

And, I get a 403 error when try to fetch chapter from site.

PR #2727, opened 2026-06-14
feat: add readrift.net parser

matthewsong18:
Adding a parser for [ReadRift](https://readrift.net). It is fully working for multiple different novels on the site. I was able to figure out the API and the parser is able to collect all novel metadata and chapter content directly from the API.

* Note for the maintainer that might be testing this: I frequently ran into issues where the website was timing out. But if I cleared cookies and used a vpn, it would immediately connect and load. This might be a site that rate limits by dropping the connection without a response.
* I've added a minimum throttle of 3000 ms and a 300 ms delay between table of content page requests.

## Pull Request Check List

- [ ] Do all existing unit tests pass?  (FYI, I find it easier to get the unit tests to run under Firefox, instead of Chrome.) **No, the MuggleNet test was failing on the latest dev branch before I made changes. I ran a bisect, but it's failing on every commit, so it is probably my environment. However, all other tests are passing after my changes.**

- [x] Have all warnings and errors reported by eslint been fixed? **Yes**

- [x] Have you updated the "contributors" section of packages.json with your name? (and ideally email, so I can contact you easily.) **Should be there from prior contributions**

- [x] Have you updated the "Credits" section of readme.md with your name? **Same as above**

- [x] Are you committing to the ExperimentalTab branch? **Yes**

- [x] Have you rebased your commit? **Yes**

### DIFF
--- a/plugin/js/parsers/ReadRiftParser.js
+++ b/plugin/js/parsers/ReadRiftParser.js
@@ -1,0 +1,260 @@
+/*
+  Template to use to create a new parser
+*/
+"use strict";
+
+// Use one or more of these to specify when the parser is to be used
+/*
+// Use this function if site's host name is sufficient.  
+// i.e. All pages are on same site, and use same format.
+parserFactory.register("template.org", () => new TemplateParser());
+
+// Use this function if site's URL is sufficient
+parserFactory.registerUrlRule(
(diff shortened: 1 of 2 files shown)


dteviot:
@matthewsong18 

PR #2780, opened 2026-07-05
feat(parser): add transcendentaltls

matthewsong18:
- All tests passing
- Site uses CSR so APIs were used for table of contents and chapter content collection
- Tested on linux firefox


dteviot:
Thank you.

PR #2784, opened 2026-07-08
Update AliceswParser

Yomafil:
Added error handling for captcha and rate limit challenge pages.
Tested on: https://www.alicesw.com/novel/52192.html
Approximately a Captcha every 60 ch, and Rate Limit every 3 Captcha (IP Locked, it seems). 
Changing the throttle above 3sec doesn't seem to chance the amount of chapter it takes to get a captcha, nor avoid rate limit.

### DIFF
--- a/plugin/js/parsers/AliceswParser.js
+++ b/plugin/js/parsers/AliceswParser.js
@@ -5,7 +5,7 @@ parserFactory.register("alicesw.com", () => new AliceswParser());
 class AliceswParser extends Parser {
     constructor() {
         super();
-        this.minimumThrottle = 3000;
+        this.minimumThrottle = 5000;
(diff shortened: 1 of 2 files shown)


Yomafil:
It seems 5sec would be better than 3sec, I don't seem to have hit IP timeout rate limits yet, only captcha. 
Still testing it for now, will commit the change when it's confirmed.

Yomafil:
Can confirm after testing, 5 sec seems to be the sweet spot. I managed to download this [book](https://www.alicesw.com/novel/44521.html) (600 ch) without hitting major rate limits (IP Timeout) only captcha.

dteviot:
Thank you.

PR #2786, opened 2026-07-08
Fix RoyalRoad parser for the site's Tailwind redesign

possibletoactual:
Royal Road's site redesign broke `RoyalRoadParser`: on the new markup the chapter list silently collapses to a single chapter, and author/cover/description/genres extraction fails. This restores the parser against the new markup while keeping the old selectors as fallbacks.

### What changed
- **Chapter list** — parse the `window.chapters` JSON array (present in every page variant) instead of the removed `table#chapters`; fall back to the legacy table.
- **Chapter content** — anchor on `div.chapter-inner` (the `portlet-body` wrapper is gone) and null-guard so a single odd chapter can't throw and abort the whole download.
- **Chapter titles** — use the authoritative title from the chapter list instead of the first `<h1>` (now an unrelated "Give Reputation to User" widget).
- **Author** — header `/profile/` link → `books:author` meta → legacy selector.
- **Cover** — `og:image` / `twitter:image` → legacy `img.thumbnail`.
- **Description / genres / information page** — the served DOM has scripts stripped (no JSON-LD) and `og:description` is truncated, so the full description is scraped from the raw page during analyse; genres read from `a[href*='tagsAdd=']`. The information page is rebuilt with title, author, genres, and full description.

### PR checklist
- **Unit tests**: pass — existing `UtestRoyalRoadParser` covers `removeOlderChapterNavJunk` / `makeHiddenElementsVisible`, which are unchanged.
- **eslint**: clean.
- **New behavior**: none — this restores existing behavior broken by the site change, so there's no new user-facing toggle.
- **Contributors / Credits**: added to `package.json` and `readme.md`.
- **Base branch**: `ExperimentalTabMode`.

Legacy selectors are preserved as fallbacks, so older/cached page variants continue to work.

### DIFF
--- a/plugin/js/parsers/RoyalRoadParser.js
+++ b/plugin/js/parsers/RoyalRoadParser.js
@@ -14,28 +14,74 @@ class RoyalRoadParser extends Parser {
     async getChapterUrls(dom) {
         // Page in browser has links reduced to "Number of links to show"
         // Fetch new page to get all chapter links.
-        let tocHtml = (await HttpClient.wrapFetch(dom.baseURI)).responseXML;
-        let table = tocHtml.querySelector("table#chapters");
+        //
+        // Royal Road's markup varies by session/redesign: the old server-rendered
+        // <table id="chapters">, a new Tailwind table (class "w-full divide-y", no id),
+        // or a JS-rendered shell with an empty <body>. The one constant across every
+        // variant is the full chapter list embedded as a `window.chapters = [...]`
+        // JSON array (each entry has id/title/slug/url). Parse that first
+        // (redesign-proof); fall back to scraping the old table for legacy pages.
+        let tocHtml = await HttpClient.fetchText(dom.baseURI);
+        // Capture book metadata (description + genres) from the RAW html now — the DOM
+        // the parser later sees has its <script> tags stripped and, on the redesign, no
+        // JSON-LD at all, so the full description must be scraped here.
+        RoyalRoadParser.lastBookMeta = RoyalRoadParser.extractBookMeta(tocHtml);
+        let fromJson = RoyalRoadParser.chaptersFromWindowChapters(tocHtml, dom.baseURI);
+        if (0 < fromJson.length) {
+            return fromJson;
+        }
+        // fallback: legacy server-rendered table
+        let doc = new DOMParser().parseFromString(tocHtml, "text/html");
+        let table = doc.querySelector("table#chapters");
         return util.hyperlinksToChapterList(table);
+    }
+
+    // Extract the chapter list from the page's `window.chapters = [...]` JSON blob.
+    // Returns [] if not present/parseable, so the caller can fall back.
+    static chaptersFromWindowChapters(html, baseUrl) {
+        let match = html.match(/window\.chapters\s*=\s*(\[[\s\S]*?\])\s*;/);
+        if (match === null) {
+            return [];
+        }
+        let chapters;
+        try {
+            chapters = JSON.parse(match[1]);
+        } catch (err) {
+            ErrorLog.log("RoyalRoadParser: found window.chapters but could not parse it: " + err);
+            return [];
+        }
+        let fictionPath = new URL(baseUrl).pathname.replace(/\/+$/, "");
+        return chapters
+            .filter(c => (c != null) && ((c.url != null) || (c.id != null)))
+            .map(c => ({
+                sourceUrl: new URL(c.url ?? `${fictionPath}/chapter/${c.id}/${c.slug}`, baseUrl).href,
+                title: c.title,
+                newArc: null
+            }));
     }
 
     // find the node(s) holding the story content
     findContent(dom) {
-        let content = util.getElement(dom, "div", 
-            e => (e.className === "portlet-body") &&
-            (e.querySelector("div.chapter-inner") !== null)
-        );
+        // Royal Road's Tailwind redesign removed the "portlet-body" wrapper; the chapter
+        // body is now div.chapter-inner inside a generic container. Anchor on
+        // div.chapter-inner (present in every variant) and use its parent as the content
+        // root; fall back to the legacy portlet-body match, then page-content-wrapper.
+        let content = dom.querySelector("div.chapter-inner")?.parentElement
+            ?? util.getElement(dom, "div",
+                e => (e.className === "portlet-body") &&
+                (e.querySelector("div.chapter-inner") !== null))
+            ?? dom.querySelector(".page-content-wrapper");
 
-        // fix embeded image links 
-        content.querySelector(".author-note")?.querySelectorAll("a")?.forEach((e) => 
+        // fix embeded image links (null-guarded so one odd chapter can't abort the book)
+        content?.querySelector(".author-note")?.querySelectorAll("a")?.forEach((e) =>
         {
             let img = e.querySelector("img");
-            if (img !== null) 
+            if (img !== null)
             {
                 e.href = img.src;
             }
         });
-        return content || dom.querySelector(".page-content-wrapper");
+        return content;
     }
 
     populateUIImpl() {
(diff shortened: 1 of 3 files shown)


plugin/js/parsers/RoyalRoadParser.js
```diff
@@ -14,28 +14,74 @@ class RoyalRoadParser extends Parser {
...
+        // <table id="chapters">, a new Tailwind table (class "w-full divide-y", no id),
+        // or a JS-rendered shell with an empty <body>. The one constant across every
+        // variant is the full chapter list embedded as a `window.chapters = [...]`
+        // JSON array (each entry has id/title/slug/url). Parse that first
+        // (redesign-proof); fall back to scraping the old table for legacy pages.
+        let tocHtml = await HttpClient.fetchText(dom.baseURI);
+        // Capture book metadata (description + genres) from the RAW html now — the DOM
+        // the parser later sees has its <script> tags stripped and, on the redesign, no
+        // JSON-LD at all, so the full description must be scraped here.
+        RoyalRoadParser.lastBookMeta = RoyalRoadParser.extractBookMeta(tocHtml);
```
kuwoyuki: Don't store book meta in a static class prop, I'm pretty sure this will cause a bug if you queue 2 RR books at once.
Instance prop `this.lastBookMeta` or could attach it to the DOM i.e. `dom.__royalRoadMeta`

plugin/js/parsers/RoyalRoadParser.js
```diff
@@ -14,28 +14,74 @@ class RoyalRoadParser extends Parser {
...
+        // fallback: legacy server-rendered table
+        let doc = new DOMParser().parseFromString(tocHtml, "text/html");
+        let table = doc.querySelector("table#chapters");
         return util.hyperlinksToChapterList(table);
     }
 
+    // Extract the chapter list from the page's `window.chapters = [...]` JSON blob.
+    // Returns [] if not present/parseable, so the caller can fall back.
+    static chaptersFromWindowChapters(html, baseUrl) {
+        let match = html.match(/window\.chapters\s*=\s*(\[[\s\S]*?\])\s*;/);
```
kuwoyuki: `[\s\S]*?` is lazy.
what will happen if chapter is named `Chapter 1 [Interlude]`?

plugin/js/parsers/RoyalRoadParser.js
```diff
@@ -14,28 +14,74 @@ class RoyalRoadParser extends Parser {
...
+        // variant is the full chapter list embedded as a `window.chapters = [...]`
+        // JSON array (each entry has id/title/slug/url). Parse that first
+        // (redesign-proof); fall back to scraping the old table for legacy pages.
+        let tocHtml = await HttpClient.fetchText(dom.baseURI);
+        // Capture book metadata (description + genres) from the RAW html now — the DOM
+        // the parser later sees has its <script> tags stripped and, on the redesign, no
+        // JSON-LD at all, so the full description must be scraped here.
+        RoyalRoadParser.lastBookMeta = RoyalRoadParser.extractBookMeta(tocHtml);
+        let fromJson = RoyalRoadParser.chaptersFromWindowChapters(tocHtml, dom.baseURI);
+        if (0 < fromJson.length) {
```
kuwoyuki, line 32: maybe subjective but I really wouldn't write it this way 
```js
if (fromJson.length) return fromJson;
```

plugin/js/parsers/RoyalRoadParser.js
```diff
@@ -114,20 +160,147 @@ class RoyalRoadParser extends Parser {
...
+        return util.isNullOrEmpty(og) ? "" : og.trim();
     }
 
-    findChapterTitle(dom) {
+    // Capture the book's description + genres from the RAW page html. The sanitized DOM
+    // the parser later sees has scripts stripped and, on the redesign, no JSON-LD at all.
+    // Prefer JSON-LD when present; otherwise locate the full description in the body by
+    // matching the og:description as a needle (the redesign's classes are unstable).
+    static extractBookMeta(html) {
+        let doc = new DOMParser().parseFromString(html, "text/html");
```
kuwoyuki: since you parsed it here anyway, you can pass it to `getChapterUrls` for the fallback extraction, to avoid parasing the DOM twice

plugin/js/parsers/RoyalRoadParser.js
```diff
@@ -114,20 +160,147 @@ class RoyalRoadParser extends Parser {
...
+        if (util.isNullOrEmpty(og)) {
+            return null;
+        }
+        let needle = og.replace(/\s+/g, " ").trim().slice(0, 20);
+        if (needle.length === 0) {
+            return null;
+        }
+        let best = null;
+        let bestLen = 0;
+        for (let el of doc.querySelectorAll("div, section, article")) {
```
kuwoyuki: Can you narrow down the search? Iterating over every `div, section, article` is kinda meh, if there's some heuristic that can be used it would be nice

kuwoyuki (changes requested):

plugin/js/parsers/RoyalRoadParser.js
dteviot: @kuwoyuki 

I'm not sure how you'd queue 2 RR books at once?  If you open two copies of WebToEpub, they're two instances.  With their own statics.

plugin/js/parsers/RoyalRoadParser.js
dteviot: My preference is what has been written.  But either is fine with me.

plugin/js/parsers/RoyalRoadParser.js
dteviot: @possibletoactual 

You should probably have a look at util.locateAndExtractJson()

plugin/js/parsers/RoyalRoadParser.js
```diff
@@ -152,6 +333,43 @@ class RoyalRoadParser extends Parser {
...
+        RoyalRoadParser.appendInfoLine(dom, nodes, "Genres", this.extractSubject(dom));
+        let heading = dom.createElement("h2");
+        heading.textContent = "Description";
+        let descriptionHtml = RoyalRoadParser.lastBookMeta?.descriptionHtml;
+        if (!util.isNullOrEmpty(descriptionHtml)) {
+            // Full description captured from the raw page, keeping its paragraph
+            // formatting (sanitized downstream by populateInfoDiv).
+            nodes.push(heading);
+            let descDiv = dom.createElement("div");
+            descDiv.innerHTML = descriptionHtml;
```
dteviot: No NOT assign un-sanitized HTML to innerHTML.  It makes approval for Mozilla and Google stores difficult.
Better yet, do not assign to innerHTML at all.

plugin/js/parsers/RoyalRoadParser.js
```diff
@@ -114,20 +160,147 @@ class RoyalRoadParser extends Parser {
...
+            descriptionHtml = jsonld.description;
+            let holder = doc.createElement("div");
+            holder.innerHTML = jsonld.description;
+            description = holder.textContent.replace(/\s+/g, " ").trim();
+        } else {
+            let og = doc.querySelector("meta[property='og:description'], meta[name='description']")
+                ?.getAttribute("content") ?? "";
+            let el = RoyalRoadParser.findDescriptionElement(doc, og);
+            if (el != null) {
+                descriptionHtml = el.innerHTML;
```
dteviot: Do not assign to innerHTML.

plugin/js/parsers/RoyalRoadParser.js
```diff
@@ -114,20 +160,147 @@ class RoyalRoadParser extends Parser {
...
+    // matching the og:description as a needle (the redesign's classes are unstable).
+    static extractBookMeta(html) {
+        let doc = new DOMParser().parseFromString(html, "text/html");
+        let jsonld = RoyalRoadParser.extractBookInfo(html);
+        let description = null;
+        let descriptionHtml = null;
+        if ((jsonld != null) && !util.isNullOrEmpty(jsonld.description)) {
+            descriptionHtml = jsonld.description;
+            let holder = doc.createElement("div");
+            holder.innerHTML = jsonld.description;
```
dteviot: do not assign to InnerHTML.

plugin/js/parsers/RoyalRoadParser.js
kuwoyuki: Ah, yeah you're right, each popup is a separate document.  
I still think this is an anti-pattern though, because every method that writes or reads `lastBookMeta` already has `this`. So it's the wrong lifetime.. metadata is per-parse, per-book. A static is per realm (popup).

PR #2794, opened 2026-07-09
Add NullTranslation parser

kuwoyuki:
[NullTranslation](https://nulltranslation.com) parser.
it's a quite annoying to parse Next.js (App Router / RSC) site that has chapters AES-GCM encrypted.
I left in 2 chapter builder methods for now, 1st decrypts in the ext and the latter just saves it in a div to decrypt in EpubEditor
```js
            // we do it here
            // const plaintext = await this._decryptChapterContent({
            //     encryptedData: encryptedBytes,
            //     b64Key: base64Key,
            // });
            // NullTranslationParser.renderChapterText(plaintext, newDoc);

            // ..or we save the payload for post-processing
            NullTranslationParser.embedEncryptedChapter({
                newDoc, encryptedBytes, base64Key, url,
            });
```
not familiar with addon store policies so lmk if the former needs to be removed.

Mutator PR here https://github.com/dteviot/EpubEditor/pull/23

`eslint/pack.js` change is because split and join by `\r` was giving me an empty popup.html in the archive, I assume it has something to do with git and line endings.. I think it's safer to just normalize the line endings before splitting.

### DIFF
--- a/eslint/pack.js
+++ b/eslint/pack.js
@@ -145,9 +145,10 @@ var addPopupHtmlToZip = function(zip) {
     return readFilePromise("../plugin/popup.html")
         .then(function(data) {
             let htmlAsString = data.toString()
-                .split("\r")
+                .replace(/\r\n?/g, "\n")
+                .split("\n")
                 .filter(s => !s.includes("/experimental/"))
-                .join("\r");
+                .join("\n");
             zip.add("popup.html", new zipjs.TextReader(htmlAsString));
         });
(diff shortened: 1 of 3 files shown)


dteviot:
@kuwoyuki 
> not familiar with addon store policies so lmk if the former needs to be removed.

It's OK to leave the decryption code in there, just not execute it.  Basically, we're not allowed to decrypt it in the extension in the store.

PR #2796, opened 2026-07-09
Update Sbxh parser

Bartuzen:
Made some improvements to Sbxh parser.
- Added site's mirror domain.
- Chapters are now fetched from all paginated chapter list pages. Previously, only the chapters visible on the currently loaded page were fetched.
- Rather than using a broad selector to get chapter title and parsing it, I instead directly fetched the title which gives a better result.
- Chapter title now always includes both chapter title and chapter count. Sometimes both chapter count and title can be same which results in a title like `1화 - 1화`, but since title format isn't consistent, I kept it like this.

### DIFF
--- a/plugin/js/parsers/experimental/Sbxh1Parser.js
+++ b/plugin/js/parsers/experimental/Sbxh1Parser.js
@@ -1,7 +1,7 @@
 "use strict";
 
 parserFactory.registerUrlRule(
-    (url) => /^https?:\/\/(?:www\.)?sbxh\d+\.com(?:\/|$)/.test(url),
+    (url) => /^https?:\/\/(?:www\.)?(sbxh|toki)\d+\.com(?:\/|$)/.test(url),
     () => new Sbxh1Parser(),
 );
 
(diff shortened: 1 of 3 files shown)


gamebeaker:
@Bartuzen do you want to add yourself to the "contributors" section of packages.json and updated the "Credits" section of readme.md with your name?

Just a question out of curiosity did you use the wiki guide https://github.com/dteviot/WebToEpub/wiki/Developer-guide?

Bartuzen:
Sure, let me add my name real quick.
For the second question, no. I didn't check out wiki too much and didn't realize there was such a page.

plugin/js/parsers/experimental/Sbxh1Parser.js
```diff
@@ -10,17 +10,22 @@ class Sbxh1Parser extends Parser {
...
-            ...dom.querySelectorAll(`a.novel-ep-link[href*="/novel/${novelId}/"]`),
-        ];
-        if (chapterLinks.length === 0) {
-            chapterLinks = [...dom.querySelectorAll(`a[href*="/novel/${novelId}/"]`)];
+        let chapterLinks = Sbxh1Parser.parseChapters(dom, novelId);
+
+        let page = 1;
+        while (chapterLinks[chapterLinks.length - 1].querySelector(".ne-num").textContent !== "1화") {
+            page++;
+            const xml = (await HttpClient.wrapFetch(`${dom.baseURI}?epage=${page}`)).responseXML;
```
kuwoyuki: maybe use the [URL](https://developer.mozilla.org/en-US/docs/Web/API/URL) constructor here? So it doesn't break if there's already a search param.
i.e.
```js
let nextUrl = new URL(dom.baseURI);
nextUrl.searchParams.set("epage", page);
const { responseXML: xml } = await HttpClient.wrapFetch(nextUrl.href);
```

dteviot:
@Bartuzen 

Apologies, but due to other MRs that got in first, can you rebase and fix package.json please.

dteviot:

PR #2798, opened 2026-07-10
Fix builds on Linux and macOS

Bartuzen:
Currently, if you build the extension on Linux or macOS, extension's popup won't open which makes the extension unusable. Using the correct line separator fixes the issue on both Linux and macOS. I haven't tested it on Windows, but it should work there as well.


dteviot:
@Bartuzen 

Sorry, @kuwoyuki got in first.
https://github.com/dteviot/WebToEpub/commit/6e4fd22328fbd53b598d9a501730f63d95684842

Issue #2785, opened 2026-07-08
Option to treat 'no content found' error as a Captcha challenge page.


Yomafil:
Add a new option in Advanced Options to treat 'no content found' errors as an 'Error 403' or 'Captcha/Rate Limit Challenge Page'.
Useful as a workaround for when 'isCustomError' and 'setCustomErrorResponse' are not implemented in the website parser, but the website still gives captcha or rate limit challenge pages.
Right now it just fails and end the downloading, with no way to resume or continue, leaving us with fragmented chapters across multiple EPUBs if we want to download a book with lots of chapters.

hiSandog:
A useful next step would be a minimal fixture for Option to treat 'no content found' error as a Captcha challenge page. that includes one failing input and one nearby control case. That would make it clear whether the fix addresses the reported edge case without widening behavior unexpectedly.

Yomafil:
For now, I only managed to make a short PoC. It only changes the 'no content found error' to an 'HTTP 403 warning' while failing. 
Still haven't found a way to allow for retries. It doesn't seem to be a trivial feature to implement.
Hopefully, it's not what I think, and there's an easier way than having to change the chapters fetch implementation...
https://github.com/Yomafil/WebToEpub/commit/c86661cf1ede45c25faf48131470da8cef58f751

gamebeaker:
Test versions for Firefox and Chrome have been uploaded to https://github.com/dteviot/WebToEpub/releases/tag/developer-build. Pick the one suitable for you, follow the "How to install from Source (for people who are not developers)" instructions at https://github.com/dteviot/WebToEpub/tree/ExperimentalTabMode#user-content-installation and let us know how it goes.

gamebeaker:
Updated version (1.0.14.0) has been submitted to Firefox and Chrome stores.
Firefox version should be available within two weeks (but might take 4 to 6 weeks)
Chrome might be available in a few hours to 21 days. (Typically 3 to 7 days)

PR #2799, opened 2026-07-10
Add new option to change `no content found error` into `HTTP 403 error`

Yomafil:
Add the option to turn `no content found error` into an `HTTP 403 error`.
It is useful as a workaround when `isCustomError` and `setCustomErrorResponse` are not implemented in a parser.
When a site gives the wrong error code when giving the user a challenge page (captcha, Cloudflare, rate limit, etc...).
It allows the user to clear the challenge page and resume downloading from where they were, instead of losing all progress or ending up with multiple fragmented EPUBs when downloading a single book.

Tested and working on alicesw.com, with and without `isCustomError` and `setCustomErrorResponse` implemented in the AliceswParser.

Only issue is the fact that this implementation doesn't work if a custom parser already include `fetchChapter` without passing a reference to the `parser` in the `wrapOptions`.
If it fails in this way and the `NoContentToError403 option` is selected, it will now throw an `HTTP 403 error` instead of a `no content found error`, before aborting the download.

Tested and failing (has expected) on alicesw.com, without passing a reference to the `parser` in `wrapOptions` when `fetchChapter` implemented in the AliceswParser.
Tested and working on alicesw.com, while passing a reference to the `parser` in `wrapOptions` when `fetchChapter` implemented in the AliceswParser.

#2785

### DIFF
--- a/plugin/_locales/en/messages.json
+++ b/plugin/_locales/en/messages.json
@@ -467,6 +467,10 @@
 		"message": "Skip chapters that return HTTP 404 error",
 		"description": "Label in front of checkbox for skipping chapters that can't be fetched"
 	},
+	"__MSG_label_No_Content_To_Error_403_Page__": {
+		"message": "Treat 'no content found error' as 'HTTP 403 error'.",
+		"description": "Label in front of checkbox for treating 'no content found' error as 'HTTP 403 error'"
+	},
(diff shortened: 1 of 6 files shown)


dteviot:
@Yomafil 

It's an advanced option.  I'll accept it.

Thank you.

PR #2808, opened 2026-07-11
Add off-store builds with external parser scripts loading

kuwoyuki:
This is a draft of an off-store build variant that can load full parser scripts from remote repos at runtime. I tried to leave the store build untouched.
The motivation is this PR https://github.com/dteviot/WebToEpub/pull/2225. MV3 can't load remote code per Chrome Web Store policy. But some sites need complex decryption, login sessions, Cloudflare cookie handling, etc.
Off-store build (MV2 with unsafe-eval CSP) evals external scripts directly in the popup context so they have the full context.

- `npm run build` now produces 4 zips: 2 store (unchanged) + 2 off-store (Chrome MV2 + Firefox MV2, both with unsafe-eval CSP)
- Off-store build has `ExternalScriptLoader.js` that fetches an `index.json` from configured repo URLs, then fetches and evals each listed parser script
- External scripts use the same `parserFactory.register()` as bundled parsers so a script can be bundled or external with no changes
- If the external script registers a parser for a host that already has a bundled one, the external one wins
- Scripts cached in `chrome.storage.local` with optional auto-update

This will not work in MV3 (thanks Google) CSP blocks `eval()`. Sandboxed iframes can eval but are isolated from extension APIs, cookies, and login sessions, so that's useless for this ext.


dteviot:
@kuwoyuki 
It's probably going to be a while before I have time to fully review this.

PR #2811, opened 2026-07-13
update ScribblehubParser

gamebeaker:
i got blocked with the current minimumThrottle 
image

i think that the fast toc loading is the culprit.


gamebeaker:
I changed my ip and cleared browser cache so maybe it is a problem on the backend and not my ip.

gamebeaker:
image

not just me

gamebeaker:
i am getting blocked maybe need to set referer.

gamebeaker:
we need to set the referer.
#2814

PR #2817, opened 2026-07-16
Trust image bytes over Content-Type when detecting media type

possibletoactual:
Some CDNs send a wrong `Content-Type`. Concretely, Royal Road serves cover images with `Content-Type: image/png` when the file is actually a JPEG. WebToEpub trusts the header (`imageInfo.mediaType = xhr.contentType`), so the image ends up stored as `.png` / `image/png` while the bytes are JPEG — an invalid EPUB image that readers reject (no cover, and in strict readers like Apple Books the whole book can render degraded).

### Fix
`fixupInvalidMediaType` now byte-sniffs the real type with `util.detectMimeType()` and prefers it, falling back to the header/extension only when the bytes are an unrecognised format. Correctly-typed images are unaffected because the sniffed type matches the header.

### Repro
`curl -I "https://www.royalroadcdn.com/public/covers-large/<cover>.jpg"` returns `content-type: image/png` for JPEG bytes; the packed EPUB's cover is declared `image/png` but `file` reports JPEG.

### Checklist
- Unit tests: unaffected.
- eslint: clean.
- No new user-facing behavior/toggle (bug fix).
- Base branch: `ExperimentalTabMode`.

### DIFF
--- a/plugin/js/ImageCollector.js
+++ b/plugin/js/ImageCollector.js
@@ -413,17 +413,22 @@ class ImageCollector {
     }
 
     fixupInvalidMediaType(imageInfo) {
+        // Trust the actual bytes over the server's Content-Type. Some CDNs (e.g. Royal
+        // Road) return a wrong type (image/png for a JPEG), producing an invalid EPUB
+        // image that readers reject. Fall back to the header/extension only when the
+        // bytes are an unrecognised format.
+        let detected = util.detectMimeType(imageInfo.getBase64(25));
+        if (detected != null) {
+            imageInfo.mediaType = detected;
+            return;
+        }
         if (!imageInfo.mediaType?.startsWith("image")) {
-            imageInfo.mediaType = util.detectMimeType(imageInfo.getBase64(25));
-            if (imageInfo.mediaType == null)
-            {
-                let path = new URL(imageInfo.sourceUrl).pathname;
-                let index = path.lastIndexOf(".");
-                let format = (index < 0)
-                    ? "jpeg"
-                    : path.substring(index + 1);
-                imageInfo.mediaType = "image/" + format;
-            }
+            let path = new URL(imageInfo.sourceUrl).pathname;
+            let index = path.lastIndexOf(".");
+            let format = (index < 0)
+                ? "jpeg"
+                : path.substring(index + 1);
+            imageInfo.mediaType = "image/" + format;
         }
     }
 


dteviot:
@possibletoactual 
I'm confused.  If you look at the code, the existing code already sniffs the bytes, and only uses the media type if the bytes are not recognized.  So if WebToEpub is getting it wrong for RoyalRoad, the problem is elsewhere.
Can you provide URL of RoyalRoad with png for testing?

Kiradien:
I'll just add my own 2cents here; as the one who originally implemented that mime detection I did rigorous testing between this mode of operation and the original. 

What I found, at that time, is the xhr - the request response object  containing the image details - doesn't base the mediatype solely off the extension and was under most circumstances, more reliable than the clutched together list of MIME Types we're scanning for. Filling it in when empty is one thing - because that means it was either unrecognized by the server infrastructure or request in general - but...

Relying solely on the util's MIME signature scanning will work for common image types, but I can tell you now that several image formats will not be recognized at all - they also tend not to load at all without the appropriate extension, so I didn't create an exhaustive list at the time.

There should be another solution for the issue you're hitting; we just need to figure it out from this specific case - getting that RoyalRoad URL would be ideal.

possibletoactual:
Here's a live repro URL — cover of https://www.royalroad.com/fiction/35372:

`https://www.royalroadcdn.com/public/covers-large/35372-system-change.jpg?time=1666810759`

```
$ curl -sI <url> | grep -i content-type
content-type: image/png
$ curl -s <url> | xxd -l 4
ffd8 ffe0    # JPEG/JFIF magic — file(1) says "JPEG image data, 400x600"
```

Verified end-to-end with the patch applied: packing this fiction produces `<item ... media-type="image/jpeg"/>` for the cover, with the stored bytes matching (JFIF, 400×600).

@dteviot — I believe the sniff you're thinking of runs in only two places: (1) `fixupInvalidMediaType` sniffs only inside `if (!imageInfo.mediaType?.startsWith("image"))`, so it never fires when the server sends a wrong-but-valid-looking type like `image/png` for JPEG bytes (`fetchImage` assigns `xhr.contentType`, which comes straight from `response.headers.get("content-type")` with no sniffing); and (2) `runCompression`'s "auto" output type, which only runs when compression is enabled. With compression off, the wrong header survives into the EPUB manifest. Does this match your experience?

@Kiradien — on unrecognized formats: the patch keeps your fallback intact, and the same test pack happened to demonstrate it. One resource in the book is served with an `image/*` content-type but its bytes aren't a recognizable image; `detectMimeType` returned null and the patch preserved the header-derived type verbatim, identical to current behavior. The change only takes effect when the sniff positively identifies a known format that contradicts the header. Make sense?

dteviot:
> it never fires when the server sends a wrong-but-valid-looking type like image/png

You are correct.  I misread the code.

teloscientist-hub:
BTW - this is a good chance to say - Thanks! WebToEpub is awesome! :-)

PR #2818, opened 2026-07-16
Add extractAuthor to wtr-lab

xRahul:
Currently wtr-lab does not extract author from the page and it always comes as unknown. This PR fixes by adding extractAuthor to the parser


dteviot:
@xRahul 
Thank you.

PR #2841, opened 2026-07-24
Fix wuxiaworld parser 2

ARYAN-9099:
fixed not detecting chapters when bookmarked.
Chapter titles are now inside chapters.

### DIFF
--- a/plugin/js/parsers/WuxiaworldParser.js
+++ b/plugin/js/parsers/WuxiaworldParser.js
@@ -26,18 +26,20 @@ class WuxiaworldParser extends Parser {
             let novelUrlMatch = dom.baseURI.match(/wuxiaworld\.com\/novel\/([^/?#]+)/);
             if (novelUrlMatch) {
                 let novelSlug = novelUrlMatch[1];
-                let searchArea = dom.querySelector("#full-width-tabpanel-1") || dom.body;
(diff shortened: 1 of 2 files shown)


ARYAN-9099:
@gamebeaker for unlocked chapters the website fetches the content after a while so I'm working on a approach in which a new tab will be opened and we will get content from there.
Do you have any better idea?

gamebeaker:
@ARYAN-9099 as i don't have unlocked chapters to test it. I am not sure.

It would be great if you find a solution with the tab opening and than copying the content from there. I tried it in the past but gave it up. I wish you good luck.

ARYAN-9099:
@gamebeaker for a new account you do get 3 free unlocks. 
I was just checking if the tab opening method is allowed.
Also for doing another approach is it okay to get the jwt?

gamebeaker:
@ARYAN-9099 
> I was just checking if the tab opening method is allowed.

it is. (afaik)

>Also for doing another approach is it okay to get the jwt?

yes i see it the same as cookies.

here is an example where something similar to jwt get used:
https://github.com/dteviot/WebToEpub/blob/ded3854771c0b6f32b72f59f223683ecf68d8ea5/plugin/js/parsers/WtrlabParser.js#L19-L24

the magickey value can change with time.

PR #2844, opened 2026-07-25
Wuxiaworld - properly get unlocked chapters

ARYAN-9099:
previously manually unlocked chapters werent properly pulled. this was cause the unlocked chapters were separately requested 1-2 seconds after opening that chapter.
so now we open the chapter in a new tab, get the content then close.

note- i have only have 3 unlocked chapters. It works properly on that. I dont think more no of unlocked chapters would be a problem tho.

### DIFF
--- a/plugin/js/parsers/WuxiaworldParser.js
+++ b/plugin/js/parsers/WuxiaworldParser.js
@@ -91,8 +91,41 @@ class WuxiaworldParser extends Parser {
     }
 
     extractTitleImpl(dom) {
-        let titleNode = dom.querySelector("span[data-testid='title'], h4");
-        return titleNode ? titleNode.textContent.trim() : super.extractTitleImpl(dom);
+        // Modern Wuxiaworld novel pages contain many h4 headings ("Details",
+        // "Reviews", etc.).  The old selector `span[data-testid='title'], h4`
+        // could grab the "Details" h4 and make every EPUB filename "Details".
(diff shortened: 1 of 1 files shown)


gamebeaker:
> note- i have only have 3 unlocked chapters. It works properly on that. I dont think more no of unlocked chapters would be a problem tho.

how does this work with WebToEpub?
Are only the free Chapter selected or are all Chapter selected?
I am not sure how the website works but if all Chapters are selected and WebToEpub tries to load them could it be that it incurs cost because it unlocks them?

ARYAN-9099:
@gamebeaker for free chapters it works as old.
only for unlocked chapters, it one by one opens it in other tab.

gamebeaker:
@ARYAN-9099 i mean in the toc which chapter are selected?

ARYAN-9099:
both free and unlocked. locked are not selected

gamebeaker:
https://www.wuxiaworld.com/novel/tribulation-of-myriad-races this novel only loads 50 Chapters in WebToEpub so how can the user select unlocked chapters manually?

ARYAN-9099:
if you have any unlocked chapters, they would automatically come in extension.

ARYAN-9099:
@gamebeaker another person checked and it works nicely even with more chapters. Tell if there are any issues.

gamebeaker:
@ARYAN-9099 thanks for your involvement.

PR #2853, opened 2026-07-28
added boluomao parser

azathoth4:
added a parser for https://www.boluomao.com/

### DIFF
--- a/plugin/js/parsers/BoluomaoParser.js
+++ b/plugin/js/parsers/BoluomaoParser.js
@@ -1,0 +1,95 @@
+"use strict";
+
+parserFactory.register("boluomao.com", () => new BoluomaoParser());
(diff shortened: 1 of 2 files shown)


gamebeaker:
@azathoth4 thanks for your work but because of the add-on store policies will not be merged because of:

```
    decodeObf(s) {
        if (!s) return "";
        let raw = atob(s);
        let bytes = new Uint8Array(raw.length);
        for (let i = 0; i < raw.length; i++) {
            bytes[i] = raw.charCodeAt(i) ^ ((i % 127) + 1);
        }
        return new TextDecoder('utf-8').decode(bytes);
    }
```
tldr: we aren't allowed to de obfuscate you can remove the code from WebToEpub and add the de obfuscation in this project: https://github.com/dteviot/EpubEditor

azathoth4:
ah ok, sorry didnt know

PR #2857, opened 2026-07-29
Trim lines (elements) from chapter body before packing into EPUB file

g-play01:
Ran into cases where sites add content like number of words per chapter, the chapter title again, a please report if any issues at the start / end of a chapter and that ends up getting packed into the epub and can get annoying especially when you use a TTS reader created this advanced option that defaults to 0 i.e so off by default that removes a specified number of elements from the processed chapter bodies in convertRawDomToContent. Tested a local build on firefox on 2 sites works as expected. 

Let me know if ya'll want anything else changed on this.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,3 +9,5 @@ package-lock.json
 .vscode
 /plugin/dompurify/
 /plugin/@zip.js/
+mise.toml
+bun.lock
--- a/package.json
+++ b/package.json
@@ -119,7 +119,8 @@
     { "name": "possibletoactual" },
     { "name": "kuwoyuki", "email": "kuwoyuki.gh@cock.li" },
     { "name": "bendur"},
-    { "name": "Bartuzen", "email": "git@bartuzen.dev" }
+    { "name": "Bartuzen", "email": "git@bartuzen.dev" },
+    { "name": "g-play01", "email": "41573585+g-play01@users.noreply.github.com"}
   ],
   "license": "GPL-3.0-only",
   "bugs": {
--- a/plugin/_locales/en/messages.json
+++ b/plugin/_locales/en/messages.json
@@ -367,6 +367,14 @@
 		"message": "Delay per chapter in ms",
 		"description": "Delay per chapter in ms"
 	},
+	"__MSG_label_Remove_First_Lines_Of_Chapter__": {
+		"message": "Number of lines to remove from start of chapter body",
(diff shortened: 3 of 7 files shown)


ARYAN-9099:
You can also fix the parsers of the problematic sites.

g-play01:
The sites have varying behaviour based on the story series, some stories only have 1 line to remove while otheres have 2 and others have none, I felt this was a more universal solution.

ARYAN-9099:
@g-play01 how about a additional feature that will remove lines containing specific phrases?

gamebeaker:
I think stuff like this should be included in the specific parser or made with another program.

> The sites have varying behaviour based on the story series, some stories only have 1 line to remove while otheres have 2 and others have none, I felt this was a more universal solution.

Any examples?

As this implies manual involvement per downloaded epub i think it is reasonable that the person post processes the ebook. With for example calibre or https://github.com/dteviot/EpubEditor

It could be for example a EpubEditor script where the first variable is the count of elements removed.

g-play01:
> @g-play01 how about a additional feature that will remove lines containing specific phrases?

Yes that will be helpful to remove additional "site ads" 
but feels more complex and many potential edge cases, might look into it later.

> Any examples?

novelfire.net/book/lord-of-the-mysteries/
vs
novelfire.net/book/reverend-insanity/

> As this implies manual involvement per downloaded epub i think it is reasonable that the person post processes the ebook.

Yes totally can be done by external tools however, it's many more steps vs a quick trim setting on the epub packing. 
I run this on my phone and get books on the go and this felt like a more natural/universal/easy solution to many problems such as varying behaviour on the site (my current issue) and also other cases including when it takes time to get parsers fixed on some sites or (hypothetically) whenever the default parser gets used and stuff gets added on the bottom/top that you dont normally want.

As for EpubEditor it looks like its a good place for many transformations including this and what ARYAN-9099 mentioned 
That being said I dont think it's undeserving of a place here, it keeps the experience direct and easy. 

I feel its a very lightweight feature that is off by default and can be called upon when needed to make life easier.

dteviot:
@g-play01 

It's a nice idea, but I'm not comfortable with it.  WebToEpub's job is collect the epub.  And to err on the side of collect too much, rather than miss stuff.  Something like this, might result in content being lost.  Yes, I know it's a use at own risk, but I think too much chance of chapter format changing and stuff being lost without knowing, until the chapter is read.

PR #2868, opened 2026-08-02
update ChyoaParser

gamebeaker:
@ARYAN-9099 as you made the last commit please have a look.

I changed it because i didn't like that it always downloads all chapters. As the site requires an account you can fetch a toc page.

And i wanted to have a look on how you implemented branching chapters.

As i saw that you changed epubItemSupplier with a lot of code i wanted to test if it could have been solved simpler. (I didn't test self recursive links but it should be no problem as i don't walk the chapter links.)

### DIFF
--- a/plugin/js/parsers/ChyoaParser.js
+++ b/plugin/js/parsers/ChyoaParser.js
@@ -64,7 +64,7 @@ class ChyoaParser extends Parser {
         let ChapterArray = this.reponseToToc(resp);
         chapterUrlsUI.showTocProgress(ChapterArray);
         let nextOffset = resp.data.nextOffset;
-        while(resp.data.hasMorePages == true){
+        while (resp.data.hasMorePages == true) {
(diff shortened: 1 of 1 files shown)


ARYAN-9099:
Great work!

1) do we also add the chapter number inside the chapter too?
2) in mobile readers, in more nested stories in the TOC the headings begin to move more and more to the right after which they aren't able to be read. (any ideas what can we do for this?)

gamebeaker:
> do we also add the chapter number inside the chapter too?

i find this more clear for navigation but i can also remove the number if you want.

ARYAN-9099:
@gamebeaker 
Im not saying to remove the numbers but rather to add them before chapter heading in chapter.

gamebeaker:
@ARYAN-9099 i made it optional you now have a checkbox like on quidian parser. (you need to reload webtoepub so that it takes effect.)

PR #2898, opened 2026-08-09
add: novelturk.com

kerimmkirac:


dteviot:
@kerimmkirac 

Sorry, but if chapter content is encrypted, WebToEpub is not allowed to decrypt it.  That violates Chrome store rules for extensions.  Best we can do is have WebToEpub put the encrypted content into an element and have EpubEditor decrypt it as an external step.  See the FictioneerParser.  (and LilyOnTheValley.)

PR #2909, opened 2026-08-11
add: sammyandpassion.com

### LINKED ISSUE #2855: Please add site https://sammyandpassion.com/
Hello please add this site, or if there's no time please teach me how? I tried using Default Parser as you can see here;

Image

but whenever I test it and packed the epub, it shows this;

Image

I tried following the steps on how to properly use the Default Parser and where could I find the CSS but when I click the first chapter then open the devtool it only show this;

Image

you may ask "where did I get the CSS on first attempt?" I follow what my friend send me to put in and said I just need to copy/paste it, that's where I found it.

kerimmkirac:
Closes #2855

### DIFF
--- a/plugin/js/parsers/SammyAndPassionParser.js
+++ b/plugin/js/parsers/SammyAndPassionParser.js
@@ -1,0 +1,82 @@
+"use strict";
+
+parserFactory.register("sammyandpassion.com", () => new SammyAndPassionParser());
+
+class SammyAndPassionParser extends Parser {
+    constructor() {
+        super();
+        this.minimumThrottle = 500;
+    }
+
+    extractTitleImpl(dom) {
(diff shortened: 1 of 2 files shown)


plugin/js/parsers/SammyAndPassionParser.js
```diff
@@ -0,0 +1,82 @@
...
+            })
+            .map((link) => ({
+                sourceUrl: link.href,
+                title: link.textContent.trim()
+            }));
+    }
+
+    findContent(dom) {
+        let content = dom.querySelector("div.entry-content");
+        if (content) {
```
dteviot, line 60: FYI.
findContent() is called multiple times.  If you're going to modify the content, it should be done in preprocessRawDom()

Have a look at Template.js for list of the standard hook points.

PR #2910, opened 2026-08-11
add: chikari.moe

kerimmkirac:


plugin/js/parsers/ChikariParser.js
```diff
@@ -0,0 +1,90 @@
...
+
+    async fetchChapter(url) {
+        let { slug, number } = this.extractChapterInfo(url);
+        let chapterData = await this.fetchApiJson(`/novels/${slug}/chapters/${number}/read`);
+        if (!chapterData?.body) {
+            throw new Error("Unexpected chapter response for " + url);
+        }
+
+        let newDoc = Parser.makeEmptyDocForContent(url);
+        let wrapped = "<div>" + chapterData.body.split(/\n+/)
```
dteviot, line 77: see Parser.addTextToChapterContent()

gamebeaker: @dteviot should this be changed?

dteviot:
@gamebeaker @kerimmkirac 

I assume you're referring to the use of sanitize() later.  Which may trip up the Mozilla automated tests.
Also, there's an existing function to do the work.
I've created an issue to fix, #2913 but I've run out of time for tonight.

dteviot:
@kerimmkirac 

I just tried this with https://chikari.moe/series/the-legendary-mechanic,
and it doesn't work because your looking for a URL with /novels/ in it.  Has site changed, or have I gone on the wrong path?

kerimmkirac:
@dteviot 

Well, https://chikari.moe/series/the-legendary-mechanic is a manhwa, not a novel — I made it only look for /novels/ so we can only download novels. You could try https://chikari.moe/novels/got-dropped-into-a-ghost-story-still-gotta-work for example. We could change it to support things other than just novels.

dteviot:
@kerimmkirac 
The distinction on the site wasn't obvious.
Not asking you to do anything but I do have minor concern that someone will try to use a manhwa on the site and get the error.  Then we will need to hunt it down.  And probably get confused, if it's not you.

gamebeaker:
@dteviot should WebToEpub also support manga/ manhwa sites or only novels?

kerimmkirac:
@dteviot 

The update you did on chikari broke the extension. On this series — https://chikari.moe/novels/got-dropped-into-a-ghost-story-still-gotta-work — the body field contains real html tags, like <strong>[Dark Exploration Records Character Maker]</strong>. My version was handling those correctly, but yours doesn't — it escapes them instead of rendering them. We can fix it a different way than I did, whatever works best for you.

PR #2911, opened 2026-08-11
add: novelarchive.cc

kerimmkirac:


plugin/js/parsers/NovelArchiveParser.js
```diff
@@ -0,0 +1,73 @@
...
+        return dom.querySelector("#novel-title");
+    }
+
+    extractAuthor(dom) {
+        let author = dom.querySelector("#novel-author");
+        return author ? author.textContent.trim() : super.extractAuthor(dom);
+    }
+
+    findCoverImageUrl(dom) {
+        return util.getFirstImgSrc(dom, "#novel-cover")
```
dteviot, line 22: getFirstImgSrc() looks for first img element that is a child of the given selector.  If the site always gives the title img an id of "novel-cover", there's probably no need to call getFirstImgSrc().

plugin/js/parsers/NovelArchiveParser.js
```diff
@@ -0,0 +1,73 @@
...
+    async fetchChapter(url) {
+        let response = await HttpClient.fetchJson(url);
+        let chapterData = response.json?.chapter;
+        if (!chapterData?.content) {
+            throw new Error("Unexpected chapter response for " + url);
+        }
+
+        let newDoc = Parser.makeEmptyDocForContent(url);
+        chapterData.content.split(/\n+/).filter((line) => line.trim() !== "").forEach((line) => {
+            let p = newDoc.dom.createElement("p");
```
dteviot, line 63: Duplicated code from previous parser.
Have a look at Parser.addTextToChapterContent()

PR #2924, opened 2026-08-14
feat(parser): add ad-bypass for readrift

matthewsong18:
Readrift started blocking the latest chapters until you watch their Ads. They changed their API around this, preventing you from requesting the chapter content.

I figured out how the ad API works and have added a bypass. The bypass works as follows:

- Request ad token for chapter. This uses the user session id.
- Send POST request with token. This unlocks the chapter for your user session id.

Worked perfectly in testing on chromium.


gamebeaker:
@matthewsong18 amazing work. Thanks for the help.

PR #2930, opened 2026-08-17
feat: add Smart Rule engine for auto-detecting CSS selectors

312Aroha:
### Summary
This PR introduces a heuristic engine (`HeuristicScanner`) to automatically detect Chapter Title and Content CSS selectors for unknown websites, reducing manual configuration effort.

### Key Changes:
*   **Auto-detection**: Scans pre-loaded DOM using `TreeWalker` to find content and titles (supports EN, CN, JP, KR naming conventions).
*   **Zero Extra Network Overhead**: Reuses the live `dom` object provided by `main.js` (no extra `fetch` calls).
*   **Robustness**: 
    *   Added `CSS.escape()` for modern CSS frameworks (like TailwindCSS with colons).
    *   Added `try-catch` safety blocks when reading cached CSS from LocalStorage to prevent crashes.

### Testing:
Tested locally with various sites (e.g., Qidian, QuShuWang), working smoothly.

*P.S. I actually have 3 more PRs in my local pipeline (a No-ToC Chain Crawler, UI/UX Enhancements, and i18n Localization), but I wanted to submit this one first to ensure my code style and approach align with your standards. Let me know if any changes are needed! 😄*

### DIFF
--- a/plugin/js/DefaultParserUI.js
+++ b/plugin/js/DefaultParserUI.js
@@ -87,7 +87,9 @@ class DefaultParserSiteSettings {
                         for (let e of element.querySelectorAll(config.removeCss)) {
                             e.remove();
                         }
-                    } catch (e) {}
+                    } catch (e) {
+                        // Ignore invalid CSS selector errors
+                    }
                 };
             }
         }
--- a/plugin/js/DefaultParserUI.js
+++ b/plugin/js/DefaultParserUI.js
@@ -1,12 +1,10 @@
 "use strict";
 
-/** Keep track of how to user tells us to parse different sites */
 class DefaultParserSiteSettings {
     constructor() {
         this.loadSiteConfigs();
     }
 
-    /** @private */
     loadSiteConfigs() {
         let config = window.localStorage.getItem(DefaultParserSiteSettings.storageName);
         this.configs = new Map();
@@ -40,7 +38,6 @@ class DefaultParserSiteSettings {
         }
     }
 
-    /** @private */
     isConfigChanged(hostname, contentCss, titleCss, removeCss, testUrl) {
         let config = this.configs.get(hostname);
         return (config === undefined) || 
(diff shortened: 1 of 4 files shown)


ARYAN-9099:
is this added functionality for default parser or a new parser?

312Aroha:
> is this added functionality for default parser or a new parser?

It is an added functionality for the existing Default Parser.

Here is how it fits into the architecture:
* It does not introduce a new parser; it just makes the existing `DefaultParserUI` smarter.
* It runs locally to pre-fill CSS selectors when a user opens an unsupported site.
* It safely falls back to the original behavior without overriding any existing user configurations.

plugin/js/DefaultParserUI.js
```diff
@@ -118,16 +129,91 @@ class DefaultParserUI { // eslint-disable-line no-unused-vars
...
 
         let config = parser.siteConfigs.getConfigForSite(hostname);
+        let activeUrl = document.getElementById("startingUrlInput").value;
+
         if (config != null) {
             DefaultParserUI.getContentCssInput().value = config.contentCss;
             DefaultParserUI.getChapterTitleCssInput().value = config.titleCss;
             DefaultParserUI.getUnwantedElementsCssInput().value = config.removeCss;
             DefaultParserUI.getTestChapterUrlInput().value = config.testUrl;
         }
```
dteviot, line 139: I think should return at this point, as there's already a configuration.

Not sure why you're continuing on.

plugin/js/HeuristicScanner.js
```diff
@@ -0,0 +1,156 @@
...
+            if (!doc) {
+                let xhr = await HttpClient.fetchHtml(url);
+                doc = xhr.responseXML;
+            }
+
+            if (!doc) {
+                return { status: "failed", error: "Failed to parse DOM" };
+            }
+
+            let walker = doc.createTreeWalker(doc.body, NodeFilter.SHOW_TEXT, null, false);
```
dteviot, line 17: This should be pulled into it's own function, so it' obvious what it's doing. (Trying to figure out if page is a Table of Contents, based on how much of the page text is links.)

plugin/js/DefaultParserUI.js
```diff
@@ -118,16 +129,91 @@ class DefaultParserUI { // eslint-disable-line no-unused-vars
...
+        // Always ensure the Test URL is filled out, falling back to activeUrl
+        if (!DefaultParserUI.getTestChapterUrlInput().value) {
+            DefaultParserUI.getTestChapterUrlInput().value = activeUrl;
+        }
+
+        DefaultParserUI.bindSmartScanner();
+
+        // Proactively scan using the live DOM to avoid network/403 issues.
+        // This will override the old/default config purely in the UI if successful.
+        if (dom && activeUrl) {
```
dteviot: Not sure why you always bind, but not execute.

plugin/js/DefaultParserUI.js
312Aroha: Updated, added early return.

plugin/js/HeuristicScanner.js
312Aroha: Extracted to `isTableOfContents`.

312Aroha:
Updated based on the feedback.

Besides that, I also added one minor safety guard:
- Wrapped LocalStorage parsing in a `try-catch` to safely handle corrupted data.

dteviot:
@312Aroha Thank you for your contribution

PR #2941, opened 2026-08-21
Add Konkon/Kuupress support

MineRobber9000:
[Konkon](https://konkon.ink) and [Kuupress](https://kuupress.com) are novel sharing sites from the team originally behind [ZetroTranslation](https://zetrotranslation.com) (shutting down at the end of this month). They share a similar backend, so I figured I'd PR in support for both of them at the same time.

~~Unfortunately, due to how their API authenticates requests, we can't actually get chapters that aren't completely unlocked to the public (basically, no `Origin` header, no user authentication, even if we have the user's `konkon-`/`kuupress-session` cookie). Still, being able to download all free chapters of a novel is better than nothing.~~ So apparently I completely missed `HttpClient.setDeclarativeNetRequestRules` when I was looking at other parsers... that one's on me, lol. I can now confirm it works (at least, I was able to download a chapter I bought on Konkon; like I said, the two sites are extremely similar on the backend, so I believe it should also work on Kuupress).

One thing to note: currently, all Konkon/Kuupress novels trigger the WEBP "your ePub reader may not be able to render this" warning. This is because all covers are stored as WEBP on the backend (except the default covers, which are stored as JPG).

### DIFF
--- a/plugin/js/parsers/KonkonKuupressParser.js
+++ b/plugin/js/parsers/KonkonKuupressParser.js
@@ -4,13 +4,33 @@ parserFactory.register("konkon.ink", () => new KonkonKuupressParser("https://kon
 parserFactory.register("kuupress.com", () => new KonkonKuupressParser("https://kuupress.com", "https://api-kp.kuupress.com", "Kuupress"));
 
 class KKPFetchCache {
-    constructor() {
+    constructor(siteUrl, apiUrl) {
         this.cache = {};
+        this.siteUrl = siteUrl;
+        this.apiUrl = apiUrl;
     }
 
     async fetch(url) {
         url = url.toString();
         if (this.cache[url]) return this.cache[url];
+        await HttpClient.setDeclarativeNetRequestRules([
+            {
+                id: 1,
+                priority: 1,
+                action: {
+                    type: "modifyHeaders",
+                    requestHeaders: [{
+                        header: "Origin",
+                        operation: "set",
+                        value: this.siteUrl
+                    }]
+                },
+                condition: {
+                    requestDomains: [(new URL(this.apiUrl).hostname)],
+                    resourceTypes: ["xmlhttprequest"]
+                }
+            }
+        ]);
         let contents = (await HttpClient.fetchJson(url)).json;
         this.cache[url] = contents;
         return contents;
@@ -23,7 +43,7 @@ class KonkonKuupressParser extends Parser {
         this.siteUrl = siteUrl;
         this.apiUrl = apiUrl;
         this.publisher = publisher;
-        this.fetchCache = new KKPFetchCache();
+        this.fetchCache = new KKPFetchCache(siteUrl, apiUrl);
     }
 
     dispatchOnChapterOrNovel(url, chapterBranch, novelBranch) {
--- a/plugin/js/parsers/KonkonKuupressParser.js
+++ b/plugin/js/parsers/KonkonKuupressParser.js
@@ -86,10 +86,14 @@ class KonkonKuupressParser extends Parser {
     async getChapterUrlsFromChapter(url) {
         // Get chapter JSON object
         let data = await this.fetchCache.fetch(url);
+        // Make sure we have data.data
+        if (!data.data) {
+            throw new Error("Something changed on the site backend (no chapter info)");
+        }
         // Get novel slug
-        let novelSlug = data.data?.novel?.slug;
+        let novelSlug = data.data.novel?.slug;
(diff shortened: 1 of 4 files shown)


plugin/js/parsers/KonkonKuupressParser.js
```diff
@@ -0,0 +1,333 @@
...
+        // Converge on novel flow
+        return this.getChapterUrlsFromNovel(this.novelAPIFromSlug(novelSlug));
+    }
+
+    async getChapterUrlsFromNovel(url) {
+        // Get novel JSON object
+        let data = await this.fetchCache.fetch(url);
+        // Store novel URL and title on parser (used in addFirstPageUrlToWebPages below)
+        this.novelUrl = this.siteUrl + "/read/" + data.data?.slug;
+        this.novelTitle = data.data?.title;
```
dteviot: Is this actually going to work if data.data is null?  That will result in an undefined novelTitle.

plugin/js/parsers/KonkonKuupressParser.js
```diff
@@ -0,0 +1,333 @@
...
+        // Collect chapters (will be turned into URLs later)
+        let chapters = [];
+        // First pass: all unlocked chapters
+        let volumes = data.data?.volumes;
+        if (volumes == undefined) {
```
dteviot, line 118: OK, it's faulting here.  But should probably fault earlier, where we fetch data.

plugin/js/parsers/KonkonKuupressParser.js
```diff
@@ -0,0 +1,333 @@
...
+
+    async fetchActualChapter(baseUrl, dataUrl) {
+        // Get chapter JSON object
+        let data = await this.fetchCache.fetch(dataUrl);
+        // Make doc for content
+        let doc = Parser.makeEmptyDocForContent(baseUrl);
+        // Chapter title
+        doc.dom.title = data.data?.title;
+        // Chapter content into doc
+        doc.content.innerHTML = data.data?.content;
```
dteviot: Do not assign unsanitized data to innerHTML.
That's just sets up a painful, time consuming fight with the Mozilla code auditor.

Have a look at https://github.com/dteviot/WebToEpub/blob/4630402ec1b71b6cbafc0f7b1c41a903e6d7e782/plugin/js/parsers/DarkNovelsParser.js#L52-L59

plugin/js/parsers/KonkonKuupressParser.js
```diff
@@ -0,0 +1,333 @@
...
+        let data = await this.fetchCache.fetch(dataUrl);
+        // Make doc for content
+        let doc = Parser.makeEmptyDocForContent(baseUrl);
+        // Generate table of contents based on the novel
+        let createElementWithTextContent = KonkonKuupressParser.createElementWithTextContent.bind(null, doc.dom);
+        // Title
+        doc.dom.title = data.data?.title;
+        // Description (nested in blockquote)
+        let description = createElementWithTextContent("blockquote", "");
+        description.innerHTML = data.data?.description;
```
dteviot: Again, using innerHTML

plugin/js/parsers/KonkonKuupressParser.js
```diff
@@ -0,0 +1,333 @@
...
+    extractAuthor() {
+        let authorName = this.noveldata.author_name;
+        let authorUserName = this.noveldata.author_user_name;
+        if (authorName == authorUserName) return authorName;
+        return `${authorName} (TL by ${authorUserName})`;
+    }
+
+    extractDescription() {
+        let doc = Parser.makeEmptyDocForContent(this.siteUrl);
+        doc.content.innerHTML = this.noveldata.description;
```
dteviot: innerHTML again

dteviot (changes requested):
Thank you for your hard work.
Fix the innerHTML issue and I'll happily accept.

plugin/js/parsers/KonkonKuupressParser.js
MineRobber9000: I generally assume `data.data` exists, though I should definitely add a check for it after fetching the data. If the title or slug get moved, they will indeed be `undefined`. Should this raise an error instead?

plugin/js/parsers/KonkonKuupressParser.js
MineRobber9000: This check is actually for the volumes themselves (making sure they haven't been moved). I added the check by the fetch call for `data.data` itself.

dteviot:
Thank you for your contribution

PR #2944, opened 2026-08-22
feat(readrift): optimize initial load

matthewsong18:
### Problem: 
Initial load (populating the chapter list) for a novel with 300 chapters took around 21 seconds, which is pretty bad. Mb guys.

### Solution:
I originally copied the ToC pagination amount of 10, but by pinging the server, turns out the limit is actually 30. I also made the delay between requests a bit smaller at 150 to 250 ms with a jitter compared to the original 300 ms.

Overall, 64.52% less requests and initial load is around 3 times faster.

- Old: 31 requests finished in 21.25 seconds
- New: 11 requests finished in 6.45 seconds

### DIFF
--- a/plugin/js/parsers/ReadRiftParser.js
+++ b/plugin/js/parsers/ReadRiftParser.js
@@ -74,8 +74,10 @@ class ReadRiftParser extends Parser {
         while (apiUrl) {
             try {
                 apiUrl = await this.getChaptersFromApi(apiUrl, urls);
+                if (apiUrl) {
+                    await util.sleep(150 + Math.random() * 100);
+                }
             } catch (err) {
-                apiUrl = null;
                 throw new Error(
                     `ReadRiftParser failed while scanning novel's chapters: ${err.message}`,
                 );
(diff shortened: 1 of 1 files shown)


gamebeaker:
I would prefer something like this:

```
let saveThrottle = this.minimumThrottle;
.
.
.
this.minimumThrottle = 150 + Math.random() * 100
await this.rateLimitDelay();
.
.
.
this.minimumThrottle =saveThrottle;

```

The same is in the [ScribblehubParser.js](https://github.com/dteviot/WebToEpub/blob/ExperimentalTabMode/plugin/js/parsers/ScribblehubParser.js)

I find it better as in case the website suddenly enforces slower ratelimits for toc api requests the user can smimply change the slowdown speed in the advanced option and if the user choosen slow down is greater than150 + rand() the larger slow down is choosen.

matthewsong18:
@gamebeaker Implemented the user-defined delay override

gamebeaker:
@matthewsong18 thanks for your help <3

Issue #2920, opened 2026-08-14
Fenrirealm.com Crashing every 15-20 or so chapters


bookimp:
**Describe the bug**

**To Reproduce**
1. Go to ['...'](https://fenrirealm.com/series/her-talent-for-courting-death-has-always-been-impressive)
2. https://fenrirealm.com/series/her-talent-for-courting-death-has-always-been-impressive)
3. Download
**Expected behavior**
For it to download without crashing

**Screenshots**

**Desktop (please complete the following information):**
 - OS: Win 11 64
 - Browser Opera Gx 8 134.0.5954.44, Edge Version 151.0.4129.78 (Official build) (64-bit)

**Additional context**
The chapters is crashed on (22.2, 30.2, 38.3) all open just fine in browser. 

Just to be sure it wasn't just that novel I also tried it on https://fenrirealm.com/series/cultivating-immortality-i-am-the-evil-ancestor-shuttling-between-two-worlds-apocalypse-and-cultivation with the same issue.

ARYAN-9099:
@bookimp I think this is a issue with the rate limits. try increasing the delay.
If you find the minimum delay that works i will make the change in the code.

ARYAN-9099:
this is duplicate of #2719 
There someone managed to fix it with 5000ms delay.

bookimp:
> [@bookimp](https://github.com/bookimp) I think this is a issue with the rate limits. try increasing the delay. If you find the minimum delay that works i will make the change in the code.

3000 worked for me on both novels I tried although I didn't try anything between 2000 and 3000 after working up to it. If you do your downloads too close together it skips chapters but waiting a bit worked fine.

ARYAN-9099:
Nice, I'll put in a pr to change it then.

gamebeaker:
Thanks to @ARYAN-9099 for the fix.

Test versions for Firefox and Chrome have been uploaded to https://github.com/dteviot/WebToEpub/releases/tag/developer-build. Pick the one suitable for you, follow the "How to install from Source (for people who are not developers)" instructions at https://github.com/dteviot/WebToEpub/tree/ExperimentalTabMode#user-content-installation and let us know how it goes.

bookimp:
> Thanks to [@ARYAN-9099](https://github.com/ARYAN-9099) for the fix.
> 
[...]

Issue #2719, opened 2026-06-08
FenriRealm not working


juraijd12:
**Describe the bug**
The extension shows that chapters are successfully processed (read and marked as downloaded), but no “Save As” prompt or file download appears after reaching 100%.

Previously, the extension worked normally. This issue started recently.

**To Reproduce**

Open a supported web novel/manga site
Use the extension to download multiple chapters (tested range: 100–500 chapters)
Wait for processing to reach 100%
Observe that no download prompt appears

**Expected behavior**
After reaching 100%, a “Save As” dialog or automatic download should appear, allowing the EPUB file to be saved locally.

**Actual behavior**

Chapters are processed and marked as completed
Progress reaches 100%
No file download or “Save As” popup appears
No EPUB file is saved

**Screenshots**

<img width="826" height="568" alt="Image" src="https://github.com/user-attachments/assets/e2380410-d9f0-475a-a9ff-523ce1638cd8" />

**Desktop (please complete the following information):**

OS: W11
Browser: Latest
Extension: WebToEpub 

**Additional context**

Issue tested with 3 chapters and also large batches (100–500 chapters), same result
Extension previously worked fine last month under the same usage
Issue seems site-dependent: works on some websites, but fails on others
No browser download prompt appears even though completion reaches 100%

penguinbrief:
Getting the same issue here. The same problem happens when using "Add to Library" or "Pause to Library" instead of "Pack EPUB", the chapters never actually get added to the library.

IsamZareer:
Same issue here. Tried on Firefox and chrome and a different novel from the same website, still same issue

Jin-ar:
Same issue as well

RulerWentricosus:
+1

maphie14:
I used dev build and it works or you can wait for update

zytancl:
> I used dev build and it works or you can wait for update

unfortunately it doesnt work as well for me

dteviot:
I just tested the current test version with: 
* https://fenrirealm.com/series/that-time-i-got-reincarnated-as-an-elf-god-op-mc, chapters 0, 1, 2
* https://fenrirealm.com/series/how-could-the-villainous-young-master-be-a-saintess. chapters 1 and 2

There were no problems.

If running in chrome, after opening WebToEpub and before pressing "Pack Epub" you could open the DevTools and go to the Network tab.
This should allow you to see if WebToEpub is actually downloading the chapters.
[...]

PR #2945, opened 2026-08-23
Fix Fenrirealm.com

ARYAN-9099:
Add 3 sec delay to respect the rate limits.
fixes the problems in #2920 and #2719


ARYAN-9099:
@gamebeaker you forgot to put labels to the issues 😊

gamebeaker:
I added them thanks for the reminder.

PR #2948, opened 2026-08-24
fix: export sleepController as getter to fix post-pause truncation

xRahul:
## Summary
Fixes an issue where pausing a download to Library permanently broke subsequent downloads, causing them to stop after 1 chapter and save to Library instead of downloading.

## What Was Broken
When you clicked "Pause to Library":
1. Download stopped and saved partial book to Library as intended.
2. But every novel you tried to download *afterwards* stopped after fetching just 1 chapter (or 1 batch) and wrote that 1-chapter book to Library instead of downloading an EPUB file.
3. Clicking "Pause to Library" on later downloads also stopped working.

## Root Cause
In `plugin/js/Util.js`, `util.sleepController` was exported as a static property reference evaluated at load time:
```javascript
// Old code:
return {
    sleepController: sleepController,
    // ...
};
```
When `util.resetSleepController()` created a new `AbortController`, the closure variable changed, but external consumers (`main.js`, `Parser.js`) continued reading the original aborted controller reference.

## Changes Made
1. **`plugin/js/Util.js`**: Export `sleepController` using an ES6 getter (`get sleepController() { return sleepController; }`). All callers now dynamically access the active `AbortController`.
2. **`plugin/js/main.js`**: Add a defensive `util.resetSleepController()` call at the start of `fetchContentAndPackEpub()` so every new pack starts with a clean controller state.
3. **`unitTest/UtestUtil.js`**: Add regression tests verifying:
   - `sleepController` returns a fresh un-aborted instance after `resetSleepController()`.
   - `util.sleep()` resolves immediately when `abort()` is triggered.

## How to Test
1. Start downloading any multi-chapter novel.
2. Click **Pause to Library**. Confirm partial book is stored in Library.
3. Start downloading another novel (or the same one) normally.
4. Confirm it fetches all chapters and triggers the standard file download instead of aborting after 1 chapter.

### DIFF
--- a/plugin/js/Parser.js
+++ b/plugin/js/Parser.js
@@ -599,7 +599,7 @@ class Parser {
                 await Promise.all(group.map(async (webPage) => this.fetchWebPageContent(webPage)));
                 index += group.length;
                 group = this.groupPagesToFetch(pagesToFetch, index);
-                if (util.sleepController.signal.aborted) {
+                if (util.getSleepController().signal.aborted) {
                     break;
                 }
             }
--- a/plugin/js/Util.js
+++ b/plugin/js/Util.js
@@ -1174,7 +1174,7 @@ const util = (function() {
         BLOCK_ELEMENTS: BLOCK_ELEMENTS,
         HEADER_TAGS: HEADER_TAGS,
         sleep: sleep,
-        sleepController: sleepController,
+        get sleepController() { return sleepController; },
         resetSleepController: resetSleepController,
         randomInteger: randomInteger,
         isFirefox: isFirefox,
--- a/plugin/js/Util.js
+++ b/plugin/js/Util.js
@@ -1174,7 +1174,7 @@ const util = (function() {
         BLOCK_ELEMENTS: BLOCK_ELEMENTS,
         HEADER_TAGS: HEADER_TAGS,
         sleep: sleep,
-        get sleepController() { return sleepController; },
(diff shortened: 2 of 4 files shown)


gamebeaker:
@xRahul i can only test it next week. Sorry for the delay.

dteviot:
@gamebeaker 

I'll leave this to you to test and merge of OK.

gamebeaker:
@xRahul thank you for the PR. I just changed the getter to be more like the other getter in the project.

@dteviot the tests fail on firefox. I tested the firefox version and it seems to work it is just that the tests fail. Should the tests be removed as they fail on firefox?

gamebeaker:
I was wrong the test passes the AI found the issue:
>Why it fails in Firefox but not Chrome:
QUnit has a built-in feature called reorder (enabled by default). It saves the names of failed tests in the browser's sessionStorage and automatically runs those failed tests first on the next page load.
>
>  If you previously had a failing test in Firefox (even just a temporary syntax error), QUnit is now running your tests in a different order in Firefox than in Chrome.
>
>  If a test that aborts the controller now runs before the test you shared, your assertion asert.notOk(controller1.signal.aborted) will fail because the singleton was already aborted by the previous test.
>
>The Fix:
Clear your sessionStorage in Firefox (via DevTools -> Storage), or temporarily disable QUnit reordering. To prevent this permanently, force a reset before every single test using QUnit modules:

PR #2949, opened 2026-08-24
fix(images): drop failed image fetches and prune dead tags from epub

xRahul:
## Description
When image fetches fail (network timeouts, HTTP errors, blocked hosts), `ImageCollector.fetchImage` caught the error but unconditionally pushed the unpopulated `ImageInfo` to `imagesToPack`. Because its `arraybuffer` remained `null`, `ImageInfo.packInEpub` packed a 4-byte string `"null"` (`new Blob([null])`) into the EPUB zip, while OPF manifest, spine, and chapter XHTML referenced the missing/corrupted file.

This change:
- Drops failed image fetches from `imagesToPack` in the `fetchImage` catch block and marks `imageInfo.isFailed = true`.
- Resets `this.coverImageInfo = null` if a cover image fetch fails, and guards `EpubItemSupplier.hasCoverImageFile()`.
- Updates `ImageTagReplacer.replaceTag` to prune wrapper/image elements from chapter XHTML when image fetch failed or `imageInfo` is missing.
- Advances the progress indicator on fetch failure so the UI progress counter remains accurate.
- Adds defensive guards in `ImageInfo.packInEpub` and `FontInfo.packInEpub` preventing null buffers from being added to the zip writer.
- Adds `"FontInfo": "readonly"` to ESLint globals and disables unused var warning on class declaration.
- Adds regression unit tests in `UtestImageCollector.js` and `UtestEpubItem.js`.

## Testing
- `cd eslint && node pack.js && eslint *.js` passes with 0 errors.
- Unit tests added and verified for `fetchImage` failure handling, `replaceImageTags` element pruning, and `packInEpub` null buffer guards.

### DIFF
--- a/eslint/.eslintrc.js
+++ b/eslint/.eslintrc.js
@@ -70,6 +70,7 @@ module.exports = {
         "FetchErrorHandler": "readonly",
         "FetchImageErrorHandler": "readonly",
         "Firefox": "readonly",
+        "FontInfo": "readonly",
         "FootnoteExtractor": "readonly",
         "HttpClient": "readonly",
         "ImageCollector": "readonly",
--- a/plugin/js/EpubItem.js
+++ b/plugin/js/EpubItem.js
@@ -175,6 +175,7 @@ class ImageInfo extends EpubItem { // eslint-disable-line no-unused-vars
         this.width = null;
         this.dataOrigFileUrl = dataOrigFileUrl;
         this.queuedForFetch = false;
+        this.isFailed = false;
     }
 
     getZipHref() {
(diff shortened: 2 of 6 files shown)


gamebeaker:
I see this as intendet behaviour as the author wanted an image at this position.
I think it is important information for the reader that there should be a picture but it is now missing. If the user wants to he can try downloading the image as the original url is saved in the epub.

If you want to remove the 0 bytes images you can use [EpubEditor](https://github.com/dteviot/EpubEditor) to remove all of them.

xRahul:
@gamebeaker if inntended, we should show it better. Showing null is not user friendly. Let me know if need to change this behavior to show something like [Image not found] instead of removing these.

dteviot:
@xRahul 
As gamebeaker said, the intention was to allow someone to fix the missing images by just replacing them in the epub, without needing to change anything else.  I'm not sure it actually worked, because we don't know the size of the intended image.

>  Let me know if need to change this behavior to show something like [Image not found] 

That's not a bad suggestion.  I'm not sure if we'd have one stock image (e.g. a SVG that says "Placeholder: WebToEpub unable to retrieve image.") Or if we need to provide one for each common image type.  png, jpg, webp, etc.

PR #2957, opened 2026-08-28
Fix Sbxh parser

Bartuzen:
Updated chapter link fetching to support the site's latest changes.


dteviot:
Hard to test without a URL, but as I think you wrote the original one, I'll accept this.

PR #2962, opened 2026-08-31
Fix typo in `Parser.js` function and 63 other Parsers

Yomafil:
Fixed a minor typo in the function `findConstrutedContent` in `Parser.js`
`Construted` was missing the `c` before the `t` to make it `Constructed`

### DIFF
--- a/plugin/js/Parser.js
+++ b/plugin/js/Parser.js
@@ -780,7 +780,7 @@ class Parser {
         };
     }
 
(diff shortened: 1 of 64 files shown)


dteviot:
@Yomafil 
You really didn't need to do this.  But thank you for doing so.

Yomafil:
It was a quick fix. And It was annoying me when I couldn't easily find the function with autocomplete.

PR #2971, opened 2026-09-03
Display the current version of the extension in the popup

Yomafil:
Add the current version of the extension at the top of the popup.
image

### DIFF
--- a/plugin/css/default.css
+++ b/plugin/css/default.css
@@ -30,6 +30,14 @@
     border-bottom: 1px #CCC solid;
 }
 
+#extensionVersionSection {
+    text-align: center;
+}
+
+#spanWebToEpubVersion {
+    display: inline-block;
+    text-align: center;
+}
 
 progress {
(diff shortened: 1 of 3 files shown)


gamebeaker:
Why?

Yomafil:
For somewhat minor reasons.
Mostly when developing new feat or fix, it is useful to know what version is currently installed when testing.
And for when people post screenshots and reports bugs, allow us to know what version they are using, and allow a user to know what version they have if they install the dev build of the extension.

I myself had multiple time wanted to know what version of the extension I was currently using, without having to go check for it in the extensions tab.

If it's really bothersome, it could be hidden by default and put behind an advanced option, or be moved to another location in the popup.

dteviot:
@gamebeaker 
> and allow a user to know what version they have 

This would be the main reason.  Most users probably don't know the version is shown on the extensions tab/page.
It's a small thing, but it does have benefit.

Issue #2976, opened 2026-09-09
[Feature Request] Hyperlink Table of Contents URL in EPUB


Dawn99999:
Normally I download my books using Calibre's FanFicFare(FFF), but this tool is excellent for whenever I have issues with FFF or a site just doesn't work for FFF. One of the best features of FFF is that it hyperlinks the title, letting me click on it to take me to the site I downloaded it from.

In the information chapter of the EPUB, the first thing that appears is the Table of Contents URL, and having the link be a hyperlink which I can click to take me to the place I downloaded the book from would be incredible.

gamebeaker:
@Dawn99999 thanks to @Yomafil for the PR

Test versions for Firefox and Chrome have been uploaded to https://github.com/dteviot/WebToEpub/releases/tag/developer-build. Pick the one suitable for you, follow the "How to install from Source (for people who are not developers)" instructions at https://github.com/dteviot/WebToEpub/tree/ExperimentalTabMode#user-content-installation and let us know how it goes.

Dawn99999:
Worked perfectly, thank you.

PR #2980, opened 2026-09-13
Change tocUrl in InfoPage to HyperLink

Yomafil:
Change the tocUrl in the InfoPage of the Epub into an HyperLink.
#2976


dteviot:
@Yomafil 
Note, not all epub readers support this.  But for those that do, it won't hurt.  And I don't think it will hurt the ones that don't either.

PR #2987, opened 2026-09-15
Custom Error Message for NoContentToError403 userPref

Yomafil:
Add a custom error message when a user has selected the `Treat 'no content found error' as 'HTTP 403 error'.` user preference.


plugin/js/HttpClient.js
```diff
@@ -58,8 +58,14 @@ class FetchErrorHandler {
 
     promptUserForRetry(url, wrapOptions, response, failError) {
         let msg;
+        let userPreferences = main.getUserPreferences();
         if (wrapOptions.retry.HTTP === 403) { 
-            msg = new Error(UIText.Warning.warning403ErrorResponse(new URL(response.url).hostname) + this.makeFailCanRetryMessage(url, response.status));
+            if (userPreferences.noContentToError403) {
```
gamebeaker: Should be:

            if (userPreferences.noContentToError403.value) {

gamebeaker:
It also shows the new warning even if it is a real 403 error. Is that intended?

plugin/js/HttpClient.js
gamebeaker: i don't know what github marked there as git diff. (It is wrong)

Yomafil:
No, that shouldn't be happening.

Yomafil:
Tested and should be working now.

gamebeaker:
@Yomafil i meant that it shows the new warning if "Treat 'no content found error' as 'HTTP 403 error'." is selected and WebToEpub gets a real 403 error from the Network.

WebToEpub gets a real 403 Network error and the message is that:
```
Error: Warning: No content found on target page, treated as site `www.scribblehub.com` sending an Access Denied (403) error.
You may need to logon to site, or browse site normally
until you get a Cloudflare "Are you a human" page or satisfy some other CAPTCHA
before WebToEpub can continue.
...
```

As it says "No content found" i find it misleading. But it is just a minor complaint. I mean technically it couldn't find the content as it got a 403 response.

Yomafil:
@gamebeaker 
I kept it that way because the primary use case for the option is for when a site sends a challenge page and we have no way in the parser to recognize it, leading to a `No Content Found` error which stops the downloading with no option to clear the captcha or retry downloading the chapter.
So instructing the user on how to proceed from there sounded good to me.
I can always change the name of the option to something clearer as `Treat 'no content found error' as 'HTTP 403 error'.` isn't very clear on what an HTTP 403 error means.
Maybe changing the option name in the advanced option view to `Allow retry on 'No Content Found' error` would be better?

Yomafil:
@gamebeaker 
I might have misread your comment.
> i meant that it shows the new warning if "Treat 'no content found error' as 'HTTP 403 error'." is selected and WebToEpub gets a real 403 error from the Network.
>
> WebToEpub gets a real 403 Network error and the message is that:

This should have been fixed. When the option is unselected, it should now throw the real HTTP 403 error message and not the `No Content -> Error 403` error message.
This was unintended behavior.

gamebeaker:
I think there is a slight misunderstanding. It doesn't matter.

> Maybe changing the option name in the advanced option view to Allow retry on 'No Content Found' error would be better?

I would be for that as 403 etc is cryptic for non it people.

Issue #2991, opened 2026-09-16
"Allow retry on 'No Content Found' error." makes problems


gamebeaker:
The new advanced function "Allow retry on 'No Content Found' error." makes problems in case the parser has a function like buildChapter() where the Chapter gets created from a json as it calls the function 
```
    findContent(dom) {
        return Parser.findConstructedContent(dom);
    }
```
from the parser with the json response witch isn't even HTML but json.

@Yomafil as it is your feature can you fix it?

gamebeaker:
You can test it with https://wtr-lab.com/en/novel/50964/anime-crossover-starting-with-nexus-i-am-justice for example for faster debugging just force reduce the delay to 0.

Yomafil:
I'll take a look at it.

Yomafil:
Can't test with wtr-lab.
`WARNING: Site 'p3-reading-sign.fqnovelpic.com' has sent an Access Denied (403) error.`
Can't really debug and find a solution when the site doesn't even let me get content regularly...
Can't seem to reproduce the error in other parser.

Yomafil:
```js
isNoContentToError403AndContentNull(response) {
    if (this.userPreferences.noContentToError403.value) {
        if (response.responseXML) {
            let content = this.findContent(response.responseXML);
            return (content == null);
        }
        else {
            return false;
        }
    }
    else {
        return false;
    }
}
```
This might work? 
But when the site sends an `HTTP Error 403`, `response.responseXML` is true...
We could also reverse the order and prioritize `isCustomError` which could be better?
```js
let response = await fetch(url, wrapOptions.fetchOptions);
let ret = await HttpClient.checkResponseAndGetData(url, wrapOptions, response);
if (wrapOptions.parser?.isCustomError(ret)) {
    let CustomErrorResponse = wrapOptions.parser.setCustomErrorResponse(url, wrapOptions, ret);
    return wrapOptions.errorHandler.onResponseError(CustomErrorResponse.url, CustomErrorResponse.wrapOptions, CustomErrorResponse.response, CustomErrorResponse.errorMessage);
}
if (wrapOptions.parser?.isNoContentToError403AndContentNull(ret)) {
    let CustomNoContentToError403Response = wrapOptions.parser.setNoContentToError403Response(url, wrapOptions, ret);
[...]

PR #2994, opened 2026-09-16
Fix `isNoContentToError403` always failing when `response.responseXML` is undefined

Yomafil:
Fix for `isNoContentToError403AndContentNull()` always failing and throwing an error when `response.responseXML` is undefined when using another fetch function.
Tested and working with https://wtr-lab.com/en/novel/99260/soul-land-yu-xiaogang-says-my-martial-soul-is-only-for-support-so-i-become-a-divinity-through-music
#2991


Yomafil:
There are still some minor issue with the option.
Example: Masking the real `HTTP Error 403` when fetching images in https://wtr-lab.com/en/novel/50964/anime-crossover-starting-with-nexus-i-am-justice

gamebeaker:
@Yomafil what happens with preprocessRawDom()? I don't know of an example but it could be that `this.findContent(response.responseXML);` fails because preprocessRawDom would rearrange the document.

Yomafil:
@gamebeaker 
`preprocessRawDom()` is called later in the chain in `fetchWebPageContent()` and not in `wrapFetchImpl()` where we do the check, so it shouldn't have an impact on it.
But I might be able to do something about it.

gamebeaker:
> preprocessRawDom() is called later in the chain in fetchWebPageContent() and not in wrapFetchImpl() where we do the check, so it shouldn't have an impact on it.

That is my point i think that the check shouldn't be in HTTPClient.js at all.

Yomafil:
The only reason it is there is because I couldn't find a way to fit it later in the chain (skill issue...)
I'll see what I can do about it. 
At worse we can just comment the check in `HTTPClient` if it's causing more problems then it solves.
Doing that shouldn't break anything...

gamebeaker:
I would try to place the code in Parser.js line 642
https://github.com/dteviot/WebToEpub/blob/8f0ac97457a659bf150ec00e78b6fa28753d5c32/plugin/js/Parser.js#L640-L648

and than wrap the inner try statement in a do...while loop that only gets triggert in case line 642 gets reached.

In line 642 i would maybe try something like this that is currently in HttpClient.js

https://github.com/dteviot/WebToEpub/blob/8f0ac97457a659bf150ec00e78b6fa28753d5c32/plugin/js/HttpClient.js#L223-L226

